PR #221, opened 2026-01-07
Feat redb adapter

jlucaso1:
This can crosscompiling to another platforms easily, I got better performance in high traffic (in memory both sqlite and redb, 700 message/s vs 1000 messages/s). Also the final binary size is smaller (~7mb vs ~5.5mb). Compile faster too.

The drawbacks are the inspection thing compared to sqlite. (maybe we can create some useful scripts in adapter). Also I've noticed a increased initial memory usage compared to sqlite (need to investigate this).

## Summary by CodeRabbit

* **New Features**
  * Redb-based storage is now the default backend with in-memory and on-disk modes and per-device scoping; public store APIs are exposed.
* **Chores**
  * Project manifests and default feature set updated to enable the Redb storage backend; runtime startup now uses the Redb backend.
* **Tests**
  * Test helpers centralized and refactored to use shared Redb-backed factories, simplifying test setup and device-specific tests.


jlucaso1:
memory usage will improve after: https://github.com/cberner/redb/pull/1111

jlucaso1:
@codex review

PR #384, opened 2026-03-18
perf: remove unnecessary locks identified by WA Web comparison

jlucaso1:
## Summary

Compared our mutex/lock usage against captured WhatsApp Web JavaScript (`docs/captured-js/`) to identify locks that are unnecessary or over-engineered for the actual access patterns.

- **Remove `plaintext_buffer_pool`** — WA Web creates a fresh `WABinary.Binary()` per `encodeStanza()` call with no pooling. Our pool added 2 async mutex acquisitions on every outbound `send_node()` call. For typical message sizes (~1-4KB), `Vec::with_capacity(1024)` is nearly free (~20ns) while tokio Mutex lock/unlock under contention is ~50-200ns.
- **Remove `last_successful_connect`** — Dead code: written once on auth success but never read anywhere in the codebase.
- **Replace `pending_retries` `Mutex<HashSet>` → `DashSet`** — Simple insert/remove/contains operations. `DashSet::insert()` atomically returns whether the value was new (replacing the lock-check-insert block). The scopeguard cleanup becomes synchronous, eliminating a `tokio::spawn` that existed solely to acquire the mutex for removal.
- **Fix `pending_retries` dedup key** — Changed from bare `message_id` to participant-aware `dedupe_key` (`chat:msg_id:participant` for groups), matching WA Web's retry receipt model where different group participants can independently request retries for the same message.
- **Simplify `EncryptSendError`** — Removed `plaintext_buf` and `out_buf` fields that existed solely for buffer reuse. Simplified `SendResult` from `Result<(Vec<u8>, Vec<u8>)>` to `Result<()>` since no caller needs buffers back.

## WA Web verification

- **Buffer pool**: WA Web's `encodeStanza()` (`1IPUUpcJg9V.js:8820`) creates fresh `WABinary.Binary()` per call, no recycling
- **Retry dedup**: WA Web uses per-chat `PromiseQueueMap` serialization (`cnub3BAEa6g.js:12668`), and retry receipts include `participant` for groups (`2ZwnB7JiNTs.js:850`)
- **`last_successful_connect`**: WA Web embeds timestamps in media connection data, no separate tracking field

## What was kept (and why)

After deep analysis, these locks were confirmed as justified:
- `app_state_key_requests` — multi-step transaction (iterate → check → insert → retain) needs atomicity
- `app_state_syncing` — batch insert in `sync_collections_batched` requires atomicity to prevent duplicate syncs
- `presence_subscriptions` — works fine as-is, small set, marginal improvement from DashSet
- `chatstate_handlers` — RwLock matches write-once-read-many pattern perfectly

## Test plan

- [x] `cargo fmt --all` clean
- [x] `cargo clippy --all-targets` clean
- [x] `cargo test --all --exclude e2e-tests` all pass

### DIFF
--- a/src/socket/error.rs
+++ b/src/socket/error.rs
@@ -35,65 +35,41 @@ pub struct EncryptSendError {
     pub kind: EncryptSendErrorKind,
     #[source]
     pub source: anyhow::Error,
-    pub plaintext_buf: Vec<u8>,
-    pub out_buf: Vec<u8>,
 }
 
 impl EncryptSendError {
-    pub fn crypto(
-        source: impl Into<anyhow::Error>,
-        plaintext_buf: Vec<u8>,
-        out_buf: Vec<u8>,
-    ) -> Self {
+    pub fn crypto(source: impl Into<anyhow::Error>) -> Self {
         Self {
             kind: EncryptSendErrorKind::Crypto,
             source: source.into(),
-            plaintext_buf,
-            out_buf,
         }
     }
 
-    pub fn framing(
-        source: impl Into<anyhow::Error>,
-        plaintext_buf: Vec<u8>,
-        out_buf: Vec<u8>,
-    ) -> Self {
+    pub fn framing(source: impl Into<anyhow::Error>) -> Self {
         Self {
             kind: EncryptSendErrorKind::Framing,
             source: source.into(),
-            plaintext_buf,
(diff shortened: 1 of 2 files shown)


jlucaso1:
@codex review

jlucaso1:
@claude review

PR #397, opened 2026-03-20
fix: align digest key validation with WA Web behavior

jlucaso1:
## Summary
- Switch background init from `send_digest_key_bundle()` to `validate_digest_key()` — the full WA Web `digestKey()` flow
- Handle `ParseError` (empty/malformed `<digest/>`) at debug level instead of warn
- Stop re-uploading prekeys on validation failures (regId mismatch, missing prekey, hash mismatch) — matches WA Web's catch-all behavior

## Problem
1. `send_digest_key_bundle()` sends the raw IQ and discards the response. WA Web always runs the full `digestKey()` validation flow.
2. When the server returns an empty `<digest/>` (no keys on server), the parser fails with `ParseError(<registration> child not found)`, producing noisy WARN logs on every connection.
3. On validation failures (regId mismatch, missing local prekey, hash mismatch), the old code triggered `upload_pre_keys_with_retry()`. WA Web does NOT re-upload in these cases.

## WA Web Analysis
WA Web's `WAWebDigestKeyJob.digestKey()` uses a tri-state variable `a`:
- `a = true` (only on 404) → re-upload
- `a = false` (all other cases) → skip
- `a = null` (unreachable) → skip

The success path wraps `validateLocalKeyBundle` in:
```javascript
try {
  (yield y(r.result), (a = !1));  // success → a = false
} catch (e) {
  a = !1;                          // ANY exception → a = false
}
```
Both branches set `a = false`, meaning validation failures (regId mismatch, missing prekey, hash mismatch) are silently caught without triggering re-upload. Only 404 (no record on server) triggers `uploadPreKeys()`.

## Fix
- Use `validate_digest_key()` in background init (full WA Web flow)
- `ParseError`: catch silently at debug level (WA Web has an unhandled rejection here — we're stricter)
- All validation failures: log warning, return `Ok(())`, do NOT re-upload — matches WA Web exactly
- Only 404 triggers `upload_pre_keys_with_retry()`

## Test plan
- [x] `cargo fmt --all`
- [x] Verified against WA Web source (`WAWebDigestKeyJob.digestKey()`, `WADeprecatedWapParser`, `WAParseIqResponse`)
- [x] Traced all 12 code paths through both implementations — exact match

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved digest validation to gracefully handle unparseable responses, avoiding unnecessary re-uploads.
  * Reduced repeated key re-uploads by treating various validation failures as non-fatal, preventing redundant network activity.
  * Clarified failure logging for digest validation to make operational issues easier to diagnose without changing visible behavior.

### DIFF
--- a/src/prekeys.rs
+++ b/src/prekeys.rs
@@ -215,13 +215,16 @@ impl Client {
         }
     }
 
-    /// Validate server key bundle digest and re-upload if mismatched.
+    /// Validate server key bundle digest, re-uploading only when the server has no record.
     ///
     /// Matches WA Web's `WAWebDigestKeyJob.digestKey()`:
     /// 1. Queries server for key bundle digest (identity + signed prekey + prekey IDs + SHA-1 hash)
-    /// 2. Loads local keys and computes SHA-1 over the same material
-    /// 3. If hash mismatch or server returns 404: triggers `upload_pre_keys_with_retry()`
-    /// 4. If server returns 406/503/other: logs and does nothing
+    /// 2. If server returns 404 (no record): triggers `upload_pre_keys_with_retry()`
+    /// 3. If server returns 406/503/other error: logs and does nothing
+    /// 4. On success: loads local keys and computes SHA-1 over the same material
+    /// 5. If validation fails (regId mismatch, missing prekey, hash mismatch): logs warning,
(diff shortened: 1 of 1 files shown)


jlucaso1:
@claude review

PR #411, opened 2026-03-22
feat: add Runtime::yield_frequency() for configurable cooperative yielding

jlucaso1:
## Summary

Replaces hardcoded yield-every-10-frames in the frame processing loop with `Runtime::yield_frequency()`, allowing runtimes to control how often cooperative yielding happens.

- Default: `10` (multi-threaded — other tasks run on separate threads)
- Single-threaded runtimes override to `1` (yield every item to avoid starving the event loop)
- `yield_now()` still controls *whether* to yield (`None` = skip, `Some(fut)` = yield) — `yield_frequency()` only controls *how often* to check

## Files changed

- `wacore/src/runtime.rs` — add `yield_frequency()` with default `10` to both native and wasm32 `Runtime` trait variants
- `src/client.rs` — replace hardcoded `10` with `self.runtime.yield_frequency()`

## Test plan

- [x] `cargo clippy --all-targets` clean
- [x] Backwards compatible — default is `10` (same as before)

## Release Notes

* **Bug Fixes**
  * Enhanced event loop handling with configurable task scheduling frequency, improving responsiveness and preventing event loop starvation, especially in single-threaded environments.

### DIFF
--- a/wacore/src/runtime.rs
+++ b/wacore/src/runtime.rs
@@ -30,7 +30,7 @@ pub trait Runtime: Send + Sync + 'static {
     /// the call zero-cost on runtimes that don't need cooperative yielding.
     fn yield_now(&self) -> Option<Pin<Box<dyn Future<Output = ()> + Send>>>;
 
-    /// How often to yield in tight loops (every N items).
+    /// How often to yield in tight loops (every N items). Defaults to 10.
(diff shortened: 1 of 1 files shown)


jlucaso1:
@claude review

PR #412, opened 2026-03-22
perf: skip redundant signal store writes when data is unchanged

jlucaso1:
## Summary
- **Byte-equality dedup in `SignalStoreCache`**: `put_session`, `put_identity`, and `put_sender_key` now skip marking entries dirty when the serialized bytes are identical to what's already cached. Prevents unnecessary DB writes during flush on hot send/receive paths.
- **Remove `store_session` debug logging**: The old code acquired a device read lock, loaded the existing session from cache, deserialized it, and compared base keys — all just for debug log output. Removed to match WhatsApp Web's silent `storeSession` behavior.
- **DRY refactor**: Extracted duplicated `put`/`delete` logic (previously copy-pasted 3x each) into `StoreState::put()` and `StoreState::delete()` methods.

Net result: **-41 lines**, fewer allocations on every `store_session` call, fewer redundant DB writes on flush.

## Test plan
- [ ] `cargo clippy --all --tests` passes clean
- [ ] `cargo test --all` passes
- [ ] E2E message send/receive works correctly (sessions, identity, sender keys still persist)

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal session storage logic by removing unnecessary cache lookups and comparisons.
  * Centralized cache management helpers to improve code maintainability and reduce duplication across storage operations.

### DIFF
--- a/src/store/signal_adapter.rs
+++ b/src/store/signal_adapter.rs
@@ -84,49 +84,7 @@ impl SessionStore for SessionAdapter {
     ) -> Result<(), SignalProtocolError> {
         let addr_str = address.to_string();
         let record_bytes = record.serialize()?;
-
-        // Check for base key changes using cache (not DB) for logging
-        let device = self.0.device.read().await;
-        let existing_data = self
-            .0
-            .cache
-            .get_session(&addr_str, &*device.backend)
-            .await
-            .ok()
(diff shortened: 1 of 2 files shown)


jlucaso1:
@claude review

PR #433, opened 2026-03-23
feat: cstoken (NCT) privacy token fallback + proto update

jlucaso1:
## Summary

Update the proto to WhatsApp Web `2.3000.1035617621` and add the `cstoken` / NCT fallback path for privacy-token-gated 1:1 messaging.

This completes the send-side privacy-token chain to match WA Web:

```text
tctoken -> cstoken -> no token (463)
```

Where:
- `tctoken`: server-issued trusted-contact token from prior interaction
- `cstoken`: `HMAC-SHA256(nct_salt, recipient_account_lid)` using the bare `user@lid`

## Why this matters

WhatsApp's privacy-token rollout gates access to sensitive endpoints behind proof of relationship. Without a valid token, the server rejects requests with `463`.

`tctoken` was already implemented. This PR adds the missing `cstoken` fallback so first-contact messaging can succeed when the server has provisioned `nct_salt`, even before any trusted-contact token exists.

## Scope

- `waproto`: proto update adds `NctSaltSyncAction` and `HistorySync.nctSalt`
- `wacore/store`: persist `nct_salt`
- `wacore/history_sync`: parse `HistorySync.nctSalt`
- `src/client.rs`: handle `nct_salt_sync` set/remove mutations
- `src/send.rs`: attach `cstoken` when no usable `tctoken` exists and the recipient LID can be resolved
- `sqlite-storage`: add `nct_salt` migration/schema persistence
- `tests/e2e`: add coverage for provisioned/unprovisioned salt flows and token-gated endpoints

## Important behavior / precedence

- App-state `nct_salt_sync` is authoritative.
- History sync only backfills NCT salt when no authoritative app-state mutation has been seen yet. It does not resurrect a removed salt or overwrite a newer syncd value.
- Empty salts are ignored instead of being persisted.
- Empty tc tokens are treated as missing and are not cached, so they cannot suppress the `cstoken` fallback on later sends.
- tc token refresh state is only updated after a successful post-send issuance flow.
- Sent-node waiters are cleared on disconnect so transport-scoped test/diagnostic waiters cannot leak across reconnects.

## Endpoint behavior

| Endpoint | Behavior after this PR |
|----------|------------------------|
| 1:1 message send | use valid `tctoken`, else `cstoken` if `nct_salt` + recipient LID are available, else no token / server may return `463` |
| Profile picture | still `tctoken`-gated |
| Presence subscribe | still `tctoken`-gated |

## E2E coverage added

- no salt -> first contact still gets `463`
- history sync salt -> first contact succeeds via `cstoken`
- syncd salt -> first contact succeeds via `cstoken`
- local salt clear/remove -> first contact fails again
- `nct_salt` survives reconnect and still enables first contact
- PN-target first contact still uses `cstoken` after PN -> LID resolution
- `tctoken`-only reply path remains valid when `cstoken` is disabled
- restricted profile picture and presence remain `tctoken`-gated

Live availability still depends on server AB props / rollout (`wa_nct_token_*`). Accounts that are not provisioned with `nct_salt` will still follow the no-salt / `463` branch.

## Validation

- [x] `cargo test -p wacore --lib`
- [x] `cargo test -p whatsapp-rust --lib`
- [x] `cargo test -p e2e-tests --no-run -q`
- [ ] live verification on accounts that actually receive `nct_salt` via server rollout

## Release Notes

* **New Features**
  * Enhanced privacy token mechanism with cstoken fallback for improved message delivery reliability.
  * Server-provisioned NCT salt integration for privacy token computation.
  * Improved handling of first-contact scenarios with privacy-restricted recipients.

* **Bug Fixes**
  * Corrected tc-token issuance timing to occur after message transmission.

* **Tests**
  * Comprehensive test coverage for privacy tokens, cstoken scenarios, and restricted contact handling.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1061,7 +1061,11 @@ impl Client {
     }
 
     async fn cleanup_connection_state(&self) {
-        self.clear_node_waiters();
+        // Note: node_waiters are intentionally NOT cleared here — they are
+        // cross-connection (callers may register a waiter before an action that
+        // completes on a subsequent connection, e.g. after 515 reconnect).
+        // sent_node_waiters ARE cleared because they match pre-encryption
+        // outgoing stanzas, which are transport-scoped.
         self.clear_sent_node_waiters();
         self.is_logged_in.store(false, Ordering::Relaxed);
         self.is_ready.store(false, Ordering::Relaxed);
@@ -3053,18 +3057,6 @@ impl Client {
             } else {
                 i += 1;
             }
-        }
-    }
-
-    fn clear_node_waiters(&self) {
-        let mut waiters = self
-            .node_waiters
-            .lock()
-            .unwrap_or_else(|poisoned| poisoned.into_inner());
-        let count = waiters.len();
-        if count > 0 {
-            waiters.clear();
-            self.node_waiter_count.fetch_sub(count, Ordering::Release);
         }
     }
 


jlucaso1:
@claude review

PR #441, opened 2026-03-26
fix: parse digest prekey IDs from <id> nodes, not <key>

jlucaso1:
## Summary
- The digest response parser filtered `<list>` children by `tag == "key"`, but the real server sends them as `<id>` nodes — silently dropping all prekey IDs and causing a SHA-1 hash mismatch on every digest validation
- Removed the tag filter to match WA Web's `mapChildren` behavior (iterates all children without tag filtering)
- Added e2e test that queries the mock server digest endpoint, recomputes the hash locally, and asserts equality

## Test plan
- [x] Unit tests pass (`cargo test -p wacore --lib -- prekeys::tests::test_digest_key_bundle`)
- [x] E2E test passes (`cargo test -p e2e-tests --test digest_key`)


jlucaso1:
@claude review

PR #446, opened 2026-03-26
fix: audit findings — prekey wrap-around, chain key overflow, log levels

jlucaso1:
## Summary

Fixes two Signal protocol bugs and improves observability based on a comprehensive codebase audit.

### PreKey ID wrap-around fix (`src/prekeys.rs`)

- IDs now wrap via modulo 2^24 instead of stopping at the boundary
- `start_id` itself is wrapped before the generation loop, preventing counter pinning when lingering high-ID rows exist
- Prevents eventual prekey exhaustion that would block session establishment

### Chain key index overflow fix (`wacore/libsignal/src/protocol/ratchet/keys.rs`)

- `next_chain_key()` and `step_with_message_keys()` now use `checked_add()` and return `Result`
- Matches the sender chain key's existing bounds checking
- 4 regression tests using pattern-matched error variants

### Log level fix (`src/client.rs`)

- Unhandled `<call>` stanzas downgraded from WARN to DEBUG

### Retry comment cleanup (`src/retry.rs`)

- Removed hardcoded JS file reference, uses constant name instead

## Test plan

- [x] 4 overflow regression tests (pattern-matched)
- [x] `cargo clippy --all --tests -- -D warnings` clean
- [x] `cargo fmt --all` clean

### DIFF
--- a/src/prekeys.rs
+++ b/src/prekeys.rs
@@ -111,11 +111,6 @@ impl Client {
             key_pairs_to_upload.push((pre_key_id, key_pair));
         }
 
-        if keys_to_upload.is_empty() {
-            log::warn!("No pre-keys available to upload");
-            return Ok(());
-        }
-
         // Encode once — reused for both pre-upload store and post-upload mark.
         let encoded_batch: Vec<(u32, Vec<u8>)> = {
             use prost::Message;


jlucaso1:
@claude review

PR #451, opened 2026-03-27
fix: recognize own LID devices as self in DM device partitioning

jlucaso1:
## Summary

- Extract `partition_dm_devices()` and `is_exact_dm_sender_device()` from inline logic in `prepare_dm_stanza`
- Match own devices by both PN **and** LID user (via `matches_user_or_lid`), not just PN
- Pass `own_lid` through `prepare_dm_stanza` so LID-addressed devices are correctly classified
- Avoid unnecessary allocations: borrow `device_snapshot.pn`/`.account` instead of cloning, move `Jid`s instead of cloning each one

### Bug

The old code only checked `device_jid.user == own_jid.user` (PN). In LID-addressed DMs, own devices appear with a LID user string, so they were misclassified as recipient devices instead of own-other-devices. This caused incorrect DeviceSentMessage wrapping.

### WA Web parity

WA Web uses `isMeAccount(deviceJid)` in `MsgCreateFanoutStanza.js:33,73` which checks both PN and LID identity. The new `matches_user_or_lid` check matches this behavior.

## Test plan

- [x] `cargo fmt --all` clean
- [x] Existing `test_self_encryption_prevention_regression` passes
- [x] New `test_dm_encryption_treats_own_lid_devices_as_self` covers the LID case

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved direct message delivery to correctly exclude sender's own devices using enhanced device identification.

* **Refactor**
  * Optimized data flow in direct message preparation logic.

### DIFF
--- a/src/send.rs
+++ b/src/send.rs
@@ -988,9 +988,8 @@ impl Client {
             let device_snapshot = self.persistence_manager.get_device_snapshot().await;
             let own_jid = device_snapshot
                 .pn
-                .clone()
+                .as_ref()
                 .ok_or(crate::client::ClientError::NotLoggedIn)?;
-            let account_info = device_snapshot.account.clone();
 
             // Include tctoken in 1:1 messages (matches WhatsApp Web behavior).
             // Skip for newsletters, groups, and own JID.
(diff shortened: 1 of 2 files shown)


jlucaso1:
@claude review

PR #458, opened 2026-03-28
refactor!: add TaskTracker for structured task lifecycle management

jlucaso1:
## Summary

- Add `TaskTracker` to `wacore::runtime` — a lightweight container that stores `AbortHandle`s and aborts them all on demand or when dropped
- Add `#[must_use]` to `AbortHandle` so the compiler warns when a spawned task's handle is silently dropped (forces explicit `.detach()` or tracking)
- Add `connection_tasks` / `client_tasks` fields to `Client`:
  - **connection_tasks**: tracked for diagnostics and drained at the start of each `connect()` call to prevent unbounded handle accumulation across reconnects. Tasks exit cooperatively via `connection_generation`, `is_shutting_down()`, and `shutdown_notifier` — no forceful abort on disconnect (tasks like the post-login sequence must survive 515 reconnect cycles)
  - **client_tasks**: aborted when `Client` drops — covers LID-PN cache warm-up, device registry cleanup, persistence background saver
- Invalidate `message_queues` on disconnect (bug fix: stale per-chat workers survived reconnects)
- `run_background_saver` now returns `AbortHandle` instead of detaching internally
- 14 `.detach()` calls remain as intentional fire-and-forget (event dispatch callbacks, best-effort acks/receipts, transport disconnect teardown, background DB writes, sync worker with Weak ref, per-stanza processing)

## Breaking changes

- `PersistenceManager::run_background_saver()` now returns `AbortHandle` instead of `()`. Callers must either track the handle (e.g., via `client_tasks.track()`) or call `.detach()` to preserve the old fire-and-forget behavior.

## Test plan
- [x] `cargo test --all --exclude e2e-tests` — all 1090 tests pass
- [x] Verified `#[must_use]` produces a warning when `runtime.spawn()` result is unused
- [x] Verified remaining 14 `.detach()` calls are all intentional fire-and-forget
- [x] Verified `connection_tasks.abort_all()` is called at start of `connect()` (not during `cleanup_connection_state`) to avoid aborting the post-login task during 515 reconnect cycles
- [x] Verified `message_queues.invalidate_all()` closes channels so per-chat workers exit
- [x] Verified cooperative cancellation via `connection_generation` / `shutdown_notifier` handles staleness

## Summary by CodeRabbit

**New Features**
- Added background task activity monitoring to diagnostics output

**Improvements**
- Enhanced cleanup of stale background tasks during connection state transitions
- Improved background task lifecycle management by assigning appropriate ownership scopes
- Refined task handling during reconnection cycles to prevent leftover operations from prior sessions

### DIFF
--- a/src/bot.rs
+++ b/src/bot.rs
@@ -650,7 +650,7 @@ impl BotBuilder<Provided, Provided, Provided, Provided> {
                 .map_err(|e| anyhow::anyhow!("Failed to create persistence manager: {}", e))?,
         );
 
-        persistence_manager
+        let bg_saver_handle = persistence_manager
             .clone()
             .run_background_saver(runtime.clone(), std::time::Duration::from_secs(30));
 
@@ -686,6 +686,9 @@ impl BotBuilder<Provided, Provided, Provided, Provided> {
             self.cache_config,
         )
         .await;
+
+        // Track the background saver on the client's lifetime
+        client.client_tasks.track(bg_saver_handle);
 
         // Register custom enc handlers
         for (enc_type, handler) in self.custom_enc_handlers {
--- a/src/client.rs
+++ b/src/client.rs
@@ -156,6 +156,9 @@ pub struct MemoryDiagnostics {
     // -- Misc --
     pub chatstate_handlers: usize,
     pub custom_enc_handlers: usize,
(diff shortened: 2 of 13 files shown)


jlucaso1:
@claude review

PR #460, opened 2026-03-28
fix: address 9 audit findings across correctness, safety, and performance

jlucaso1:
## Summary

Fixes 9 issues found during a codebase audit, covering correctness bugs, safety improvements, and performance optimizations.

### HIGH — Correctness bugs

**1. Sender key check bypasses signal cache** (`src/send.rs`)
- Group message path used `device_guard.load_sender_key()` (direct DB read with write lock)
- Now uses `signal_cache.get_sender_key()` with read lock, matching the status broadcast path
- Previously caused unnecessary SKDM re-distribution on every group send until cache flush

**2. Missing UndecryptableMessage event for group decrypt failures** (`src/message.rs`)
- `NoSenderKeyState` errors spawned retry receipt but never dispatched `UndecryptableMessage`
- Consumer/UI never learned the message was pending decryption
- Now dispatches the event before retry, matching the session-based decrypt path

**3. Double `cleanup_connection_state` on transport disconnect** (`src/client.rs`)
- `cleanup_connection_state()` was called both inside the message loop (on Disconnected event) and in `run()` after the loop exits
- Removed the redundant call inside the message loop — `run()` handles it

### MEDIUM — Safety and performance

**4. `increment_retry_count` TOCTOU** (`src/message.rs`)
- Documented the theoretical get-then-insert race window
- `spawn_retry_receipt` detaches via `runtime.spawn`, so per-chat serialization does not cover this path; in practice retries for the same message are rare and recipients deduplicate by message ID

**5. Signal cache `flush()` per-store lock-through-IO** (`wacore/src/store/signal_cache.rs`)
- Each store (sessions, identities, sender_keys) is flushed independently under its own lock
- Only ONE store is locked during its I/O — the other two remain free for concurrent encrypt/decrypt
- No dirty-set race: the lock is held from snapshot through write through clear, so mutations to the same store are blocked until flush completes
- Strictly better than original (all 3 locks held simultaneously) and correct unlike the 3-phase approach (which had a race where mutations between snapshot and clear lost dirty markers)

**6. Message queue capacity reduced** (`src/handlers/message.rs`)
- `bounded(10000)` → `bounded(500)` per chat
- Prevents memory amplification with many active chats while still handling bursts

### LOW — Defensive improvements

**7. PortableCache removal** (`src/portable_cache.rs`)
- Kept the original `retain()`-based removal — the lazy deletion optimization was reverted because remove→reinsert breaks FIFO eviction order (stale key causes live entry to be evicted)
- O(capacity) `retain` is correct and bounded by max_capacity

**8. `now_millis() as u64` guard** (`keepalive.rs`, `client.rs`)
- Added `.max(0)` before `as u64` cast at all 11 sites to prevent silent wrap on negative clock values

**9. Fragile JID type heuristic** (`src/message.rs`)
- Replaced `server.contains(".us")` with proper `is_group()` / `is_broadcast_list()` / `is_status_broadcast()` methods

## Test plan
- [x] `cargo test --all --exclude e2e-tests` — all tests pass
- [x] Verified sender key check matches the existing status broadcast pattern
- [x] Verified UndecryptableMessage event signature matches existing dispatch_undecryptable_event
- [x] Verified signal cache flush holds only one store lock at a time during I/O
- [x] Verified dirty sets are cleared only after successful writes, under the same lock that snapshotted them
- [x] Verified no dirty-set race: lock held from snapshot through clear prevents mid-flush mutations from losing markers

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate message sender classification for correct message-type handling.
  * Hardened timestamp handling to prevent negative-time casting issues in connection tracking.
  * Now emits an undecryptable-message event and triggers retry receipts when decryption lacks keys.
  * Improved disconnection event handling to avoid duplicate or premature disconnect notifications.

* **Chores**
  * Reduced per-chat queue capacity to lower memory use under bursts.
  * Safer, incremental encryption key cache flushes to minimize lock scope and backend errors.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -873,25 +873,36 @@ impl Client {
             if let Err(connect_err) = self.connect().await {
                 error!("Failed to connect: {connect_err:#}. Will retry...");
             } else {
-                if self.read_messages_loop().await.is_err() {
+                let unexpected_disconnect = if self.read_messages_loop().await.is_err() {
                     // Check intentional_reconnect AFTER read loop exits — reconnect()
                     // sets this flag while the loop is running, so it must be read here.
                     if self.expected_disconnect.load(Ordering::Relaxed)
                         || self.intentional_reconnect.swap(false, Ordering::Relaxed)
                     {
                         debug!("Message loop exited during expected disconnect.");
+                        false
                     } else {
                         warn!(
                             "Message loop exited with an error. Will attempt to reconnect if enabled."
                         );
+                        true
                     }
                 } else if self.expected_disconnect.load(Ordering::Relaxed) {
                     debug!("Message loop exited gracefully (expected disconnect).");
+                    false
                 } else {
                     info!("Message loop exited gracefully.");
-                }
+                    false
+                };
 
                 self.cleanup_connection_state().await;
+
+                // Dispatch after cleanup so handlers see cleared connection state.
+                if unexpected_disconnect {
+                    self.core
+                        .event_bus
+                        .dispatch(&Event::Disconnected(crate::types::events::Disconnected));
+                }
             }
 
             if !self.enable_auto_reconnect.load(Ordering::Relaxed) {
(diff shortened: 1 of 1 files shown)


jlucaso1:
@claude review

jlucaso1:
@claude review

PR #480, opened 2026-04-02
feat: unknown device detection and deferred device sync

jlucaso1:
## Summary

- Add WA Web-compliant `isFromKnownDevice()` check after group message decryption
- Send correct retry reason `UnknownCompanionNoPrekey` (code 11) instead of `NoSession` (code 1) for unknown devices
- Batch unknown-device users during offline sync via `PendingDeviceSync`, flush with a single usync query after offline delivery completes (with 2s delay matching WA Web's `OFFLINE_DEVICE_SYNC_DELAY`)
- Invalidate stale device cache before sync to force network fetch of updated device lists
- Parse `offline` attribute from message stanzas to distinguish online vs offline dispatch

## Context

When receiving group messages (`skmsg`) from a device not in our device list (e.g., sender added device `:10` while we were offline and we only know `:8`/`:9`), we were:
1. Attempting decryption → `NoSenderKeyState` error
2. Sending retry receipt with wrong reason code (`NoSession = 1`)
3. **Never querying the sender's device list** to learn about the new device

WA Web handles this differently (`WAWebHandleMsgProcessUtils.preProcessMsg`):
- Checks `isFromKnownDevice(author)` after successful decryption
- Online: triggers immediate `syncDeviceListJob` (usync query)
- Offline: batches into `OfflinePendingDeviceCache`, flushes via `doPendingDeviceSync()` after offline delivery ends (2s delay)
- Sends retry reason `UnknownCompanionNoPrekey = 11`

This was observed causing 13+ unrecoverable messages per offline sync from a single user who added a new device.

## Test plan

- [x] `cargo fmt --all` passes
- [x] All 1,101 tests pass (0 failures)
- [ ] Deploy to Docker, verify logs show `UnknownCompanionNoPrekey` instead of `NoSession` for unknown devices
- [ ] Verify usync query fires after offline sync completion for batched users
- [ ] Verify subsequent messages from previously-unknown devices decrypt successfully

## Summary by CodeRabbit

* **New Features**
  * Background batching and periodic flush of pending device syncs to recover devices after offline delivery
  * Offline-delivered messages now carry an explicit offline marker for clearer status

* **Improvements**
  * Unknown-device handling now triggers targeted re-syncs, sends retry receipts, and avoids processing undecryptable payloads
  * Retry logic updated to include keys earlier for certain unknown-device cases and to improve device cache invalidation during reconnection

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -1116,15 +1116,18 @@ impl Client {
     async fn handle_unknown_device_sync(self: &Arc<Self>, info: &MessageInfo) {
         let user_jid = info.source.sender.to_non_ad();
 
+        // Dedup: skip if we already have a sync pending/in-flight for this user
+        if !self.pending_device_sync.add(user_jid.clone()).await {
+            return;
+        }
+
         if info.is_offline {
             log::debug!("Queueing {} for pending device sync (offline)", user_jid);
-            self.pending_device_sync.add(user_jid).await;
         } else {
             log::debug!("Triggering immediate device sync for {}", user_jid);
             let client = Arc::clone(self);
             self.runtime
                 .spawn(Box::pin(async move {
-                    // Invalidate stale record so get_user_devices hits the network
                     client.invalidate_device_cache(&user_jid.user).await;
                     if let Err(e) = client.get_user_devices(&[user_jid]).await {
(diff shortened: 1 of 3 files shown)


jlucaso1:
@claude review

PR #482, opened 2026-04-02
fix: route session migration through signal cache instead of backend

jlucaso1:
## Summary

Route `migrate_signal_sessions_on_lid_discovery()` through `SignalStoreCache` instead of accessing the backend directly. Also removes the now-redundant backend reload in `try_pn_to_lid_migration_decrypt()`.

## Context

`migrate_signal_sessions_on_lid_discovery()` reads sessions/identities directly from the backend store, bypassing the `SignalStoreCache`. On backends where the cache is authoritative (MemoryStore, future WASM stores), the backend contains stale serialized data from the last flush. If the session ratcheted since then (e.g., SKDM encryption during a group send), the migration reads stale bytes and writes them to the LID address, corrupting the session.

**Reproduction (MemoryStore):**
1. DM user A → works (session created in cache at PN address)
2. Create group with user A, send group message → works (usync discovers LID, triggers migration)
3. DM user A again → `NoSession` or `BadMac` (migrated session has wrong ratchet state)

**Why invisible on SqliteStore:** The SQL database is always current because every `put_session` during encryption writes directly to it. The cache is just a performance layer, not the source of truth.

WA Web doesn't even copy/move sessions during PN→LID migration — it resolves the address at lookup time via `createSignalAddress()` using the cached LID mapping. Our approach (physical migration) is valid but must go through the cache to read current state.

## Changes

- `src/client/lid_pn.rs`: All session/identity reads go through `signal_cache.get_session()/get_identity()` (cache-first with backend fallback). All writes go through `signal_cache.put_session()/put_identity()` (marks dirty for flush). Flush after migration to persist.
- `src/message.rs`: Removed redundant backend reload in `try_pn_to_lid_migration_decrypt()` — migration now populates the cache directly.

## Test plan

- [x] All tests pass (0 failures)
- [ ] On MemoryStore: DM → group send (triggers migration) → DM again should work
- [ ] On SqliteStore: no behavior change (cache falls through to backend)

## Summary by CodeRabbit

* **Refactor**
  * Enhanced session and identity data migration handling to use more efficient caching mechanisms.

### DIFF
--- a/src/client/lid_pn.rs
+++ b/src/client/lid_pn.rs
@@ -169,7 +169,10 @@ impl Client {
     }
 
     /// Migrate Signal sessions and identity keys from PN to LID address.
-    /// WA Web never stores sessions under PN when a LID mapping is known.
+    ///
+    /// All reads/writes go through `signal_cache` to avoid reading stale data
+    /// from the backend when the cache has unflushed mutations (e.g., after
+    /// SKDM encryption ratcheted the session).
     pub(crate) async fn migrate_signal_sessions_on_lid_discovery(&self, pn: &str, lid: &str) {
         use log::{info, warn};
         use wacore::types::jid::JidExt;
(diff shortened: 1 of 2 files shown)


jlucaso1:
@claude review

Issue #508, opened 2026-04-08
PDO-only approach for <unavailable> doesn't reliably deliver — add retry receipt


mcaldas:
## Context

PR #506 fixed `<unavailable>` by requesting PDO from the primary phone. However, in testing, the PDO-only approach doesn't reliably deliver messages.

## What Works

The **original approach** in #506's first commit (`handle_decrypt_failure` with `RetryReason::NoSession`) worked reliably — messages arrived via PDO relay within 1 second. This sends both:
1. A **retry receipt** to the sender
2. A **PDO request** to the primary phone (via `spawn_pdo_request` inside `handle_decrypt_failure`)

## What Doesn't Work Reliably

The second commit changed to `spawn_pdo_request_with_options(&info, true)` only (no retry receipt). In testing:
- PDO request is sent successfully
- Primary phone acknowledges with a Delivered receipt  
- But the PDO response never comes back (waited 60+ seconds)
- The phone is confirmed online (delivery receipt proves it)

## Hypothesis

The retry receipt may be needed because:
1. It tells the **sender** to re-encrypt for this specific device
2. The sender's re-encryption triggers the WA server to route properly
3. Without the retry receipt, the phone receives the PDO request but doesn't know it needs to relay

## Proposed Fix

Add `self.spawn_retry_receipt(&info, RetryReason::NoSession)` alongside the existing `spawn_pdo_request_with_options`, similar to how `handle_decrypt_failure` does both:

```rust
if let Some(unavailable) = unavailable_node {
    // ... existing code ...
    self.spawn_pdo_request_with_options(&info, true);
    self.spawn_retry_receipt(&info, RetryReason::NoSession);  // ADD THIS
    self.core.event_bus.dispatch(&Event::UndecryptableMessage(...));
    return;
}
```

## Testing

- Tested with two wa-rs companion devices on different WA accounts
- With retry receipt: message delivered via PDO in <1 second
- Without retry receipt (current code): PDO request sent, phone acks, but no relay

jlucaso1:
I had no luck trying to reproduce that issue. Can you provide some logs (debug in preference)?

mcaldas:
@jlucaso1 I can reproduce reliably. Here are debug logs with sanitized JIDs.

**Setup:** Two whatsapp-rust companion devices on different WhatsApp accounts:
- Device A: companion on Account X (device `:24`)  
- Device B: companion on Account Y (device `:2`)
- Both primary phones are online

[...]

PR #509, opened 2026-04-08
fix: PDO cache key mismatch and add response guards

jlucaso1:
## Summary
- Fix PDO cache key using LID-resolved JID while phone responds with PN JID (every lookup missed)
- Send PDO to bare own JID (`to_non_ad()`) instead of `with_device(0)`
- Add `is_from_me` + `device == 0` guards on PDO response processing
- Add `unavailable_request_id` to `MessageInfo` for PDO-recovered messages
- Rename `StanzaKey` to `ChatMessageId`, use as typed cache key instead of `format!()` strings

## Context

Debugging #508 revealed the PDO cache key was built with LID-resolved JIDs on the request side, but the phone's response always contains PN JIDs in the `WebMessageInfo` key. This caused every cache lookup to miss, falling through to the slower `message_info_from_web_message_info` reconstruction path.

Verified against WA Web (`NonMessageDataRequest.js:412-421`) which uses LID for the protobuf `MessageKey` but PN for internal tracking, and whatsmeow which sends PDO to `ownID.ToNonAD()` and guards responses with `info.Sender.Device == 0`.

## Test plan
- [x] Tested with two-account reproduction (sender wa-rust bot + receiver companion device)
- [x] `<unavailable>` triggers PDO, phone responds in ~1.3s, message delivered
- [x] Cache hit confirmed (no "Cache miss" in logs)
- [x] `cargo clippy --all` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * PDO/placeholder responses now processed only from the primary device, reducing duplicate/invalid recoveries.
  * Improved validation of response sources and JID parsing to ignore malformed entries.

* **Refactor**
  * Message caching and pending-request indexing moved to a unified message-ID key model for more consistent lookups.
  * PDO request targeting updated to use the primary non-ad addressing form for delivery.

* **New**
  * Recovered messages now include an explicit recovery-request identifier in their metadata.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -3517,7 +3517,7 @@ impl Client {
     }
 
     /// Creates a normalized ChatMessageId by resolving PN to LID JIDs.
-    pub(crate) async fn make_stanza_key(&self, chat: &Jid, id: &str) -> ChatMessageId {
+    pub(crate) async fn make_chat_message_id(&self, chat: &Jid, id: &str) -> ChatMessageId {
         // Resolve chat JID to LID if possible
         let chat = self.resolve_encryption_jid(chat).await;
 
--- a/src/client.rs
+++ b/src/client.rs
@@ -262,7 +262,7 @@ pub enum ClientError {
     NotLoggedIn,
 }
 
-use wacore::types::message::StanzaKey;
+use wacore::types::message::ChatMessageId;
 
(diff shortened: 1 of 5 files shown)


jlucaso1:
@claude review

jlucaso1:
@claude review

PR #510, opened 2026-04-09
perf: lazy ciphertext in SignalMessage, eliminate redundant allocation

jlucaso1:
## Summary

- Remove redundant `ciphertext: Box<[u8]>` field from `SignalMessage` — it stored the same bytes already embedded in `serialized`
- Replace with `OnceLock<Box<[u8]>>` lazy cache (mirrors existing `SenderKeyMessage` pattern)
- On **send path**: `.body()` is never called, so the cache stays empty — eliminates 1 heap allocation + memcpy of N ciphertext bytes per sent message (44% less struct heap memory)
- On **receive path**: `TryFrom` eagerly populates the cache, so `.body()` is still a zero-cost pointer return
- **Breaking**: `SignalMessage::body()` return type changes from `&[u8]` to `Result<&[u8]>` (only caller updated in `session_cipher.rs`)
- Adds iai-callgrind claim-validation benchmarks that isolate and measure each proposed optimization independently

## Benchmark results

| Path | Instructions | Change |
|------|-------------|--------|
| DM encrypt (subsequent) | 160,849 | +0.06% (noise) |
| DM encrypt (first) | 159,691 | -0.11% (noise) |
| DM decrypt (first) | 5,508,558 | +0.006% (noise) |

No measurable CPU regression on any real path.

## Test plan

- [x] `cargo test -p wacore-libsignal` — 102 tests pass
- [x] `cargo test -p whatsapp-rust` — 345 tests pass
- [x] `cargo bench -p wacore-libsignal --bench libsignal_benchmark` — no regressions

## Summary by CodeRabbit

* **Bug Fixes**
  - Decryption now stops and surfaces errors encountered while retrieving or decoding message contents, preventing attempts to decrypt invalid data.

* **Refactor**
  - Message handling now defers decoding ciphertext and caches decoded data lazily.
  - Cloning preserves the original serialized form and only duplicates cached plaintext when it has already been initialized, reducing eager memory use and work.

### DIFF
--- a/wacore/libsignal/src/protocol/protocol.rs
+++ b/wacore/libsignal/src/protocol/protocol.rs
@@ -13,6 +13,19 @@ use subtle::ConstantTimeEq;
 use crate::protocol::state::{PreKeyId, SignedPreKeyId};
 use crate::protocol::{IdentityKey, PrivateKey, PublicKey, Result, SignalProtocolError, Timestamp};
 
+/// Get-or-init for `OnceLock<Box<[u8]>>` with a fallible initializer.
+fn get_or_try_init_bytes(
+    cache: &OnceLock<Box<[u8]>>,
+    init: impl FnOnce() -> Result<Box<[u8]>>,
+) -> Result<&[u8]> {
+    if let Some(val) = cache.get() {
+        return Ok(val);
+    }
+    let _ = cache.set(init()?);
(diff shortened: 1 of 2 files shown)


jlucaso1:
@claude review

jlucaso1:
@claude review

PR #512, opened 2026-04-11
perf!: use CompactString for NodeValue, NodeContent, and Jid.user

jlucaso1:
## Summary

- Replace `String` with `CompactString` in `NodeValue::String`, `NodeContent::String`, and `Jid.user` to eliminate heap allocations for short strings (≤ 24 bytes stored inline)
- Add inline TTL-based eviction to `InMemoryBackend::store_sent_message` to cap unbounded growth
- Switch benchmark example to `InMemoryBackend` instead of `SqliteStore`

## Profiling results

Measured with heaptrack against a real WhatsApp session (benchmark example via bartender mock server):

| Metric | Before | After | Change |
|---|---|---|---|
| Peak heap | 7.90 MB | 5.15–5.24 MB | **-34%** |
| Peak RSS | 32.57 MB | 28.28–28.57 MB | **-13%** |
| Total allocations | 6,030,336 | 5,406,579 | -10.3% |

The two largest allocation sites from the baseline — `Attrs::FromIterator` at 1.88 MB and 750 KB (cloning `(Cow<str>, NodeValue)` tuples during `NodeRef::to_owned()`) — no longer appear in the top 15 peak consumers. `to_owned()` still allocates (Vec containers, byte content, long strings > 24 chars), but it is no longer a top-tier memory problem.

## Why CompactString

`CompactString` (`compact_str` crate) has the same `size_of` as `String` (24 bytes on x86_64) but stores strings ≤ 24 bytes inline — zero heap allocation. PoC benchmarks showed:

- 90.9% of `NodeValue::String` values are ≤ 24 chars (protocol values like `"msg"`, `"pkmsg"`, `"text"`, timestamps)
- 100% of `Jid.user` values (phone numbers, 10-15 chars) fit inline
- Only message IDs (32 chars) spill to heap

Tags and attribute keys keep `Cow<'static, str>` + `intern_cow()` — already zero-alloc for tokenized strings via static references.

## Breaking changes

- `NodeValue::String` wraps `CompactString` instead of `String`
- `NodeContent::String` wraps `CompactString` instead of `String`
- `Jid.user` is `CompactString` instead of `String`
- Jid factory methods take `impl Into<CompactString>` instead of `impl Into<String>`

### Migration guide

**Construction** — all existing patterns still work:
```rust
NodeValue::from("text")           // &str → CompactString (inline)
NodeValue::from(some_string)      // String → CompactString
Jid::pn("12345")                  // same as before
"user@server".parse::<Jid>()?     // same as before
```

**Pattern matching** — `CompactString` derefs to `&str`:
```rust
match value {
    NodeValue::String(s) => {
        // s: CompactString, but all &str methods work via Deref
        s.starts_with("pkmsg")  // works
        s == "text"             // works
    }
}
```

**When you need `String`**:
```rust
jid.user.to_string()              // CompactString → String
map.get(jid.user.as_str())        // for HashMap<String, _> lookups
```

`CompactString` is re-exported from both `wacore_binary::CompactString` and `whatsapp_rust::CompactString`.

## Test plan

- [x] `cargo test -p wacore-binary -p wacore -p whatsapp-rust` — all passing
- [x] `cargo check -p wacore-binary --all-features` — serde feature verified
- [x] heaptrack profiling (3 runs) confirms consistent ~34% peak heap reduction
- [ ] CI / e2e tests

## Summary by CodeRabbit

* **New Features**
  * In-memory message store gains configurable sent-message TTL with automatic time-based eviction.

* **Chores**
  * Example benchmark switched to the in-memory backend.
  * Workspace adopts a compact string type across crates; several public APIs now expose or return this compact string form (may affect downstream integrations).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -218,6 +218,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "castaway"
+version = "0.2.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dec551ab6e7578819132c713a93c022a05d60159dc86e7a7050223577484c55a"
+dependencies = [
+ "rustversion",
+]
+
+[[package]]
 name = "cbc"
 version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -280,6 +289,21 @@ name = "cmov"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "de0758edba32d61d1fd9f4d69491b47604b91ee2f7e6b33de7e54ca4ebe55dc3"
+
+[[package]]
+name = "compact_str"
+version = "0.8.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3b79c4069c6cad78e2e0cdfcbd26275770669fb39fd308a752dc110e83b9af32"
+dependencies = [
+ "castaway",
+ "cfg-if",
+ "itoa",
+ "rustversion",
+ "ryu",
+ "serde",
+ "static_assertions",
+]
 
 [[package]]
 name = "concurrent-queue"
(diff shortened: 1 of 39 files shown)


jlucaso1:
@claude review check for further refinements, possibly remove unnecessary clones, allocations, consistency, bugs, regressions, etc

jlucaso1:
@claude review things carefully; also, check possible additional refinements we can do before merging (since this PR is a breaking change, so we can do superb things)

PR #515, opened 2026-04-12
perf!: Arc<Event> event bus to eliminate deep clones on dispatch

jlucaso1:
## Summary

- `CoreEventBus::dispatch` takes `Event` by value, wraps in `Arc::new` once internally, shares via `Arc::clone` to all handlers
- Releases the handler `RwLock` before calling handlers (prevents deadlock if a handler calls `add_handler`)
- Skips Arc allocation entirely when no handlers are registered (early return)
- `EventHandler::handle_event` receives `Arc<Event>` instead of `&Event`
- `BotEventHandler` moves the Arc into the spawned task — zero clone
- `ChannelEventHandler` sends the Arc through the channel — zero clone
- `on_event` closure receives `Arc<Event>`; users match with `match &*event { ... }`
- Added `Event::as_message()` and `Event::message_text()` helpers
- Added `MessageContext::from_parts()` (direct construction) and `from_event()` (convenience wrapper)
- `TestEventCollector` stores `Arc<Event>` instead of deep-cloning events
- Removed dead `SharedData<T>` type, dead `handle_message` function, dead commented-out `Event::Qr` dispatch
- Updated `LazyConversation` docs — `Arc` dispatch gives parse-once across handlers for free
- Extracted duplicated `TestEventCollector` to shared `test_utils`
- Fixed stale doc examples to use `&*event` matching

## Why

The old bus dispatched `&Event` and every handler called `event.clone()` to own it for async tasks. `Event::Message(Box<wa::Message>, MessageInfo)` is expensive to deep-clone — the protobuf message has 40+ nested fields with Box allocations.

With `Arc<Event>`, the bus creates one shared allocation per dispatch. Each handler gets a cheap `Arc::clone` (atomic increment, zero bytes allocated).

## Lab results (experiments/heaptrack-impact-lab)

| Scenario | Alloc reduction | Bytes reduction | Time reduction |
|---|---|---|---|
| Core bus, 1 handler | **-94.1%** | **-98.8%** | **-77.3%** |
| Core bus, 3 handlers | **-94.1%** | **-98.8%** | **-87.5%** |

## Breaking changes

- `EventHandler::handle_event` takes `Arc<Event>` instead of `&Event`
- `ChannelEventHandler::new()` returns `Receiver<Arc<Event>>` instead of `Receiver<Event>`
- `on_event` closure takes `Arc<Event>` instead of `Event` — match with `&*event`
- `CoreEventBus::dispatch` takes `Event` by value instead of `&Event`

## Migration guide

```rust
// EventHandler impl
// Before:
fn handle_event(&self, event: &Event) {
    let owned = event.clone();
}
// After:
fn handle_event(&self, event: Arc<Event>) {
    let shared = Arc::clone(&event); // cheap — no deep copy
}

// on_event closure
// Before:
.on_event(|event, client| async move {
    match event {
        Event::Message(msg, info) => { /* msg/info owned */ }
    }
})
// After:
.on_event(|event, client| async move {
    match &*event {
        Event::Message(msg, info) => {
            // msg/info are references — zero clone for read-only handlers
            // Use from_parts when you need MessageContext:
            let ctx = MessageContext::from_parts(msg, info, client);
        }
    }
})

// ChannelEventHandler
// Before:
while let Ok(event) = rx.recv().await {
    match event { ... }
}
// After:
while let Ok(event) = rx.recv().await {
    match &*event { ... }
}

// MessageContext — still constructible from any source
let ctx = MessageContext {
    message: Box::new(msg_from_db),
    info: info_from_db,
    client,
};
```

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo test --all --exclude e2e-tests`

### DIFF
--- a/examples/benchmark.rs
+++ b/examples/benchmark.rs
@@ -48,11 +48,11 @@ fn main() {
         let mut bot = builder
             .on_event(move |event, client| async move {
                 match &*event {
-                    Event::Message(msg, _) => {
+                    Event::Message(msg, info) => {
                         if let Some(text) = msg.text_content()
                             && text == "ping"
                         {
-                            let ctx = MessageContext::from_event(&event, client).unwrap();
+                            let ctx = MessageContext::from_parts(msg, info, client);
                             info!("Received text ping, sending pong...");
 
                             let pong_text = format!("pong {}", ctx.info.id);
--- a/examples/benchmark.rs
+++ b/examples/benchmark.rs
@@ -47,11 +47,11 @@ fn main() {
 
         let mut bot = builder
             .on_event(move |event, client| async move {
-                match event {
+                match &*event {
                     Event::Message(msg, info) => {
                         let ctx = MessageContext {
-                            message: msg,
-                            info,
+                            message: msg.clone(),
+                            info: info.clone(),
                             client,
                         };
 
--- a/examples/benchmark.rs
+++ b/examples/benchmark.rs
@@ -48,12 +48,8 @@ fn main() {
         let mut bot = builder
             .on_event(move |event, client| async move {
                 match &*event {
-                    Event::Message(msg, info) => {
-                        let ctx = MessageContext {
-                            message: msg.clone(),
-                            info: info.clone(),
-                            client,
-                        };
+                    Event::Message(_, _) => {
+                        let ctx = MessageContext::from_event(&event, client).unwrap();
 
                         if let Some(text) = ctx.message.text_content()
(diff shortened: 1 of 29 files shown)


jlucaso1:
@claude review carefully, check how we can be dry, also more performatic and also improve the ergonomics of the API

jlucaso1:
@claude please review carefully again. I want your opinions about the changes, check the claims, and also suggest things that make sense.

PR #521, opened 2026-04-12
fix: suppress ack errors for all transport-unavailable conditions

jlucaso1:
## Summary

- Add `is_transport_unavailable()` to `ClientError` and `EncryptSendError`
- Matches `NotConnected`, `EncryptSend(Transport)`, and `EncryptSend(ChannelClosed)`
- Use in `maybe_deferred_ack` to suppress spurious "Failed to send ack" warnings during disconnects

Previously only `NotConnected` was silenced in the async ack path, and the sync path had no filter at all. Transport pipe breaks and channel drops during disconnect produced noisy warnings.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo test --all --exclude e2e-tests`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of transport unavailability errors to reduce unnecessary warning logs when the network connection is temporarily unavailable.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -271,6 +271,16 @@ pub enum ClientError {
     NotLoggedIn,
 }
 
+impl ClientError {
+    pub fn is_transport_unavailable(&self) -> bool {
+        match self {
+            ClientError::NotConnected => true,
+            ClientError::EncryptSend(e) => e.is_transport_unavailable(),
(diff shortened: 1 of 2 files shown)


jlucaso1:
@claude review check if it is the best performance, DRY, and consistent.

PR #522, opened 2026-04-13
perf!: zero-copy receive pipeline and take-ownership session cache

jlucaso1:
## Summary

- **Zero-copy receive pipeline**: thread `BytesMut`/`Bytes` ownership end-to-end from frame decoding → noise decryption → unpack → node decode, eliminating intermediate `Vec<u8>` copies. Uncompressed payloads go from network buffer to parsed node with zero allocation on the main path.
- **Take-ownership session cache**: `get_session()` removes the entry from cache so encrypt/decrypt can mutate the owned `SessionRecord` directly without cloning. Added `has_session()` to `SessionStore` trait for non-destructive existence checks in the prekey fanout loop.
- **Session restore on error paths**: decrypt functions now always put the session back on failure, matching WA Web behavior (verified against `docs/captured-js/WAWeb/Signal/ProtocolStoreUnifiedApi.js`) where decrypt failures never destroy cached sessions.
- **Minor hot-path wins**: reusable serialization buffer in session flush, inline hex in `generate_message_id()`, `ValueRef::to_node_value()` for ack construction.

## Breaking changes

- `SessionStore` trait gains a required `has_session(&self, &ProtocolAddress) -> Result<bool>` method
- `FrameDecoder::decode_frame()` returns `BytesMut` instead of `Bytes`
- `NoiseSocket::decrypt_frame()` takes `BytesMut` and returns `BytesMut` (was `&[u8]` → `Vec<u8>`)
- `OwnedNodeRef::new()` takes `impl Into<Bytes>` instead of `Vec<u8>`
- `NoiseCipher::decrypt_in_place_with_counter()` is now generic over `aead::Buffer`

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo test --workspace --exclude e2e-tests` (1113 tests pass)

## Summary by CodeRabbit

* **Performance Enhancements**
  * Faster crypto and parsing with fewer copies and more in-place buffer handling
  * Lower memory use via mutable frame buffers and reusable serialization buffers
  * Slightly faster message ID formatting

* **Infrastructure Improvements**
  * Improved session caching with non-destructive existence checks (has_session) and peek support
  * Refined cache checkout/eviction semantics and reduced unnecessary backend loads
  * Adjusted frame-size limit for compatibility

* **Reliability**
  * More consistent session persistence and safer migration paths to reduce stale or lost state

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2522,7 +2522,6 @@ dependencies = [
 name = "wacore-noise"
 version = "0.5.0"
 dependencies = [
- "aead",
  "aes-gcm",
  "anyhow",
  "bytes",
@@ -2662,6 +2661,7 @@ dependencies = [
 name = "whatsapp-rust"
 version = "0.5.0"
 dependencies = [
+ "aead",
  "anyhow",
  "async-channel",
  "async-lock",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -23,6 +23,7 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
 dependencies = [
+ "bytes",
  "crypto-common 0.1.7",
  "generic-array",
 ]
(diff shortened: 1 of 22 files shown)


jlucaso1:
@claude review this carefully, check for more performance improvements too, reduce allocations, being dry. Be very nitpick

jlucaso1:
@claude review again like before, the same points, check if we missed some bug or some performance opportunity.

PR #524, opened 2026-04-13
fix: WA Web-compliant DM multi-device fanout with phash validation

### LINKED ISSUE #508: PDO-only approach for <unavailable> doesn't reliably deliver — add retry receipt

jlucaso1:
## Summary

Reworks DM multi-device encryption to match WA Web's actual architecture, verified against captured JS (`WAWebSendUserMsgJob`, `WAWebDBDeviceListFanout`, `WAWebApiDeviceList`, `WAWebSendMsgCreateFanoutStanza`).

Based on the work in #523 by @mcaldas — encrypts for all known recipient devices instead of bare JID only, fixing "Waiting for this message" on linked devices.

### Changes

- **Local registry first**: read device list from local cache/DB on the send path (WA Web reads IndexedDB only, no network I/O). Network fetch via `get_user_devices` only on cache miss, errors non-fatal — falls back to bare JID (server fanout)
- **Both recipient AND own devices** use the local-first pattern — no hard `.await?` failure on either
- **Hosted device filtering**: exclude device 99 / `@hosted` / `@hosted.lid` from DM fanout (`WAWebDBDeviceListFanout` skips `device.id === 99 || device.isHosted`)
- **Bare JID fallback only when no local record exists** — if record exists but all devices are filtered, don't invent a bare JID target (matches WA Web without biz-coex gate)
- **DM phash validation**: compute `participant_list_hash` and embed in stanza `phash` attr + `DeviceSentMessage.phash` field. Server can now signal device-list drift via ack, and `spawn_phash_validation` invalidates device cache on mismatch for self-healing
- **Avoid LID-migration side effects**: check local cache before calling `get_user_devices` to avoid unnecessary write-heavy usync queries on the hot path

### Not implemented

- **Single-`<enc>` fast path** (`WAWebSendMsgCreateFanoutStanza` line 26): WA Web uses a bare `<enc>` node (no `<participants>` wrapper) when there's exactly 1 primary device. Our `encrypt_for_devices` always wraps in `<to jid=...>` nodes — implementing the fast path would require refactoring the encryption layer. The `<participants>` form is accepted by the server regardless.
- **`isHosted` per-device flag**: WA Web stores `isHosted` per device in IDB records. Our `DeviceInfo` only has `device_id` + `key_index`. We rely on `Jid::is_hosted()` (device 99 + `@hosted` server) which covers the known hosted device patterns.

### WA Web compliance matrix

| Behavior | WA Web | This PR |
|---|---|---|
| Device list source on send path | Local DB only (`getDeviceIds` → IndexedDB) | Local registry only (`get_devices_from_registry`) |
| Network fetch on cache miss | Never on send path (pre-populated via sync) | Best-effort warming, errors non-fatal |
| Hosted device filtering | `device.id === 99 \|\| device.isHosted` | `Jid::is_hosted()` (device 99 + @hosted server) |
| Bare JID fallback | Only when no local record | Only when no local record |
| Phash in DM stanza | Not sent (server returns in ack on mismatch) | Sent in stanza attr + DSM field |
| Phash mismatch handling | `syncDeviceListJob` → resend | `invalidate_device_cache` → next send re-fetches |
| Single-enc fast path | Yes (1 primary device) | No (`<participants>` always) |

## Test plan

- [x] `cargo fmt --all` clean
- [x] All unit tests pass (944 tests, 0 failures)
- [x] E2E test failures are pre-existing (community + lid_sessions), verified identical on base branch
- [ ] Manual test: send DM to user with WhatsApp Web active — message appears on all devices
- [ ] Manual test: send DM with network down on first send — falls back to bare JID gracefully

Closes #523

## Summary by CodeRabbit

* **Bug Fixes**
  * Broader cache invalidation for 1:1 and status sends to keep per-user and sender device caches consistent.
  * Excluded sender device from DM delivery targets and deduplicated targets to prevent self-delivery.

* **Reliability & Performance**
  * DM device discovery now prefers a local registry with fallback and skips hosted devices.
  * Participant hash (phash) is computed from the actual recipient+own device set and preserved through DM preparation; ACK phash handling updated.

* **Compatibility**
  * Added a legacy key migration attempt to reduce decrypt failures and retry receipts.

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -833,27 +833,34 @@ impl Client {
                                     any_duplicate = true;
                                 } else if matches!(retry_err, SignalProtocolError::InvalidPreKeyId)
                                 {
-                                    // InvalidPreKeyId after identity change means the sender is using
-                                    // an old prekey that we no longer have. This typically happens when:
-                                    // 1. The sender reinstalled WhatsApp and cached our old prekey bundle
-                                    // 2. The prekey they're using has been consumed or rotated out
-                                    //
-                                    // Solution: Send a retry receipt with a fresh prekey so the sender
-                                    // can establish a new session and resend the message.
-                                    log::warn!(
-                                        "[msg:{}] Decryption failed for {} due to InvalidPreKeyId after identity change. \
-                                         The sender is using an old prekey we no longer have. \
-                                         Sending retry receipt with fresh keys.",
-                                        info.id,
-                                        address
-                                    );
-
-                                    // Send retry receipt so the sender fetches our new prekey bundle
-                                    dispatched_undecryptable = self.handle_decrypt_failure(
-                                        info,
-                                        RetryReason::InvalidKeyId,
-                                        decrypt_fail_mode,
-                                    );
+                                    // Session may exist under PN address after identity change
+                                    if self
+                                        .try_pn_to_lid_migration_decrypt(
+                                            sender_encryption_jid,
+                                            &signal_address,
+                                            &parsed_message,
+                                            &mut adapter,
+                                            &mut rng,
+                                            &enc_type,
+                                            padding_version,
+                                            info,
+                                        )
+                                        .await
+                                    {
+                                        any_success = true;
+                                    } else {
+                                        log::warn!(
+                                            "[msg:{}] InvalidPreKeyId after identity change for {}. \
+                                             Sending retry receipt with fresh keys.",
+                                            info.id,
+                                            address
+                                        );
+                                        dispatched_undecryptable = self.handle_decrypt_failure(
+                                            info,
+                                            RetryReason::InvalidKeyId,
+                                            decrypt_fail_mode,
+                                        );
+                                    }
                                 } else {
                                     log::error!(
                                         "[msg:{}] Decryption failed even after clearing untrusted identity for {}: {:?}",
--- a/src/message.rs
+++ b/src/message.rs
@@ -951,18 +951,27 @@ impl Client {
                             self.handle_decrypt_failure(info, reason, decrypt_fail_mode);
                         continue;
                     } else if matches!(e, SignalProtocolError::InvalidPreKeyId) {
-                        // InvalidPreKeyId means the sender is using a PreKey ID that we don't have.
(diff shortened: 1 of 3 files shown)


jlucaso1:
@claude review

jlucaso1:
@claude review

jlucaso1:
@claude review

PR #533, opened 2026-04-14
feat!: replace history sync events with lazy blob + perf optimizations

jlucaso1:
## Summary

Replaces `Event::JoinedGroup(LazyConversation)` (per-conversation, misleading name) and dead `Event::HistorySync(HistorySync)` with a single `Event::HistorySync(Box<LazyHistorySync>)` per sync blob, plus performance optimizations.

## Consumer API

```rust
Event::HistorySync(hs) => {
    // Cheap metadata — no decode
    let sync_type = hs.sync_type();
    let progress = hs.progress();

    // Full proto access on demand (lazy, cached, parse-once)
    if let Some(data) = hs.get() {
        for conv in &data.conversations { /* ... */ }
        for pn in &data.pushnames { /* ... */ }
        // global_settings, past_participants, call_log_records, etc.
    }

    // Or raw bytes for custom partial decoding
    let raw = hs.raw_bytes();
}
```

### Event refactor
- `Event::JoinedGroup(LazyConversation)` removed — name was wrong (carried DMs, groups, newsletters)
- `Event::HistorySync(HistorySync)` removed — dead code, never dispatched
- New: `Event::HistorySync(Box<LazyHistorySync>)` — single event per sync blob with lazy decode

### Performance
- **Channel removed** — tctoken candidates collected directly in the streaming loop, no more `async_channel`/oneshot/secondary spawn
- **Conditional blob retention** — decompressed `Bytes` only kept when event listeners exist
- **Inline decode for small blobs** — PushName/Recent syncs (<256KB) skip `spawn_blocking`
- **Manual pushname parser** — checks id before allocating, skips 99% of entries without allocation
- **`parse_jid_fast` early-exit** — groups/newsletters/bots filtered without JID allocation
- **Decompression cap** — 8MB → 64MB estimate to avoid reallocs; hard 64MB limit via `Read::take()` to prevent OOM
- **`Box<wa::HistorySync>` in OnceLock** — avoids ~500 bytes inline reservation
- **Safe Clone** — shares `Bytes` cheaply, creates fresh `OnceLock` (no accidental deep copy)
- **Zero-clone tctoken storage** — candidates consumed by value, token moved directly into `TcTokenEntry`

### Safety
- Hard 64MB decompression limit via `Read::take()` prevents OOM on malformed blobs
- Safe `u64 -> usize` conversion via `try_from` (no truncation on 32-bit/WASM)
- `own_user.unwrap()` eliminated — moved into if-let chain
- Serialize is metadata-only and deterministic (no handler-order dependency)

### What's preserved
- TCToken extraction (prost partial decode of fields 1/21/22/28)
- Own pushname extraction (streaming field 7)
- NCT salt extraction (streaming field 19)
- No-listener fast path (`has_handlers()` gates blob retention and event dispatch)

## Breaking changes
- `Event::JoinedGroup` removed
- `Event::HistorySync` payload changed from `wa::HistorySync` to `Box<LazyHistorySync>`
- `LazyConversation` removed

## Test plan
- [x] All tests pass (7 new LazyHistorySync tests + 3 existing history sync tests)
- [x] No remaining references to `JoinedGroup`, `LazyConversation`, or old channel code

### DIFF
--- a/src/history_sync.rs
+++ b/src/history_sync.rs
@@ -1,25 +1,10 @@
 use crate::types::events::{Event, LazyHistorySync};
-use bytes::Bytes;
 use std::sync::Arc;
-use wacore::history_sync::process_history_sync;
+use wacore::history_sync::{TcTokenCandidate, process_history_sync};
 use wacore::store::traits::TcTokenEntry;
-use wacore_binary::JidExt;
 use waproto::whatsapp::message::HistorySyncNotification;
 
 use crate::client::Client;
-
-/// Partial Conversation decode — only tctoken fields, skips heavy `messages`.
-#[derive(Clone, PartialEq, prost::Message)]
-struct ConversationTcTokenFields {
-    #[prost(string, required, tag = "1")]
-    pub id: String,
-    #[prost(bytes = "vec", optional, tag = "21")]
-    pub tc_token: Option<Vec<u8>>,
-    #[prost(uint64, optional, tag = "22")]
-    pub tc_token_timestamp: Option<u64>,
-    #[prost(uint64, optional, tag = "28")]
-    pub tc_token_sender_timestamp: Option<u64>,
-}
 
 impl Client {
     pub(crate) async fn handle_history_sync(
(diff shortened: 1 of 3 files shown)


jlucaso1:
@claude review

jlucaso1:
@claude review things carefully: performance, unnecessary allocations, what can be dry, also try to find dead code

PR #538, opened 2026-04-14
feat(prekeys): expose public refresh_prekeys() for device migration

aldoeliacim:
## Motivation

The existing `upload_pre_keys*` family is `pub(crate)`, and the only force-upload path (`force=true`) is triggered internally from `validate_digest_key` when the server returns 404 for the digest key bundle. That's correct for the steady state, but leaves external crates no recovery path for a legitimate operational case: **migrating a device from an external Signal store** (e.g., importing a Baileys session into an `InMemoryBackend`, or any other backend that restores historic state without the pre-key private key material).

In that scenario the WhatsApp server still holds pre-key IDs the caller's local store cannot reproduce — any `pkmsg` that references those IDs fails with `InvalidPreKeyId` permanently because the private key material is unrecoverable. Today the caller has no supported way to push a fresh batch so the server hands out IDs the client *does* have locally.

## Change

Adds a thin public wrapper:

```rust
pub async fn refresh_prekeys(&self) -> Result<(), anyhow::Error> {
    self.upload_pre_keys_with_retry(true).await
}
```

No behavioural change on the normal startup path. Callers that never call `refresh_prekeys()` see the exact same flow as before. The retry/backoff semantics (Fibonacci, MAX 610s, disconnect bail) are inherited from `upload_pre_keys_with_retry`.

## Expected usage

After importing an external Signal session, call `refresh_prekeys()` on the first `Event::Connected` of the first process lifetime. Server appends the fresh `WANTED_PRE_KEY_COUNT` IDs (which the caller has locally); old unmatched IDs drain naturally as peers consume them.

## Why not expose `upload_pre_keys_with_retry` directly

Thought about it — the `force: bool` argument is an internal implementation detail (today it's only ever `true` from the digest repair path). Giving it a purpose-named public method keeps the meaning explicit at the call site and leaves room to tighten semantics later (e.g., only firing once per session).

## Context

Being exercised in production against `whatsapp-rust v0.5.0` to complete a Baileys → native migration that currently has 0 successfully decrypted inbound messages despite 530 imported sessions. Patched via `[patch.crates-io]` meanwhile.

Happy to adjust — name, docs, visibility boundary, wherever you'd prefer this to live.

## Summary by CodeRabbit

* **New Features**
  * Added ability to manually trigger a pre-key refresh operation on demand.

### DIFF
--- a/src/prekeys.rs
+++ b/src/prekeys.rs
@@ -248,6 +248,23 @@ impl Client {
         }
     }
 
+    /// Force-refresh the server's one-time pre-key pool with a fresh batch.
+    ///
+    /// Intended for callers that just restored a device from an external source
+    /// (e.g., migrating a Baileys session into an `InMemoryBackend`). The server
+    /// may still hold pre-key IDs whose private key material the caller cannot
+    /// reconstruct; any `pkmsg` referencing those IDs will fail forever with
+    /// `InvalidPreKeyId`. Uploading a fresh batch gives the server new IDs the
+    /// caller *does* have locally, and old unmatched IDs drain as peers consume
+    /// them.
+    ///
+    /// Acquires `prekey_upload_lock` for the duration so this force-upload
+    /// cannot race on `start_id` with the count-based and digest-repair paths.
+    pub async fn refresh_pre_keys(&self) -> Result<(), anyhow::Error> {
+        let _guard = self.prekey_upload_lock.lock().await;
+        self.upload_pre_keys_with_retry(true).await
+    }
+
     /// Validate server key bundle digest, re-uploading only when the server has no record.
     ///
     /// Matches WA Web's `WAWebDigestKeyJob.digestKey()`:


jlucaso1:
@aldoeliacim can you solve the conflicts too? I'll take a look on that soon

aldoeliacim:
Rebased onto `main` (5bd0211) and reworked to match upstream's "lock at scheduling point" pattern instead of "lock inside wrapper."

**Now only touches `src/prekeys.rs` (+17 lines, 0 deletions).** Adds `Client::refresh_pre_keys()` which acquires `prekey_upload_lock` at its own call site and delegates to `upload_pre_keys_with_retry(true)`. `notification.rs` and the existing upload helpers are untouched.

This closes the CodeRabbit-flagged race without divergence from how `upload_pre_keys_at_login` / `handle_prekey_low` / `handle_digest_key` already coordinate. The earlier "move the lock inside the wrapper" approach conflicted with that pattern and created deadlock risk with the outer locks in notification.rs; reverting it.

All previous inline threads addressed in the revision. CI is green locally (`cargo test --lib --tests`).

jlucaso1:
@claude review

jlucaso1:
@codex review

jlucaso1:
@aldoeliacim looks better now. I think we can just simplify the comment. It looks verbose and also has specific things of baileys. Just make the comment objective and agnostic

aldoeliacim:
**Great!** Thank you for your work.

jlucaso1:
I'm the one who should thank you.

PR #539, opened 2026-04-14
feat(wacore-binary): zero-copy Serialize for NodeRef type family

jlucaso1:
## Summary
- Add `serde::Serialize` impls (behind `#[cfg(feature = "serde")]`) for all borrowed/ref types: `NodeStr`, `JidRef`, `ValueRef`, `NodeContentRef`, `NodeRef`, and `OwnedNodeRef`
- Zero allocation during serialization — everything borrows from the yoke buffer via `Deref`/`AsRef`
- Output is byte-identical to the owned type serialization, enabling downstream code to drop `to_owned_node()` clones when serializing `OwnedNodeRef`

## Motivation
`OwnedNodeRef` exists to avoid cloning decoded nodes, but serialization previously required `to_owned_node()` which defeats the purpose. With these impls, consumers like `whatsapp-rust-bridge` can serialize directly from the yoke buffer:

```rust
// Before: allocates cloned Node
let val = serde_json::to_value(&node.to_owned_node())?;

// After: zero-copy from yoke buffer
let val = serde_json::to_value(node.get())?;
```

## Test plan
- [x] `cargo clippy --all --tests` — clean
- [x] `cargo test -p wacore-binary --features serde -- serde_tests` — 5 parity tests pass
  - `node_ref_serializes_same_as_node` — attrs with String + Jid values, String content
  - `nested_nodes_serialize_same` — recursive Nodes content
  - `bytes_content_serializes_same` — Bytes content
  - `value_ref_matches_node_value` — both String and Jid variants
  - `owned_node_ref_serializes_same_as_owned` — full round-trip through marshal/unmarshal

## Summary by CodeRabbit

* **New Features**
  * Optional Serde-based JSON serialization for core data structures (JID, node/value/content types, owned and borrowed node representations).

* **Tests**
  * Added Serde unit tests to verify JSON equivalence between owned and borrowed representations, including nested nodes and byte/string variants.

### DIFF
--- a/wacore/binary/Cargo.toml
+++ b/wacore/binary/Cargo.toml
@@ -34,6 +34,7 @@ serde_json = { workspace = true, features = ["std"] }
 
 [dev-dependencies]
 iai-callgrind = { workspace = true }
+serde_json = { workspace = true, features = ["std"] }
 
 [[bench]]
 name = "binary_benchmark"
--- a/wacore/binary/src/jid.rs
+++ b/wacore/binary/src/jid.rs
@@ -372,20 +372,6 @@ pub struct Jid {
     pub integrator: u16,
 }
 
-#[cfg(feature = "serde")]
-impl serde::Serialize for JidRef<'_> {
-    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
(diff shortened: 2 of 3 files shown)


jlucaso1:
@claude review

PR #542, opened 2026-04-14
perf: skip string classification for strings longer than PACKED_MAX

jlucaso1:
## Summary
- Strings > `PACKED_MAX` (127 bytes) can never be protocol tokens (max 48), packed nibble/hex, or JIDs — yet the encoder was running two full SipHash + PHF map lookups per string that always missed
- Add a length guard in `write_string_uncached` and `hint_or_insert` to short-circuit directly to raw-bytes encoding, skipping classification entirely

## Benchmark results (iai-callgrind, instruction counts)

| Benchmark | Before | After | Delta |
|-----------|--------|-------|-------|
| `marshal_auto_long_string` | 15,599 | 10,508 | **-32.6%** |
| `marshal_long_string` | 15,555 | 10,464 | **-32.7%** |
| `marshal_exact_long_string` | 17,335 | 12,163 | **-29.8%** |
| `marshal_allocating` | 91,389 | 91,887 | +0.54% |
| `marshal_many_children` | 13,204,384 | 13,262,569 | +0.44% |
| decode/attr/jid benchmarks | — | — | unchanged |

The +0.5% on short-string-only benchmarks is the cost of the added branch (~6 instructions per `write_string` call). For any message with content > 127 bytes the net improvement is thousands of instructions.

## Test plan
- [ ] CI passes (existing encoder tests + roundtrip tests cover correctness)
- [ ] Benchmark CI confirms instruction count improvements

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of very long strings during encoding to avoid incorrect classification and ensure stable, efficient encoding and correct decoding at length boundaries.

* **Tests**
  * Added test coverage verifying proper encoding/decoding behavior for boundary-length and over-boundary strings.

### DIFF
--- a/wacore/binary/src/encoder.rs
+++ b/wacore/binary/src/encoder.rs
@@ -1423,4 +1423,73 @@ mod tests {
 
         Ok(())
     }
+
+    /// Regression test: strings at the PACKED_MAX boundary must be classified
+    /// normally, while strings above it must be emitted as raw bytes (skipping
+    /// SipHash/PHF classification entirely).
+    #[test]
+    fn test_long_string_skips_classification() -> TestResult {
+        use crate::decoder::Decoder;
+        use crate::marshal::marshal;
+
+        let at_boundary = "0".repeat(token::PACKED_MAX as usize); // 127 nibble chars
+        let over_boundary = "0".repeat(token::PACKED_MAX as usize + 1); // 128 chars
+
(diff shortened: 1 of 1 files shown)


jlucaso1:
@claude review nitpicky for performance and allocation, also check future improvement to be in a follow up

PR #543, opened 2026-04-14
perf: unify single/double-byte token maps into single PHF lookup

jlucaso1:
## Summary
- The encoder was doing **two** separate SipHash + PHF lookups per string (single-byte map miss → double-byte map). Since there is zero overlap between the 235 single-byte and 1024 double-byte tokens, this merges them into a single `phf::Map<&str, TokenKind>` so every string is hashed exactly once.
- Also benefits `intern_cow` (the decode-to-owned path) which had the same two-lookup pattern.

## Benchmark results (iai-callgrind, instruction counts vs main)

| Benchmark | Before | After | Delta |
|-----------|--------|-------|-------|
| `marshal_allocating` | 91,389 | 83,215 | **-8.9%** |
| `marshal_auto_allocating` | 91,422 | 83,248 | **-8.9%** |
| `marshal_many_children` | 13,204,384 | 10,831,392 | **-18.0%** |
| `marshal_auto_many_children` | 13,171,888 | 10,798,836 | **-18.4%** |
| `marshal_auto_long_string` | 15,599 | 11,675 | **-25.2%** |
| `roundtrip large` | 86,038 | 78,276 | **-9.0%** |
| `jid_to_owned_access` | 13,307 | 13,115 | **-1.4%** |
| decode / unmarshal / unpack | — | — | unchanged |

Stacks with #542 (PACKED_MAX fast-path) for additional gains on long strings.

## Test plan
- [ ] CI passes (all 74 wacore-binary tests pass locally, including updated token roundtrip tests)
- [ ] Benchmark CI confirms instruction count improvements

## Summary by CodeRabbit

* **Refactor**
  * Internally unified token handling and lookup into a single, simpler mechanism; observable behavior remains unchanged.
* **Tests**
  * Updated unit tests to exercise the unified token lookup and validate existing behavior.

### DIFF
--- a/wacore/binary/build.rs
+++ b/wacore/binary/build.rs
@@ -21,45 +21,40 @@ fn main() {
     let tokens_json = fs::read_to_string("src/tokens.json").unwrap();
     let tokens: Tokens = serde_json::from_str(&tokens_json).unwrap();
 
-    let mut single_byte_map = Map::new();
-    let single_byte_values: Vec<String> = (0..tokens.single_byte.len())
-        .map(|i| i.to_string())
-        .collect();
+    // Unified token map: single lookup for both single-byte and double-byte tokens.
+    // TokenKind::Single(u8) for single-byte, TokenKind::Double(u8, u8) for double-byte.
+    let mut unified_map = Map::new();
+    let mut values = Vec::new();
+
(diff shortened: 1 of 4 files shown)


jlucaso1:
@claude review nitpicky

PR #558, opened 2026-04-16
fix(pair-code): accept notification without link_code_companion_reg wrapper

AsafFisher:
## What

`handle_pair_code_notification` in `src/pair_code.rs` currently bails with `return false` when the incoming notification has no `<link_code_companion_reg>` child. In the wild, the server sometimes sends the same payload with the key fields as direct children of the `<notification>` node (no wrapper), and the current code silently drops those.

User-visible effect: the first pair-code attempt often hangs — user enters the 8-character code on their phone, sees nothing happen, retries, and sometimes the retry arrives wrapped and works.

## Fix

Fall back to the notification node itself when the wrapper child is absent:

```rust
let reg_node = node
    .get_optional_child_by_tag(&[\"link_code_companion_reg\"])
    .unwrap_or(node);
```

Backward-compatible — wrapped payloads still hit the same code path through the child lookup.

## Testing

- Observed in production that first pair-code attempts frequently fail on our deployment; retries often succeed.
- Applied this patch locally against `jlucaso1/whatsapp-rust@88a5145` (current main), ran `cargo check -p whatsapp-rust` clean.
- Ran a fresh pair-code flow end-to-end against a real WhatsApp account with the patch in place; pairing completed successfully (log output: `Successfully authenticated with WhatsApp servers! (gen=1)` followed by normal history sync).

## Context

We've been carrying this as a vendored patch for a few months. Upstreaming so we can drop it and track `main` directly. Happy to iterate on the fix shape if you'd prefer the fallback handled elsewhere (e.g. inside `get_optional_child_by_tag` or with a separate shape-detection arm).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience in data parsing by gracefully handling cases where optional wrapper elements are absent from server responses.

### DIFF
--- a/src/pair_code.rs
+++ b/src/pair_code.rs
@@ -229,10 +229,14 @@ pub(crate) async fn handle_pair_code_notification(
     client: &Arc<Client>,
     node: &NodeRef<'_>,
 ) -> bool {
-    // Check if this is a link_code_companion_reg notification
-    let Some(reg_node) = node.get_optional_child_by_tag(&["link_code_companion_reg"]) else {
-        return false;
-    };
+    // The data fields may be wrapped in a child <link_code_companion_reg> node,
+    // or they may be direct children of the notification node itself (when the
+    // notification type="link_code_companion_reg"). Observed in the wild that
+    // the server sometimes omits the wrapper; without this fallback the
+    // pair-code flow silently hangs on the first attempt.
+    let reg_node = node
+        .get_optional_child_by_tag(&["link_code_companion_reg"])
+        .unwrap_or(node);
 
     // Extract primary's wrapped ephemeral public key (80 bytes: salt + iv + encrypted key)
     let primary_wrapped_ephemeral = match reg_node


jlucaso1:
Thanks for taking the time to upstream this, and for the detailed context on what you've been seeing in production.

Before I merge, I want to make sure we're fixing the right thing. I spent some time cross-referencing the three implementations I usually check against, and all three strictly require the `<link_code_companion_reg>` wrapper:

- **WA Web** (captured JS, `WASmaxInMdPrimaryHelloNotifyCompanionRequest`): uses `flattenedChildWithTag(e, "link_code_companion_reg")`, which returns a hard parse error ("to have 1 child `<link_code_companion_reg>`, but found 0") when the wrapper is missing. No fallback.
- **whatsmeow** ([pair-code.go#L150-L157](https://github.com/tulir/whatsmeow/blob/3d82d3101dd1cd904542ddca1a7970e481bad97d/pair-code.go#L150-L157)): returns `ElementMissingError` when the wrapper is absent.
- **Baileys** ([messages-recv.ts#L787-L797](https://github.com/WhiskeySockets/Baileys/blob/c4e5d1262aa143dd4f3c2df94ca5537b8db2b3d7/src/Socket/messages-recv.ts#L787-L797) plus `toRequiredBuffer` at [#L1002-L1008](https://github.com/WhiskeySockets/Baileys/blob/c4e5d1262aa143dd4f3c2df94ca5537b8db2b3d7/src/Socket/messages-recv.ts#L1002-L1008)): throws a 400 Boom error when the wrapper is absent.

If the server were really sending the unwrapped shape in the wild, Baileys would be throwing 400s and whatsmeow would be logging `ElementMissingError` for a lot of users, and I'd expect to see public reports about it. I haven't been able to find any. That makes me want to rule out a different root cause before we diverge from WA Web here.

Could you help me confirm the premise in two ways?

**1. A log of the actual stanza when the fallback fires.**

Could you add a temporary `warn!` (or `info!`) right before `reg_node.unwrap_or(node)` that prints the full notification node? Something like:

```rust
if node.get_optional_child_by_tag(&["link_code_companion_reg"]).is_none() {
    warn!(target: "Client/PairCode", "pair-code notification without wrapper: {:?}", node);
}
```

Then reproduce the "first attempt hangs" case and share the captured node (base64 the raw bytes or the debug-printed tree is fine). If the wrapper really is absent, that's a protocol finding worth reporting upstream to whatsmeow/Baileys as well. If the wrapper is actually present, we're chasing the wrong bug.

**2. A pointer to WA Web handling the unwrapped shape.**

If you've seen WA Web itself accept the unwrapped form, a reference would help a lot. The captured JS I have only shows the strict parser, but captures can lag the live client. Any pointer (a different module name, a different notification `type` attribute value, anything) would let me match the fix to real WA Web behavior.

**One alternative hypothesis worth ruling out.**

In `pair_with_code` (`src/pair_code.rs:191-212`), we `await send_iq(...)` and only set `pair_code_state = WaitingForPhoneConfirmation` after it returns. If the user is fast enough on their phone, the `link_code_companion_reg` notification can arrive before the state is written, and the handler hits the `_ => warn!("not in waiting state")` arm and returns false silently. That matches your "first attempt hangs, retry works" symptom fairly cleanly (the retry gives the state a chance to be set before the notification arrives). The fallback in this PR wouldn't help that case directly, but a retry would naturally succeed regardless. Worth checking your logs for `"Received pair code notification but not in waiting state"` around the failed attempts.

Happy to land something here once we know what's actually going on. If the logs confirm the unwrapped shape, I'd want to at least gate the fallback behind a `warn!` so we can quantify how often it fires, and ideally also add the `link_code_pairing_ref` validation that whatsmeow and Baileys both do (the current handler doesn't check it at all). If the logs point at the state-race instead, we can fix that properly and drop the fallback.

Thanks again for the patience. Not trying to make this harder than it needs to be, just want to make sure we don't paper over something that matters.

AsafFisher:
Closing this — after 3 successful pair-code attempts with diagnostic logging, the wrapper was always present (`children_count=1`). We couldn't reproduce the flat-notification shape.

The state-race hypothesis you raised (notification arriving before `WaitingForPhoneConfirmation` is set) is more plausible for the "first attempt hangs" behavior we saw in production. We'll investigate that path instead.

Thanks for the thorough review and the pointer to the race condition — saved us from papering over the wrong thing.

jlucaso1:
You're welcome

PR #559, opened 2026-04-17
WA Web compliance: retry flow + group/status session rebuild

jlucaso1:
## Summary

Aligns the retry receipt handler with WhatsApp Web's `WAWebUpdateLocalSignalSession` + `ensureE2ESessions` flow (`docs/captured-js/WAWeb/Update/LocalSignalSession.js`, `WAWeb/Handle/RetryRequest.js:199-200`).

### 1. `libsignal: demote NoSenderKeyState log from ERROR to DEBUG`
Unknown chain IDs are normal (sender key rotation, re-registration, out-of-order SKDM). The caller already converts the typed error into a retry receipt and logs at debug level with richer context. WA Web uses the `SenderKeyExpired` WAM event instead of error-level logging. Eliminates double-logging and false alerts in production observability tools.

### 2. `retry: ensure E2E sessions before group/status resend`
Fixes a concrete bug observed in production logs (debug-1776271138): when a retry receipt triggered reg-ID-mismatch session deletion or arrived from a newly-added device, `prepare_group_retry_stanza` failed with `session not found` and the duplicate-handler filter silenced all subsequent retries, permanently losing the message for that device. WA Web calls `ensureE2ESessions([g])` for all chat types (RetryRequest.js:200); the DM branch already did, the group/status branch didn't. Now it does.

### 3. `retry: mirror updateLocalSignalSession fully for all chat types`
Replaces the DM-only `delete_dm_retry_session_target` + inline processKeyBundle with a single helper mirroring WA Web semantics for DM/group/status:
- `markForgetSenderKey` for group/status
- `processKeyBundle` for all chats (was gated behind `!is_status_broadcast()`)
- reg-ID-mismatch delete for all chats (was DM-only)
- retry==2 saves base key, does NOT delete
- retry>2 with same base key deletes
- retry==1 / retry>2 with different base key: session preserved

**Behavioral delta:** DM retry #1 no longer destroys healthy sessions unconditionally. Group/status now benefit from base-key collision detection.

### 4. `retry: remove retried_group_messages dedupe to match WA Web`
WA Web has no dedupe layer for retry receipts. `WAWeb/Message/Queue.js` (`onMessageQueue`) uses a `PromiseQueueMap` that serializes per-chat but lets every receipt through. Loop prevention is covered by `MAX_RETRY_COUNT` (== WA Web's `MAX_RETRY = 5`), and `pending_retries` guards against concurrent races on session state. The extra dedupe layer was silently dropping retries #2+, which broke the WA Web-compliant recovery path that relies on driving retry_count to 3+ to trigger base-key collision. Removed the cache field, config, and stats; replaced `build_retry_dedupe_key` with `build_retry_processing_key` used only for `pending_retries`.

### 5. `retry: harden update_local_signal_session tests with real SessionRecords`
Follow-up to review feedback. The original tests seeded backend sessions with `b"invalid-session"`, which `peek_session` swallowed via `.ok().flatten()`. Every session-touching branch was skipped, so the tests verified the function didn't panic but did NOT verify the "retry #1 preserves a valid session" rule the names promised. Added a helper that builds and serializes a real `SessionRecord` via `SessionState::from_session_structure`, so the asserted branches actually run. Renamed `handles_regid_mismatch_gracefully` to `deletes_on_regid_mismatch` and made it verify the real delete branch; added `handles_unparseable_session_gracefully` to preserve coverage of the garbage-bytes short-circuit.

## WA Web references

- `WAWeb/Update/LocalSignalSession.js` — session management flow mirrored in commit 3
- `WAWeb/Handle/RetryRequest.js:199-200` — `updateLocalSignalSession` + `ensureE2ESessions([g])` for every chat type (commit 2)
- `WAWeb/Message/Queue.js` — per-chat serialization without dedupe (commit 4)
- `WAWeb/Sender/KeyExpiredWamEvent.js` — telemetry-only path for unknown chain IDs (commit 1 rationale)

## Test plan
- [x] `cargo clippy --all --tests` clean on every commit
- [x] `cargo test --workspace --exclude e2e-tests` — all suites pass
- [x] `cargo test -p e2e-tests --test retry_dm_multidevice` — passes locally with mock server
- [x] Unit tests (per commit):
  - `unknown_chain_id_returns_no_sender_key_state` — typed error path guard
  - `ensure_e2e_sessions_resolved_is_noop_when_session_exists` — regression guard against accidental network calls
  - `update_local_signal_session_preserves_dm_session_at_retry_1` — real SessionRecord; verifies no-delete at retry #1
  - `update_local_signal_session_preserves_group_session_at_retry_1` — same rule for group/status
  - `update_local_signal_session_deletes_on_regid_mismatch` — real SessionRecord; verifies the WA Web delete branch fires
  - `update_local_signal_session_handles_unparseable_session_gracefully` — garbage-bytes short-circuit
  - `update_local_signal_session_no_session_is_noop` — fast path for never-messaged devices
  - `retry_processing_key_per_participant` — serializes same-participant retries, different participants don't block each other

## Notes
- Two previous DM tests (`dm_retry_deletes_only_requested_session`, `dm_retry_deletes_resolved_session_without_registry_devices`) validated the old unconditional-delete behavior and were replaced with compliance-preserving equivalents in commit 3.
- Status broadcast retry short-circuit (no resend) preserved — WA Web does resend status, but requires audience reconstruction the Rust client doesn't maintain yet.
- Pre-existing gaps intentionally out of scope:
  - `processKeyBundle`'s `isOffline` flag is not plumbed through the Rust side
  - Peer-device reg-ID mismatch: WA Web throws to abort `updateLocalSignalSession` entirely; Rust throws but the outer flow still runs the reg-ID-mismatch delete branch

### DIFF
--- a/src/cache_config.rs
+++ b/src/cache_config.rs
@@ -163,8 +163,6 @@ pub struct CacheConfig {
     pub device_registry_cache: CacheEntryConfig,
     /// LID-to-phone cache (time_to_idle). Default: 1h timeout, 10000 entries.
     pub lid_pn_cache: CacheEntryConfig,
-    /// Retried group messages tracker (time_to_live). Default: 5m TTL, 2000 entries.
-    pub retried_group_messages: CacheEntryConfig,
     /// Optional L1 in-memory cache for sent messages (retry support).
     /// Default: capacity 0 (disabled — DB-only, matching WA Web).
     /// Set capacity > 0 to enable a fast in-memory cache in front of the DB.
@@ -208,7 +206,6 @@ impl std::fmt::Debug for CacheConfig {
             .field("group_cache", &self.group_cache)
             .field("device_registry_cache", &self.device_registry_cache)
             .field("lid_pn_cache", &self.lid_pn_cache)
-            .field("retried_group_messages", &self.retried_group_messages)
             .field("recent_messages", &self.recent_messages)
             .field("message_retry_counts", &self.message_retry_counts)
             .field("pdo_pending_requests", &self.pdo_pending_requests)
@@ -241,7 +238,6 @@ impl Default for CacheConfig {
             group_cache: CacheEntryConfig::new(one_hour, 250),
             device_registry_cache: CacheEntryConfig::new(one_hour, 5_000),
             lid_pn_cache: CacheEntryConfig::new(one_hour, 10_000),
-            retried_group_messages: CacheEntryConfig::new(five_min, 2_000),
             recent_messages: CacheEntryConfig::new(five_min, 0),
             message_retry_counts: CacheEntryConfig::new(five_min, 1_000),
             pdo_pending_requests: CacheEntryConfig::new(Some(Duration::from_secs(30)), 500),
--- a/src/client.rs
+++ b/src/client.rs
@@ -171,7 +171,6 @@ pub struct MemoryDiagnostics {
     pub device_registry_cache: u64,
     pub lid_pn_lid_entries: u64,
     pub lid_pn_pn_entries: u64,
-    pub retried_group_messages: u64,
     pub recent_messages: u64,
     pub sender_key_device_cache: u64,
     pub message_retry_counts: u64,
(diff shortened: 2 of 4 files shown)


jlucaso1:
@claude review

jlucaso1:
@claude review

PR #560, opened 2026-04-17
Graceful shutdown for detached tasks (fixes WASM setTimeout leaks)

jlucaso1:
## Summary

Two `.detach()`-ed background tasks held pending `sleep` calls that kept WASM event loops alive for minutes after `client.disconnect()`: the background saver (`sleep(30s)`) and the offline device-sync flush (`sleep(2s)`). Each `sleep` compiles to `setTimeout` on the Node bridge, and detached tasks are never dropped, so the timers never `clearTimeout`.

This PR adds a proper shutdown signal abstraction, ties task lifetimes to their owning `Arc<Client>`, and guarantees a final flush when the saver exits.

Benefits apply to every platform, not just WASM: Tokio servers get a final flush on disconnect (resolves a long-standing "no final flush on drop" TODO), saver tasks stop promptly on disconnect instead of lingering until natural completion, and `Bot` drop cleanly aborts the saver even if `disconnect()` was never called.

### `wacore::runtime::ShutdownNotifier` / `ShutdownSignal`
A publisher/subscriber pair with a **sticky flag** backed by an `AtomicBool` alongside an `event_listener::Event`:

- Publisher (`notify()`) stores the flag **before** waking listeners.
- Subscribers (`wait_for_shutdown` / `ShutdownNotifier::listen()`) register the listener **before** loading the flag.

The ordering guarantees a late subscriber still observes the shutdown — `event_listener` notifications are edge-triggered on their own, so naive `listen()`-then-check races. `ShutdownSignal::never()` is an inert handle for tests and callers that don't wire a real notifier.

### Two shutdown scopes on `Client`
- `shutdown_notifier: ShutdownNotifier` — **terminal**, fired only by `disconnect()`. Long-lived subscribers that outlive reconnects (saver, device-registry cleanup) subscribe via `shutdown_signal()`.
- `connection_shutdown: Mutex<ShutdownNotifier>` — **per-connection**, fired on `cleanup_connection_state`, `<xmlstreamend/>`, stream error, and `handle_connect_failure`. Replaced with a fresh notifier at the start of each new connection so reconnect subscribers see a clean signal. Per-connection tasks (keepalive, request waiters, `read_messages_loop`, offline flush) subscribe via `connection_shutdown_signal()`.

`disconnect()` propagates through both (terminal directly; per-connection via `cleanup_connection_state`).

### Saver lifecycle tied to `Arc<Client>`
`PersistenceManager::run_background_saver` takes `ShutdownSignal` and returns `AbortHandle`. `Bot::build` stores the handle in `Client::saver_handle: OnceLock<AbortHandle>` so the task lifetime follows the last `Arc<Client>` strong ref. The saver self-terminates on shutdown notify (after a final flush); if the last `Arc<Client>` drops without a notify, the `AbortHandle::drop` aborts the task.

### Race-mitigation details
- Saver performs an initial `save_to_disk()` before entering the `select!` loop, covering state dirtied during `Bot::build` construction.
- Saver does not flip `saver_halted` when a failing flush happens on the shutdown path — a terminating task shouldn't permanently mark the store degraded.
- Extracted `wait_for_shutdown` helper dedupes three identical patterns (saver wrapper, wacore saver, `ib.rs` offline flush).

## WASM impact

Measured: Rust-origin `setTimeout`s pending after `client.disconnect()` dropped from **~47 (worst case 180s)** to **0**. Only undici keep-alive (~4s, Node layer) remains.

## Commits

| Commit | Change |
| --- | --- |
| `d87a648` | baseline: shutdown signal + saver `AbortHandle` + `ib.rs` |
| `1d917ea` | wacore: same shutdown pattern in runtime-agnostic saver |
| `c9c9164` | saver: initial flush to close reorder race during `Bot::build` |
| `751f845` | runtime: extract `wait_for_shutdown` helper, dedupe three callsites |
| `f52b794` | saver: `should_exit` as select expression value |
| `63353aa` | saver: don't halt on shutdown-path flush failure |
| `5bf8a28` | saver: restore startup/drop/shutdown debug logs |
| `a95adbc` | shutdown: document shutdown-before-listen race (later obsoleted by sticky flag) |
| `4b35db4` | runtime: `ShutdownSignal` newtype |
| `000aef4` | saver: initial unit tests |
| `6b3246e` | **fix (Codex P2)**: tie saver handle to `Arc<Client>` via `OnceLock`, not `Bot` |
| `d131cf5` | **fix (CodeRabbit critical)**: sticky shutdown flag so notify before listen is observed |
| `5562983` | **fix (CI breakage)**: split per-connection vs terminal shutdown so reconnect works |
| `7f063df` | docs/cleanup: refresh stale race comments, trim dead code |
| `1496a1d` | test: strengthen saver exit test with observable runtime Arc count |

## Test plan
- [x] `cargo clippy --all --tests` clean on every commit
- [x] `cargo test --workspace --exclude e2e-tests` — all suites pass
- [x] `cargo test -p e2e-tests --test app_state test_push_name_survives_reconnect` — one of the eight tests broken by the pre-split sticky flag now passes
- [x] Regression tests added:
  - `wacore::runtime::shutdown_tests::*` — sticky-flag ordering (4 tests)
  - `saver_flushes_and_exits_on_shutdown` — final flush + clean exit
  - `saver_exits_when_abort_handle_dropped_without_signal` — observes runtime Arc count
  - `abort_handle_in_arc_drops_only_when_last_ref_released` — Codex-fix drop semantics
  - `per_connection_notify_leaves_terminal_signal_untouched` — CI-fix scope split
  - `reset_gives_fresh_per_connection_notifier` — reset semantics
  - `terminal_disconnect_propagates_to_per_connection_signal` — disconnect wakes both
- [ ] Full e2e suite (mock server required) — reviewer should run before merge

## Notes
- Pre-existing `AbortHandle` in `wacore::runtime` has `#[must_use]`; callers can't accidentally drop the saver handle on the floor.
- `ShutdownSignal::never()` for opt-out or tests; `wait_for_shutdown` composes in `select!` against other exit conditions.
- `event_listener` stays internal; only `ShutdownNotifier` / `ShutdownSignal` are in the public API.
- `disconnect()` still does an explicit `persistence_manager.flush()` as defense-in-depth alongside the saver's own final flush.

### DIFF
--- a/src/bot.rs
+++ b/src/bot.rs
@@ -147,8 +147,6 @@ pub struct Bot {
     sync_task_receiver: Option<async_channel::Receiver<crate::sync_task::MajorSyncTask>>,
     event_handler: Option<EventHandlerCallback>,
     pair_code_options: Option<PairCodeOptions>,
-    /// Held for its Drop: aborts the saver task when the Bot is dropped.
-    _saver_handle: wacore::runtime::AbortHandle,
 }
 
 impl std::fmt::Debug for Bot {
@@ -692,6 +690,10 @@ impl BotBuilder<Provided, Provided, Provided, Provided> {
             std::time::Duration::from_secs(30),
             client.shutdown_signal(),
         );
+        // Tie the saver task to Arc<Client> so extracting client() and outliving
+        // Bot keeps periodic persistence alive. Client::drop on the last Arc
+        // drops the AbortHandle and aborts the task.
+        let _ = client.saver_handle.set(saver_handle);
 
         // Register custom enc handlers
         for (enc_type, handler) in self.custom_enc_handlers {
@@ -711,7 +713,6 @@ impl BotBuilder<Provided, Provided, Provided, Provided> {
             sync_task_receiver: Some(sync_task_receiver),
             event_handler: self.event_handler,
             pair_code_options: self.pair_code_options,
-            _saver_handle: saver_handle,
         })
     }
 }
--- a/src/bot.rs
+++ b/src/bot.rs
@@ -147,6 +147,8 @@ pub struct Bot {
     sync_task_receiver: Option<async_channel::Receiver<crate::sync_task::MajorSyncTask>>,
     event_handler: Option<EventHandlerCallback>,
     pair_code_options: Option<PairCodeOptions>,
+    /// Held for its Drop: aborts the saver task when the Bot is dropped.
+    _saver_handle: wacore::runtime::AbortHandle,
 }
 
 impl std::fmt::Debug for Bot {
@@ -652,10 +654,6 @@ impl BotBuilder<Provided, Provided, Provided, Provided> {
                 .map_err(|e| anyhow::anyhow!("Failed to create persistence manager: {}", e))?,
         );
 
-        persistence_manager
-            .clone()
-            .run_background_saver(runtime.clone(), std::time::Duration::from_secs(30));
-
         // Apply initial push name if specified (for deterministic mock server phone assignment)
         if let Some(name) = self.initial_push_name {
             persistence_manager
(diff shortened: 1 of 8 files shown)


jlucaso1:
@claude review

PR #563, opened 2026-04-18
fix: status reactions via send_message with correct E2EE wire format

jlucaso1:
## Summary

Three related fixes that together let a bot react to someone's status via the natural `client.send_message(chat, reaction)` API.

### Bug 1 — `Status::send_reaction` emitted the newsletter wire format

`Status::send_reaction(owner, server_id, emoji)` built a plaintext `<message type="reaction"><reaction code="…"/></message>` with a `server_id` attribute. That's the **newsletter** reaction format (`WASmaxOutMessagePublishNewsletterQuestionResponsePublishOrReactionOrReactionRevokeOrPollVoteMixinGroup`). For status reactions — and for every non-newsletter reaction — WA Web encrypts a `ReactionMessage` protobuf via the same path as DM/group reactions (`WAWeb/Send/ReactionMsgAction.js:13-65` + `WAWeb/Reactions/MsgActionUtils.js:85-99`). The server silently drops the wrong format, and the bot's typical `let _ = client.status().send_reaction(...).await` discards the error — so this failed as a no-op in production.

Method removed. `Newsletter::send_reaction` keeps `send_server_reaction` which is the correct format for newsletters.

### Bug 2 — `send_message_impl` blocked every send to `status@broadcast`

The early return `if to.is_status_broadcast() { return Err("Use send_status_message() …") }` was the correct guard for posting new statuses, but too wide: status **reactions** must go out as `to=status@broadcast` too (confirmed via `WAWeb/Send/MsgCreateFanoutStanza.js:436` where `to = CHAT_JID(addon.data.to)` and `addon.data.to = status@broadcast` from `msgKeyToTargetInfo`, see `WAWeb/Msg/KeyUtils.js:54-79`).

When a reaction hits this path, `send_message_impl` now:

- Extracts the status author from `reaction_message.key.participant` (required; returns a clear error if missing).
- Rebinds `to = author_jid` so the existing DM fanout path (pairwise Signal to each of the author's devices, `<enc type="msg">/<enc type="pkmsg">`) runs unchanged.
- Rewrites the outer stanza `to="status@broadcast"` just before send so the wire matches WA Web.

Two downstream gates kept the fix from working correctly by itself:

- `maybe_include_tc_token` already skips `status@broadcast` internally, but since the rebind happened upstream of the check, the skip never fired. Gated by `is_status_addon` at the call site.
- `add_recent_message` was keyed by the rebound author JID. Retry receipts for status reactions arrive with `from=status@broadcast` (same pattern as the receipts our bot already emits for inbound status at `src/receipt.rs:391`), so `take_recent_message(&info.chat, …)` would miss. Now caches under the original broadcast chat when `is_status_addon`.

### Bug 3 — `MessageContext::message_key()` helper

Bot-side mistake that the old broken API masked: constructing `wa::MessageKey` with `participant = None` for statuses (since `is_group = false` for `status@broadcast`). `MessageContext::message_key()` now returns a correctly-shaped key for reactions/revokes, populating `participant` for groups AND status.

## Consumer-visible change

Before (broken): `client.status().send_reaction(owner, server_id, "❤️").await`.

After (uniform across chat types):

```rust
let reaction = wa::Message {
    reaction_message: Some(wa::message::ReactionMessage {
        key: Some(ctx.message_key()),
        text: Some("❤️".into()),
        sender_timestamp_ms: Some(Utc::now().timestamp_millis()),
        ..Default::default()
    }),
    ..Default::default()
};
ctx.send_message(reaction).await?;
```

## Evidence sources

- `docs/captured-js/WAWeb/Send/ReactionMsgAction.js:13-65` — reaction path delegates to newsletter only when chat is newsletter; otherwise builds `ReactionMessage` via `addAndSendAddonToChat`.
- `docs/captured-js/WAWeb/Reactions/MsgActionUtils.js:85-99` — returns `{type: REACTION, to: status@broadcast, …}` for status.
- `docs/captured-js/WAWeb/Msg/KeyUtils.js:54-79` — `msgKeyToTargetInfo` produces `{from: me, to: status@broadcast, author, broadcastId}` for status reactions.
- `docs/captured-js/WAWeb/Send/MsgCreateFanoutStanza.js:25-67, 436` — pairwise Signal fanout path, `to = CHAT_JID(addon.data.to)`, `<enc type="msg"/pkmsg>`, conditional `<device-identity>`, no `<meta content_type="add_on"/>` on outgoing.
- User's real bot log: inbound status reactions decode as `<enc type="msg">` with server-added `<meta content_type="add_on"/>`; no outgoing reaction ever carries that meta.

## Non-goals / not in scope

- Newsletter reaction path (unchanged, already correct).
- Receiving status reaction events with richer metadata (current receive path already works).
- A typed `Status::react` helper — the natural `ctx.send_message(reaction)` path does the job now.

## Test plan

- [ ] `cargo test -p whatsapp-rust --lib send::tests::send_message_to_status` (2 regression cases on the validation paths)
- [ ] `cargo clippy --workspace --exclude e2e-tests --tests`
- [ ] Manual check with a live bot reacting to a posted status (outside the CI scope)

### DIFF
--- a/src/bot.rs
+++ b/src/bot.rs
@@ -67,8 +67,8 @@ impl MessageContext {
         )
     }
 
-    /// [`wa::MessageKey`] referencing this message. `participant` is populated
-    /// for groups and status so reactions/revokes can identify the author.
+    /// Referential [`wa::MessageKey`] for [`wa::message::ReactionMessage::key`].
+    /// Sender-side revokes have a different shape; use [`Client::revoke_message`].
     pub fn message_key(&self) -> wa::MessageKey {
         use wacore_binary::JidExt;
         let needs_participant =
--- a/src/bot.rs
+++ b/src/bot.rs
@@ -67,6 +67,20 @@ impl MessageContext {
         )
     }
 
+    /// [`wa::MessageKey`] referencing this message. `participant` is populated
+    /// for groups and status so reactions/revokes can identify the author.
+    pub fn message_key(&self) -> wa::MessageKey {
+        use wacore_binary::JidExt;
+        let needs_participant =
+            self.info.source.is_group || self.info.source.chat.is_status_broadcast();
+        wa::MessageKey {
+            remote_jid: Some(self.info.source.chat.to_string()),
+            from_me: Some(self.info.source.is_from_me),
+            id: Some(self.info.id.clone()),
+            participant: needs_participant.then(|| self.info.source.sender.to_string()),
+        }
+    }
+
     pub async fn edit_message(
         &self,
         original_message_id: impl Into<String>,
--- a/src/features/status.rs
+++ b/src/features/status.rs
@@ -180,22 +180,6 @@ impl<'a> Status<'a> {
             .send_status_message(revoke_message, recipients, options)
             .await
     }
-
-    /// React to someone's status update (e.g. the "like" heart in WA Web).
-    ///
-    /// `status_owner` is the JID of the person whose status you're reacting to.
-    /// `server_id` is the server-assigned ID of the status message.
-    /// `reaction` is the emoji code (e.g. "💚" for the default like). Pass empty to remove.
-    pub async fn send_reaction(
-        &self,
-        status_owner: &Jid,
-        server_id: u64,
-        reaction: &str,
-    ) -> Result<(), anyhow::Error> {
-        self.client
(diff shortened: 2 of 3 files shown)


jlucaso1:
@claude review nitpicky

PR #565, opened 2026-04-18
fix: cache-aside fallback in get_lid_pn_entry

jlucaso1:
## Summary

`Client::get_lid_pn_entry` only queried the in-memory cache. Any mapping that lived exclusively on the backend (because the cache had been evicted, or warm-up wasn't run for that client instance) would return `None` even though the persistent data had the answer.

Now it's cache-aside: on cache miss, falls back to `Backend::get_lid_mapping` / `get_pn_mapping`, converts to `LidPnEntry`, populates the in-memory cache, and returns the entry. Any backend implementation gains this fallback without additional wiring.

## Change

`src/client/lid_pn.rs`:
- `get_lid_pn_entry` tries the in-memory cache first, then the backend on miss
- Conversion from `LidPnMappingEntry` → `LidPnEntry` mirrors the existing `warm_up_lid_pn_cache` path (`LidPnEntry::with_timestamp` + `LearningSource::parse`)
- Backend errors are logged at DEBUG and treated as "no mapping" — the lookup never propagates a backend failure to the caller

## Test plan

- [ ] `cargo test -p whatsapp-rust --lib client::lid_pn::tests` — 6 cases pass, including new `test_get_lid_pn_entry_falls_back_to_backend` regression
- [ ] `cargo clippy -p whatsapp-rust --tests`

### DIFF
--- a/src/client/device_registry.rs
+++ b/src/client/device_registry.rs
@@ -553,7 +553,7 @@ mod tests {
     async fn setup_lid_pn(client: &Arc<Client>, lid: &str, pn: &str) {
         use crate::lid_pn_cache::LidPnEntry;
         let entry = LidPnEntry::new(lid.to_string(), pn.to_string(), LearningSource::Usync);
-        client.lid_pn_cache.add(entry).await;
+        client.lid_pn_cache.add(&entry).await;
     }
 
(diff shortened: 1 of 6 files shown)


jlucaso1:
@claude review nitpicky also check opportunities of allocation-less reduce clone etc

PR #577, opened 2026-04-20
fix: concurrent disconnect hang (flush_scope counter leak + cleanup_connection_state socket teardown)

jlucaso1:
## Summary

Fixes an end-to-end hang where `Client::disconnect()` called concurrently on 2+ clients (e.g. `tokio::join!(a.disconnect(), b.disconnect(), c.disconnect())`) would leave at least one client's WebSocket un-closed, pinning the host's event loop indefinitely. Surfaced by the `baileyrs` consumer (Node/WASM bridge) whose e2e suite went from `28s, all green` on alpha.11 to `hangs forever` on alpha.12 (after #576 landed).

Two independent defects were found and fixed:

### 1. `FlushScope` counter leak when the future is dropped pre-poll
`FlushScope::spawn` constructed the `DecrementOnDrop` guard *inside* the spawned future's body, so the guard only existed from the first poll onwards. If a runtime drops the future before polling it (WASM scheduling edge case, aborted spawn), the counter stays bumped; `flush()` then waits its full timeout on every disconnect.

**Fix:** capture the guard as an upvalue so it lives in the state machine from construction. Drop runs whether the future was polled or not.

### 2. `cleanup_connection_state` cleared `self.transport` without closing the socket
The run loop's graceful-exit path (woken by `notify_connection_shutdown()`) called `cleanup_connection_state`, which did `*self.transport.lock().await = None;` — clearing the reference but **never** calling `transport.disconnect()`. Under concurrent disconnect:

```
t+0    client-A disconnect()  →  transport.disconnect()  →  cleanup  ✅
t+0    client-B run loop     →  cleanup (clears transport=None)
t+0    client-B disconnect()  →  lock.as_ref() = None     →  skips close  ❌
```

The JS-side WebSocket / tokio `TcpStream` never received `close()`; the socket stayed in `readyState: \"open\"` forever, keeping the consumer's event loop alive.

**Fix:** `cleanup_connection_state` now takes the transport and calls `transport.disconnect()` before the clear. Whichever path reaches cleanup first authoritatively closes the socket. `Client::disconnect()` still calls `transport.disconnect()` explicitly after `outbound_flush.flush()` to preserve the flush→close ordering (receipts go out against a live socket); `Transport::disconnect` is idempotent so the double-fire is safe.

## Empirical validation

Branch built into the WASM bridge and pointed at the baileyrs e2e suite:

| Run | Before | After |
|---|---|---|
| `send-receive-message` (2 clients) | hang (1 TLSSocket pinned) | ✅ clean, 23s |
| `group-lifecycle` (3 clients) | hang | ✅ clean, 22s |
| `baileys-handoff` (3 clients + upstream swap) | hang | ✅ clean, 55s |
| **Full suite (54 tests)** | **hangs forever** | ✅ **105s, exit=0, 0 active handles, `beforeExit` fires naturally** |

## Test plan
- [x] `decrement_runs_when_future_is_dropped_before_first_poll` — unit test covering the FlushScope fix (inside `flush_scope.rs`).
- [x] `tests/e2e/tests/concurrent_disconnect.rs` — 4 regression tests: 2-client multi-thread, 2-client single-thread (WASM-like), **3-client multi-thread (the actual field-report repro)**, pending-receipts-in-flight.
- [x] `cargo test --lib` — 1198 passing, 0 failed.
- [x] `cargo clippy --all --tests --no-deps` — clean.
- [x] `cargo fmt --all` — applied.

## Commits
- \`c4ac4f6\` — fix(flush_scope): plug counter leak if tracked task is dropped before first poll
- \`c46c782\` — fix(client): disconnect transport from cleanup_connection_state
- \`2629305\` — chore(pr): clean up debug logs, tighten tests

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1211,7 +1211,11 @@ impl Client {
     }
 
     pub async fn disconnect(self: &Arc<Self>) {
+        use wacore::time::Instant;
+
         info!("Disconnecting client intentionally.");
+        let t_total = Instant::now();
+
         self.expected_disconnect.store(true, Ordering::Relaxed);
         self.is_running.store(false, Ordering::Relaxed);
         self.shutdown_notifier.notify();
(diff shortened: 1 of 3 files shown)


jlucaso1:
@claude review

PR #578, opened 2026-04-20
perf(message): off-load LID-PN persist + migrations from the decrypt hot path

jlucaso1:
## Summary
- Field report: 100-130ms pre-decrypt stalls in `classify_incoming_message` for ~3% of group messages with LID-PN migration. Root cause: the hot path awaits a SQLite `put_lid_mapping` + two Signal/device-registry migrations before `resolve_encryption_jid`, even though those only affect future messages — the in-memory cache already has what resolution needs.
- WA Web does this split (`docs/captured-js/WAWeb/DB/CreateLidPnMappings.js`): `warmUpLidPnMapping` is sync, DB write is either batched (`flushImmediately=true`) or queued in `lidPnCacheDirtySet`. For offline replays `flushImmediately=false`.

## Change
- `record_lid_pn_in_memory` (sync cache update) + `persist_and_migrate_lid_pn` (DB + migrations) extracted so `add_lid_pn_mapping` (awaited, used by usync/pair/notifications) and a new `learn_lid_pn_mapping_fast` (detached, used by message hot path) share code.
- `learn_lid_pn_mapping_fast` takes `is_offline`; offline replays only warm the cache — same shape as WA Web.
- `cache_lid_pn_from_message` threads `MessageInfo.is_offline` through.

## Properties
- **Correctness**: after `learn_lid_pn_mapping_fast` returns, `resolve_encryption_jid` finds the mapping (unit test added).
- **Allocations**: zero extra clones on the hot path. `entry` is moved into `LidPnMappingEntry`; migrations borrow from it.
- **Race**: two concurrent learns for the same PN both spawn detached persists. `put_lid_mapping` is upsert; both migrations are idempotent after the first completes. Documented in the method doc rather than adding a dedup lock.
- **Durability**: fast path is best-effort. Callers needing a durable guarantee should use `add_lid_pn_mapping` (doc explains this).

## Test plan
- [x] `cargo test -p whatsapp-rust --lib` (390 passed, +1 test)
- [x] `cargo test -p e2e-tests --test receipts --test offline_messages --test connection --test messaging --test concurrent_disconnect` green
- [ ] Not included: an e2e reproduction of the 100-130ms latency. Would need a 1500-member group with active LID migration — outside our mock server's scope. Field validation by the reporter on their workload is the next step.

### DIFF
--- a/src/client/lid_pn.rs
+++ b/src/client/lid_pn.rs
@@ -9,6 +9,8 @@
 //! - Resolving JIDs to their LID equivalents
 //! - Bidirectional lookup (LID to PN and PN to LID)
 
+use std::sync::Arc;
+
 use anyhow::Result;
 use log::debug;
 use wacore::store::traits::LidPnMappingEntry;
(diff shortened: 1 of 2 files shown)


jlucaso1:
@claude review

PR #579, opened 2026-04-20
fix(send): close LID↔PN zombie path for group prekey 406 latency spikes

jlucaso1:
## Summary
Field report against alpha.12: 23 batch 406s on the same group in 3h45m, same stale devices across batches, producing 12-41s latency spikes per group send. This PR closes the zombie-formation loop end-to-end.

## Root cause
Asymmetric key-storage in the device registry:
- `maybe_compute_skdm_devices` resolves LID participants to **PN form** before usync; usync responses are persisted under the PN key.
- On 406, `stale_device_users` is **LID form** (from `distribution_list` in LID-mode groups).
- `invalidate_device_cache(lid)` resolves aliases via `lid_pn_cache`; with no mapping it only invalidates the LID key, leaving the PN-keyed row as a zombie → every subsequent send refetches the same stale devices.

The LID↔PN mapping was never learned for silent-observer participants because `query_info` built `lid_to_pn_map` only inside `GroupInfo` — never fed the global `lid_pn_cache`. WA Web's `Create/OrReplaceDisplayNamesAndLidPnMappingsJob` populates the cache from group data with `learningSource: "other"`.

## Changes
1. **`groups.rs::query_info`**: push each `lid→pn` pair into `lid_pn_cache` via `learn_lid_pn_mapping_fast` (synchronous cache update + detached persist), guarded by `pn_jid.is_pn()`. Closes the silent-observer gap. WA Web parity.
2. **`wacore::send.rs::encrypt_for_devices`**: drop the O(N·RTT) per-device retry on batch 406. The server returns all-or-nothing here; the field log shows every individual retry also returned 406. Match WA Web `GroupSkmsgJob`: log, mark `had_406`, continue without those devices.
3. **`wacore::send.rs::prepare_group_stanza`**: emit both LID and PN aliases in `stale_device_users` when `GroupInfo` knows the mapping (guarded with `pn_jid.is_pn()`). Defensive against `lid_pn_cache` not being populated yet.
4. **`device_registry.rs::update_device_list`**: delete the stale DB row when the canonical key flips, not just the cache. Closes a zombie-creation path.
5. **`device_registry.rs::migrate_device_registry_on_lid_discovery`**: delete the PN-keyed DB row during LID migration, not just the cache.
6. **Race closure (applies to #4 and #5)**: invalidate cache → delete DB → invalidate cache again. Without the second invalidate, a concurrent reader between the invalidate and the delete can re-populate the cache from the about-to-be-deleted DB row, resurrecting the zombie.

## Review notes
Codex reviewed twice and all flags were addressed:
- Round 1: confirmed `self_weak.upgrade()` safe, Fix 2 doesn't lose salvageable bundles, no races in Fix 1.
- Round 1 flag: `pn_jid.is_pn()` guard missing in Fix 3 → added.
- Round 1 flag: pre-existing migration leak (PN-keyed DB row not deleted) → plugged via Fixes 4/5.
- Round 2 flag: `pn_jid.is_pn()` guard missing in Fix 1 too → added.
- Round 2 flag: TOCTOU window between invalidate and delete in Fixes 4/5 → closed via invalidate-delete-invalidate.

WA Web compliance:
- Fix 1 matches `Create/OrReplaceDisplayNamesAndLidPnMappingsJob.js:54` (`learningSource: "other"`).
- Fix 2 matches `Send/GroupSkmsgJob.js:18-43` (try/catch + continue).
- Fixes 3-6 address architecture specific to our dual-key storage; WA Web uses LID-primary IndexedDB so doesn't hit this.

## Test plan
- [x] `cargo test -p whatsapp-rust --lib` (390 passed)
- [x] `cargo test -p e2e-tests --test groups --test messaging --test offline_messages --test concurrent_disconnect` all green
- [ ] Field validation by the reporter on the affected bot — the 23 batches/3h45m metric is the success criterion
- [ ] Not included: a synthetic e2e reproducing the zombie state (would require seeding the registry with a stale PN-keyed LID entry; doable but outside this PR's scope)

### DIFF
--- a/src/client/device_registry.rs
+++ b/src/client/device_registry.rs
@@ -1234,4 +1234,207 @@ mod tests {
             "sender key cache should have been invalidated after device removal"
         );
     }
+
+    // ── LID↔PN zombie-path regression tests (PR #579) ───────────────────
+
+    /// U1 — `update_device_list` deletes the stale DB row when the canonical
+    /// key flips (e.g. the LID↔PN mapping is learned between two writes).
+    /// Without this, the old PN-keyed row lingers and re-surfaces as a zombie
+    /// through alias lookup, causing 406s on group sends.
+    #[tokio::test]
+    async fn test_update_device_list_canonical_flip_deletes_old_db_row() {
+        use wacore::store::traits::{DeviceInfo, DeviceListRecord};
+
+        let client = create_test_client().await;
+        let pn = "15550000011";
+        let lid = "100000000000011";
+        let backend = client.persistence_manager.backend();
+
+        // Legacy state: DB row stored under PN (mapping wasn't known yet).
+        backend
+            .update_device_list(DeviceListRecord {
+                user: pn.to_string(),
+                devices: vec![DeviceInfo {
+                    device_id: 5,
+                    key_index: None,
+                }],
+                timestamp: wacore::time::now_secs(),
+                phash: None,
+                raw_id: None,
+            })
+            .await
+            .unwrap();
+
+        setup_lid_pn(&client, lid, pn).await;
+
+        // New write: `update_device_list` with original_user = PN, canonical
+        // now resolves to LID because the mapping is known.
+        client
+            .update_device_list(DeviceListRecord {
+                user: pn.to_string(),
+                devices: vec![DeviceInfo {
+                    device_id: 7,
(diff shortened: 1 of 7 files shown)


jlucaso1:
@claude review

PR #589, opened 2026-04-24
fix(decrypt): fall through when <unavailable> comes alongside <enc>

jlucaso1:
## Bug

`<unavailable>` was treated as terminal regardless of sibling children. When the server delivers `<unavailable>` **alongside** `<enc>` — which it does for companions it trusts — we were dropping the cipher and firing PDO uselessly.

## Fix

Guard the short-circuit with `all_enc_nodes.is_empty()`, matching whatsmeow [`message.go:298-309`](https://github.com/tulir/whatsmeow/blob/02ec817e7cf4012e68826a7384921e590b17eabf/message.go#L298-L309).

| `<enc>` | `<unavailable>` | Action |
|---|---|---|
| empty | absent | warn + return |
| empty | present | PDO + UndecryptableMessage |
| present | any | decrypt |

## Tests

- `test_unavailable_with_enc_skips_unavailable_shortcut` — regression guard
- `test_unavailable_without_enc_dispatches_view_once_event` — preserves stub path

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -402,7 +402,9 @@ impl Client {
             return None;
         }
 
-        if let Some(unavailable) = unavailable_node {
+        if let Some(unavailable) = unavailable_node
+            && all_enc_nodes.is_empty()
+        {
             let unavailable_type = match unavailable.get_attr("type").map(|v| v.as_str()).as_deref()
             {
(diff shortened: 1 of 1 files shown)


jlucaso1:
@claude review

PR #591, opened 2026-04-24
[codex] fix e2e reconnect and receipt flakes

jlucaso1:
## What changed

- Keep the per-connection pipeline alive until `outbound_flush` drains before signaling connection shutdown in `disconnect()`, `reconnect()`, and `reconnect_immediately()`.
- Add a deterministic unit test that blocks a transport send and proves cleanup cannot close the transport before the tracked flush completes.
- Make the E2E `reconnect_and_wait()` helper use immediate reconnects; tests that need an offline window still call `reconnect()` directly.
- Adjust the receipt flush E2E so it asserts client-side receipt stanza creation instead of depending on the mock server routing every receipt after a near-simultaneous close frame.

## Why

The latest `main` commit reduced channel/cache preallocations and exposed two existing timing assumptions in CI:

- App state reconnect tests used a helper that spent roughly half of its 10s wait budget in the intentional `reconnect()` backoff, leaving too little room for post-login readiness under CI load.
- The receipt disconnect test assumed that once B built delivery receipts, the mock server would always route every receipt to A before B's close completed. Logs showed B sending all receipts while A sometimes received only a subset.

There was also a real client shutdown race: connection cleanup could be signaled before the tracked receipt flush completed, allowing transport teardown to race in-flight outbound receipts.

## Impact

This does not add capacity, channels, or hot-path send overhead. The behavior change is limited to shutdown/reconnect paths: tracked outbound receipts are allowed to drain before connection cleanup is signaled.

## Validation

- `cargo test -p whatsapp-rust disconnect_does_not_signal_connection_cleanup_before_outbound_flush -- --nocapture`
- `cargo test -p e2e-tests --test receipts -- --nocapture`
- `cargo test -p e2e-tests --test app_state -- --nocapture`
- `cargo test -p whatsapp-rust --lib`
- `cargo clippy -p whatsapp-rust -p e2e-tests --tests`

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1256,24 +1256,18 @@ impl Client {
         self.expected_disconnect.store(true, Ordering::Relaxed);
         self.is_running.store(false, Ordering::Relaxed);
         self.shutdown_notifier.notify();
-        // Stop the read loop first so no new stanzas enter the pipeline while
-        // we're flushing receipts (issue #571).
-        self.notify_connection_shutdown();
 
+        // Keep the per-connection pipeline alive until tracked receipts drain.
         self.outbound_flush
             .flush(&*self.runtime, std::time::Duration::from_secs(5))
             .await;
+        self.notify_connection_shutdown();
 
         if let Err(e) = self.persistence_manager.flush().await {
             log::error!("Failed to flush device state during disconnect: {e}");
         }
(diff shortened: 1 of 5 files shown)


jlucaso1:
@claude review

PR #592, opened 2026-04-24
fix(pair-code)!: derive companion_platform_{id,display} from DeviceProps

jlucaso1:
## Summary

- `PairCodeOptions::default()` had a hardcoded `PlatformId::Chrome` + `"Chrome (Linux)"`, so the `companion_hello` IQ announced the companion as Chrome/Linux even when `DeviceProps` was configured as Android. Push notification on the primary phone rendered `Chrome (Linux)` while Linked Devices rendered `Android (Android)` because those two screens come from different paths (`link_code_companion_reg` vs `ClientPayload.device_pairing_data.device_props`).
- Mirror WA Web's approach (`WAWebCompanionRegClientUtils.DEVICE_PLATFORM` + `"<name> (<os>)"`) — `companion_platform_id`/`companion_platform_display` are now derived from `Device.device_props` via a new `resolve_companion_platform` helper. `DeviceProps` becomes the single source of truth end-to-end.
- `PlatformId` enum removed (covered only web IDs 0..9, couldn't express `ANDROID_PHONE = 16`). `PairCodeOptions.platform_id`/`platform_display` now `Option<...>` — explicit overrides still win; `None` falls back to derivation.

## Wire behaviour

- `Device::new()` + `with_device_props(os="Android", platform_type=ANDROID_PHONE)` + `PairCodeOptions::default()` now emits `<companion_platform_id>16</…>` and `<companion_platform_display>Android (Android)</…>`.
- Legacy web case (`os="Linux"` + `CHROME`) still emits exactly `1` / `Chrome (Linux)` — paridade com captura WA Web mantida.

## Breaking changes

- `PlatformId` enum removed. Migration: drop it, or pass `platform_id: Some(1)` if you really want the integer.
- `PairCodeOptions.platform_id: Option<i32>`, `platform_display: Option<String>`. Default is `None` for both — derivation kicks in.
- `PairCodeUtils::build_companion_hello_iq` takes `platform_id: &str` instead of the enum.

## Regression coverage

New tests in `wacore/src/pair_code.rs`:

- `derive_bare_device_props_never_emits_chrome_linux` — raw `DeviceProps::default()` must not yield the legacy web identifier.
- `pair_code_options_default_has_no_platform_hardcode` — guards `Default` against re-introducing the hardcode.
- `bug_android_device_props_no_longer_emit_chrome_linux` — end-to-end through `resolve_companion_platform` + `build_companion_hello_iq`, asserts both positive ("16" / "Android (Android)") and negative sides.
- `companion_hello_iq_chrome_linux_wire_parity` — web case unchanged.
- `companion_hello_iq_android_wire_strings`, `companion_hello_iq_shape` — wire shape pins.
- `friendly_name_exhaustive_non_empty` — exhaustive check over all proto `PlatformType` variants; adding a new variant becomes a compile error instead of silently falling through.
- Plus `derive_*` / `resolve_*` matrix: Chrome (Linux), Android phone/tablet, iOS phone, iPad, no-os, unknown proto, explicit-id-only override, explicit-display-only override, full override.

## Verification

- [x] `cargo fmt --all`
- [x] `cargo test -p wacore -p whatsapp-rust pair_code` — 37 pair_code tests pass
- [ ] Manual pair flow confirming primary phone push notification no longer shows `Chrome (Linux)` when `DeviceProps` declares Android. Cannot be covered by unit tests since it requires a live pairing. If the phone app ignores non-web `companion_platform_id` values (e.g. `16`), the override path is still there and the bridge can keep passing explicit values.

## Bridge guidance

`whatsapp-rust-bridge` / `baileyrs` should stop forwarding hardcoded `platform_id`/`platform_display` to `PairCodeOptions`. As long as `setDeviceProps` is called before `requestPairingCode`, derivation handles it. Explicit override remains available for power users.

### DIFF
--- a/src/bot.rs
+++ b/src/bot.rs
@@ -520,18 +520,19 @@ impl<B, T, H, R> BotBuilder<B, T, H, R> {
     ///
     /// # Example
     /// ```rust,ignore
-    /// use whatsapp_rust::pair_code::{PairCodeOptions, PlatformId};
+    /// use whatsapp_rust::pair_code::PairCodeOptions;
     ///
+    /// // Platform identity is derived from `DeviceProps` configured via
+    /// // `Bot::builder().with_device_props(...)`. Explicit overrides below
+    /// // are optional — omit them to let derivation do the right thing.
     /// let bot = Bot::builder()
     ///     .with_backend(backend)
     ///     .with_transport_factory(transport)
     ///     .with_http_client(http_client)
     ///     .with_pair_code(PairCodeOptions {
     ///         phone_number: "15551234567".to_string(),
-    ///         show_push_notification: true,
     ///         custom_code: Some("ABCD1234".to_string()),
-    ///         platform_id: PlatformId::Chrome,
-    ///         platform_display: "Chrome (Linux)".to_string(),
+    ///         ..Default::default()
     ///     })
     ///     .on_event(|event, client| async move {
     ///         match &*event {
--- a/src/pair_code.rs
+++ b/src/pair_code.rs
@@ -52,12 +52,12 @@ use log::{error, info, warn};
 
 use std::sync::Arc;
 use wacore::libsignal::protocol::KeyPair;
-use wacore::pair_code::{PairCodeError, PairCodeState, PairCodeUtils};
+use wacore::pair_code::{PairCodeError, PairCodeState, PairCodeUtils, resolve_companion_platform};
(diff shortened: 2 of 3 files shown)


jlucaso1:
@claude review

jlucaso1:
@coderabbitai review

jlucaso1:
@claude review

PR #593, opened 2026-04-25
fix(pair): WA Web compliant pairing QR + companion_platform_{id,display}

jlucaso1:
## Summary

Two bugs in the pairing flows, fixed together because they share a derivation source.

### QR pairing format (whatsmeow#1110, Baileys#2460 backport)

Primary devices started rejecting the legacy 4-field QR. WA Web's `WAWebLinkDeviceQrcode.react` now appends `WAWebCompanionRegClientUtils.DEVICE_PLATFORM` as a 5th field. Mirror that.

- New `CompanionWebClientType` enum, 10 wire values matching WA Web exactly, plus `Unrecognized(i32)` fallback. Derived via `WireEnum` int (same pattern as `TempBanReason` / `ConnectFailureReason`).
- `make_qr_data` auto-derives the client type from `device_props.platform_type`. `make_qr_data_with_client_type` is the override seam.
- `parse_qr_code` accepts the legacy 4-field form, the new 5-field form, the native-camera `https://wa.me/settings/linked_devices#` prefix, and the jknob-770 FAQ URL suffix. Empty fields, malformed base64, and wrong key lengths still rejected.

### `companion_platform_{id,display}` correctness (regression from 63210f7f)

The pair-code IQ derivation introduced in 63210f7f had two wire-correctness bugs surfaced during the audit:

1. `companion_platform_id` was the wrong enum's integer. WA Web sends `WAWebCompanionRegClientUtils.DEVICE_PLATFORM` (CompanionWebClientType, 10 values) per `docs/captured-js/WAWeb/Alt/DeviceLinkingIq.js:35`. We were formatting `wa::device_props::PlatformType as i32` (proto, 25+ values). The two enums collide on indices 1 and 7 only. Firefox sent "2" instead of "3", Edge "6" instead of "2", IE "3" instead of "4", and mobile DeviceProps sent values like "16" outside the WA Web range entirely.
2. `companion_platform_display` violated server validation. whatsmeow's `PairPhone` doc states the server validates this field strictly as `Browser (OS)` with browser ∈ {Chrome, Edge, Firefox, IE, Opera, Safari} and returns 400 otherwise. We were emitting `Android (Android)`, `iPhone (iOS)`, `iPad (iPadOS)` for mobile DeviceProps, breaking pair-code for any non-browser platform.

Fixes:

- `derive_companion_platform` returns `(CompanionWebClientType, String)` and shares its derivation with the QR flow. Single source of truth.
- `companion_browser_name` returns one of the 6 server-valid browsers, falling back to `Chrome` for non-browser variants. That's what WA Web's `info().name` emits even from Electron contexts.
- `companion_platform_display` always emits `<Browser> (<OS>)`, substituting `Linux` for empty OS since WA Web never sends bare browser names.

### Breaking changes (pre-1.0, intentional)

- `wacore::pair_code::derive_companion_platform`: `(String, String)` → `(CompanionWebClientType, String)`
- `wacore::pair_code::resolve_companion_platform`: same return-type change
- `wacore::pair_code::PairCodeOptions::platform_id`: `Option<wa::device_props::PlatformType>` → `Option<CompanionWebClientType>`
- `wacore::pair_code::PairCodeOptions::platform_display`: removed (WA Web has no equivalent override; any free string risks server 400)
- `wacore::pair_code::platform_friendly_name`: removed (only caller was the buggy derivation; the mixed browser/non-browser name table was the bug 2 vector)
- `wacore::pair::PairUtils::make_qr_data`: now requires `CompanionWebClientType` parameter

The whatsapp-rust-bridge consumer only uses `PairCodeOptions { phone_number, custom_code, ..Default::default() }` and reads the QR string via `Event::PairingQrCode`. Both work transparently with the new behaviour.

## Test plan

- [x] `cargo test -p wacore --lib`: 610 tests pass, including
  - 17 `companion_reg` tests: wire codes match WA Web, `From<i32>` round-trip, `Unrecognized` fallback, mapping totality, browser-name validity, display invariants
  - 14 `pair::tests`: QR 5-field format, retro-compat 4-field parse, native-camera prefix, FAQ suffix, round-trip auto-derive for every client type
  - rewritten `pair_code::tests`: per-browser correct wire, mobile → OtherWebClient + Chrome display fallback, total proto-PlatformType scan asserting display always uses one of 6 valid browsers, pair-code/QR id parity, override semantics
- [x] `cargo clippy --workspace --exclude e2e-tests --tests` clean
- [x] `cargo fmt --all -- --check`
- [ ] Live pairing smoke against a real primary device, both QR (default Chrome/Linux + AndroidPhone DeviceProps) and pair-code (mobile DeviceProps to confirm the previous 400 path now works)

### DIFF
--- a/src/client/lid_pn.rs
+++ b/src/client/lid_pn.rs
@@ -472,10 +472,9 @@ impl Client {
     }
 
     /// Resolve any user JID to its bare LID form, or `None` when no LID is
-    /// available. Mirrors WA Web's `WAWebLidMigrationUtils.toUserLid`
-    /// (`docs/captured-js/WAWeb/Lid/MigrationUtils.js:17-20`): LID passes
-    /// through, PN goes through the cache-aside mapping, anything else and
-    /// any lookup failure returns `None`.
+    /// available. Mirrors WA Web's `WAWebLidMigrationUtils.toUserLid`: LID
+    /// passes through, PN goes through the cache-aside mapping, anything
+    /// else and any lookup failure returns `None`.
     ///
     /// Used by `send_status_message` to replicate WA Web's
     /// `compactMap(list, toUserLid)` skip-on-unresolvable semantics.
--- a/src/features/media_reupload.rs
+++ b/src/features/media_reupload.rs
@@ -4,7 +4,7 @@
 //! sends a `<receipt type="server-error">` stanza and waits for a
 //! `<notification type="mediaretry">` response with a new `directPath`.
 //!
-//! Reference: WAWebRequestMediaReuploadManager (docs/captured-js/).
+//! Reference: WAWebRequestMediaReuploadManager.
 
 use crate::client::{Client, ClientError, NodeFilter};
 use anyhow::Result;
--- a/src/pair.rs
+++ b/src/pair.rs
@@ -12,7 +12,7 @@ use wacore_binary::NodeRef;
 use wacore_binary::{Jid, SERVER_JID};
 use waproto::whatsapp as wa;
 
-pub use wacore::companion_reg::CompanionWebClientType;
+pub use wacore::companion_reg::{CompanionWebClientType, NATIVE_CAMERA_DEEP_LINK_PREFIX};
 pub use wacore::pair::{DeviceState, PairCryptoError, PairUtils};
 
 /// Auto-derives client type from `device_props`; see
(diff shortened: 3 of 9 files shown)


jlucaso1:
@claude review nitpicky

PR #599, opened 2026-04-26
feat(call): detect group calls via offer_notice and group-jid attrs

Salientekill:
## Summary

WhatsApp Web's group-call signaling differs from 1:1 in two ways the parser was missing:

1. `<offer>` for a group call carries `type=\"group\"` and/or `group-jid=\"GROUP@g.us\"` on the action child, distinct from a 1:1 offer. Baileys reads both signals ([`Socket/messages-recv.ts:1552-1553`](https://github.com/WhiskeySockets/Baileys/blob/master/src/Socket/messages-recv.ts#L1552)).

2. When a group member receives notice that *another* member started a group call, the action child is `<offer_notice>` (not `<offer>`). The current parser had `offer_notice` as unknown and returned `Ok(None)`, so downstream consumers saw **no event at all** for the most common group-call delivery path.

Captured stanza shape (production WA, late 2026):

\`\`\`xml
<call from=\"caller@lid\" id=\"...\" t=\"...\">
  <offer_notice
    call-creator=\"caller@lid\"
    call-id=\"...\"
    media=\"audio\"
    type=\"group\"
    reason=\"427\"
    caller_pn=\"...\"/>
</call>
\`\`\`

## Changes

- `CallAction::Offer` gains `group_jid: Option<Jid>` and `is_group_type: bool`. Either non-`None`/`true` indicates a group call. Both serialize as skipped when absent so existing JSON consumers see no diff for 1:1 calls.
- New `CallAction::OfferNotice { call_id, call_creator, media, caller_pn }` for the group-call notice path. `offer_notice` is now in `KNOWN_ACTIONS`. Unconsumed attrs (`type`, `reason`) are ignored to stay forward-compatible.
- `CallAction::call_id()` / `call_creator()` cover the new variant.
- Unit test for the captured `<offer_notice type=\"group\">` shape.

## Note for consumers

`<offer_notice>` carries **no** `group-jid` — the WA server expects clients to know the group from their own state. Downstream consumers that need the group must cross-reference `call_creator` against their own group-membership cache. This matches the model Baileys exposes (`isGroup` flag + `groupJid` that may be `undefined`).

## Test plan

- [x] \`cargo test --workspace --lib\` — 638+ tests pass (was 615, +1 new test for offer_notice + previous parser tests still pass)
- [x] \`cargo clippy --workspace --tests -- -D warnings\` clean
- [x] \`cargo fmt --check\` clean
- [x] Verified against live production stanza: parser correctly extracts \`call-creator\`, \`call-id\`, \`media\`, \`caller_pn\` from \`<offer_notice type=\"group\">\`
- [x] Existing \`<offer>\` 1:1 path unaffected (existing tests + new test for is_group_type=false default)

Closes the gap that prevented building anti-call-group features (e.g. auto-removing members who initiate group calls).

### DIFF
--- a/wacore/src/stanza/call.rs
+++ b/wacore/src/stanza/call.rs
@@ -2,13 +2,21 @@
 //! children so future server additions don't break the handler.
 
 use anyhow::{Result, anyhow};
+use log::debug;
 use wacore_binary::builder::NodeBuilder;
 use wacore_binary::{Jid, Node, NodeRef};
 
 use crate::time::from_secs;
 use crate::types::call::{CallAction, CallAudioCodec, IncomingCall};
 
-const KNOWN_ACTIONS: &[&str] = &["offer", "preaccept", "accept", "reject", "terminate"];
+const KNOWN_ACTIONS: &[&str] = &[
+    "offer",
+    "offer_notice",
+    "preaccept",
+    "accept",
+    "reject",
+    "terminate",
+];
 
 pub fn parse_call_stanza(node: &NodeRef<'_>) -> Result<Option<IncomingCall>> {
     if node.tag != "call" {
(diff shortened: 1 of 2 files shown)


Salientekill:
Thanks for the review @coderabbitai. Pushed `dc81b1b` addressing all 3 items:

1. **`media` field** — added `#[serde(skip_serializing_if = "Option::is_none")]` so the field is omitted (not `null`) when absent, matching the rest of the struct.

2. **`offer_notice` `attrs.finish()`** — kept forward-compat (didn't strict-finish), but added a `debug!` log capturing the known unhandled attrs (`type`, `reason`) with `call_id` and `call_creator` for context. Silent ignore is gone; protocol drift will surface in logs.

3. **Test coverage for group `<offer>`** — added 2 tests:
   - `offer_with_group_jid_and_type`: full group offer with both `type="group"` and `group-jid="123456789@g.us"` — asserts `is_group_type=true` + `group_jid=Some(...)`.
   - `offer_with_only_type_group`: bare `type="group"` without `group-jid` — asserts `is_group_type=true` + `group_jid=None` (the Baileys "either signal" logic).

`cargo test --workspace --lib` passes (640 wacore tests, +3 new). Clippy + fmt clean.

jlucaso1 (approved):

PR #605, opened 2026-04-28
fix(lid_pn): WA Web compliant signal address for Hosted JIDs

jlucaso1:
## Background

Audit of the Hosted/HostedLid path triggered by a question on the encoder \`domain_type\` bug (#391) surfaced a deeper Signal-address divergence. WA Web's \`SignalAddress.toString()\` in \`WAWeb/Signal/Address.js\`:

\`\`\`js
if (this.wid.isHosted()) {
    if (bizHostedDevicesEnabled()) {
        if (t !== ":99") throw "Hosted jid with wrong device id";
        var n = asUserWidOrThrow(this.wid),
            a = !n.isLid() && !n.isHostedLid() && n.isUser(),
            i = a ? getCurrentLid(n) : n;
        return i == null
            ? [this.wid.user, t, "@hosted"].join("")
            : [i.user, t, "@hosted.lid"].join("");
    }
    throw "Unexpected hosted jid";
}
\`\`\`

Same pattern for non-hosted JIDs lower in the function: prefer LID-keyed addresses, fall back to the original namespace if no LID mapping is known.

## What was wrong

\`resolve_encryption_jid\` only handled the PN → LID upgrade. Hosted JIDs fell through the catch-all \`_ => target.clone()\`, so a hosted device whose LID was known was keyed locally at \`{user}@hosted\` while WA Web would key it at \`{lid_user}@hosted.lid\`.

Practical impact today: nil — current send paths filter \`is_hosted()\` before reaching the resolver, so the divergent key was never built or looked up. The gap was dormant but real, and any future hosted support would have inherited an incompatible keyspace.

## Fix

\`resolve_encryption_jid\` mirrors WA Web in a single match:

\`\`\`rust
let lid_server = match target.server {
    Server::Pn => Server::Lid,
    Server::Hosted => Server::HostedLid,
    _ => return target.clone(),
};
match self.lid_pn_cache.get_current_lid(&target.user).await {
    Some(lid_user) => Jid { user: lid_user.into(), server: lid_server, .. },
    None => target.clone(),
}
\`\`\`

| Input | Output | Source |
|---|---|---|
| Lid / HostedLid | unchanged | already canonical |
| Pn | Lid when mapping known, else Pn | matches existing behavior |
| Hosted | **HostedLid when mapping known, else Hosted** | new |
| anything else | unchanged | matches existing behavior |

The Hosted upgrade reuses the same \`lid_pn_cache.get_current_lid()\` lookup as the PN upgrade. PN/LID dispatch cost is unchanged.

### Allocation profile

| Branch | Heap allocs |
|---|---|
| Lid / HostedLid / other | 0 (CompactString user inline ≤24 bytes — always for phone/lid lengths) |
| Pn / Hosted with mapping | 1 wasted (String temp from \`lid_pn_cache\`, dropped after \`.into()\` to inline CompactString) |
| Pn / Hosted no mapping | 0 (cache miss + inline clone) |

The wasted \`String\` is a property of \`lid_pn_cache.get_current_lid() -> Option<String>\`. Eliminating it requires changing \`LidPnEntry.lid: String → CompactString\` cross-crate; out of scope here, dormant in hot paths today (LID groups don't go through the upgrade — devices already are LID).

## Tests

- \`test_resolve_encryption_jid_hosted_with_lid_upgrades_to_hosted_lid\` — loops over \`[99, 7]\` device ids to prove the input device round-trips and is **not** coerced to 99.
- \`test_resolve_encryption_jid_hosted_no_mapping_keeps_hosted\`
- \`test_resolve_encryption_jid_preserves_hosted_lid\`

Plus a wire-encoding regression in \`wacore/binary/src/encoder.rs\`: direct-constructed Hosted JIDs (\`Jid::new(_, Server::Hosted)\` defaults \`agent=0\`) now emit the correct \`domain_type\` byte (128 / 129) because the encoder derives it from the server enum, not from \`agent\`. Pre-#391 the same input would have written \`0\`.

## What's not in this PR

- Hosted send-path activation. Group SKDM still filters hosted (matches WA Web's \`!t.includes("hosted")\` filter in \`getGroupSenderKeyListFromParticipantRecord\`). DM hosted filters remain in place. Activating hosted as an active recipient is a feature, not part of address compliance.
- Device=99 enforcement at the JID boundary. WA Web throws on hosted JIDs with other device ids; we preserve the input and let the wire layer route it. Soft-reject is a separate hardening.
- Hosted ↔ HostedLid swap in \`swap_pn_lid_namespace\`. That helper handles cross-alias message lookup which is dormant for hosted (no stored messages keyed under hosted today).
- \`LidPnEntry.lid\` switch from \`String\` to \`CompactString\` to drop the wasted String alloc on the upgrade path.

## Test plan

- [x] \`cargo test --workspace --lib\` — 1370+ tests pass
- [x] \`cargo clippy --all --tests\` — clean
- [x] \`cargo fmt --all\`

### DIFF
--- a/src/client/lid_pn.rs
+++ b/src/client/lid_pn.rs
@@ -549,6 +549,8 @@ mod tests {
         for device in [99u16, 7] {
             let mut hosted = Jid::new(user, Server::Hosted);
             hosted.device = device;
+            hosted.agent = 0xAB;
+            hosted.integrator = 0xBEEF;
             let resolved = client.resolve_encryption_jid(&hosted).await;
 
             assert_eq!(resolved.user, lid);
@@ -557,6 +559,8 @@ mod tests {
                 resolved.device, device,
                 "device must round-trip, not be coerced to 99"
             );
+            assert_eq!(resolved.agent, hosted.agent);
+            assert_eq!(resolved.integrator, hosted.integrator);
         }
     }
 
--- a/src/client/lid_pn.rs
+++ b/src/client/lid_pn.rs
@@ -295,40 +295,51 @@ impl Client {
         resolved
     }
 
-    /// Resolve the encryption JID for a given target JID.
-    /// This uses the same logic as the receiving path to ensure consistent
-    /// lock keys between sending and receiving.
+    /// Resolve a JID into the canonical form used for Signal session and
+    /// sender-key lookups. Mirrors WA Web's `SignalAddress.toString()`
+    /// (`WAWeb/Signal/Address.js`):
     ///
-    /// For PN JIDs, this checks if a LID mapping exists and returns the LID.
-    /// This ensures that sending and receiving use the same session lock.
+    /// - LID / HostedLid: already in canonical form, returned as-is.
+    /// - PN: upgraded to `{lid_user}@lid` when a mapping is known, kept as
+    ///   `{user}@s.whatsapp.net` (mapped to `@c.us` at format time) otherwise.
+    /// - Hosted: upgraded to `{lid_user}@hosted.lid` when a mapping is known,
+    ///   kept as `{user}@hosted` otherwise. WA Web hardcodes device=99 for
+    ///   hosted; we preserve whatever device the input carries so a
+    ///   misformed JID is still routable rather than silently dropped.
     pub(crate) async fn resolve_encryption_jid(&self, target: &Jid) -> Jid {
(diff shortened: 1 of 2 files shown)


jlucaso1:
@coderabbitai review

PR #621, opened 2026-05-12
fix(send): admin revoke not applied on recipient devices

Salientekill:
## Summary

Admin revoke (`edit=\"8\"`) had **three independent bugs** that combined to leave the revoked message visible on a subset of recipients — most commonly the group admins/creator and any device that triggered a retry receipt.

Issue surfaced while writing an anti-content bot: revoke admin call returned `Ok(_)` and some clients saw the message disappear, but the bot operator's own device (which routinely retries on first contact) kept the message visible. The existing test in `wacore/src/types/message.rs` already documents the expected behavior for `AdminRevoke`.

### 1. `prepare_dm_stanza` (peer fanout) sent `decrypt-fail=\"hide\"` for admin revokes

The SKMSG path in `prepare_group_stanza` (l. 1384) already excluded `AdminRevoke` from `hide_decrypt_fail`. The peer-fanout path in `prepare_dm_stanza` (l. 862) didn't — every `<enc>` went out with the attribute, and the server dropped the revoke for those recipients.

**Fix:** mirror the same exclusion: `*e != Empty && *e != AdminRevoke`.

### 2. SKDM distribution hardcoded `hide_decrypt_fail = true`

`prepare_group_stanza` l. 1326 sent the sender-key distribution with `hide_decrypt_fail: true` always. For admin revoke this dropped the SKDM, so any device that didn't already have the group sender key never decrypted the main `<enc type=\"skmsg\">` carrying the REVOKE protocol message — even though the main payload itself was correctly attribute-clean.

**Fix:** compute `skdm_hide_decrypt_fail` with the same rule used for the main payload. Reactions, edits, pins keep the `hide` behavior; only `AdminRevoke` flips to `false`.

### 3. `prepare_{group,dm}_retry_stanza` lost the `edit` attribute

When a recipient device replies with `<receipt type=\"retry\">`, the bot rebuilds the stanza via the retry builders and resends. Those builders never accepted or emitted an `edit` attribute, so the retry resend went out as a plain `<message>`. Client received it but had no signal that this was a revoke. Symptom: the group creator's device (which always retries because its session needs a prekey bundle on first contact) kept showing the revoked image.

**Fix:**
- Both retry builders now accept `edit: Option<EditAttribute>` and emit `attr(\"edit\", ...)` when present.
- New `EditAttribute::infer_from_message(&wa::Message)` helper recovers the wire `edit` value from the cached protobuf (`add_recent_message` only persists the protobuf body, not the original stanza's wire attributes):
  - `protocol_message.r#type == Revoke` + `key.from_me` → `SenderRevoke` / `AdminRevoke`
  - `pin_in_chat_message.is_some()` → `PinInChat`
  - `protocol_message.edited_message.is_some()` → `MessageEdit`
- `retry.rs` infers `edit_attr` from `original_msg` and forwards it to both retry paths.

## Test plan

- [x] `cargo build --release --target x86_64-unknown-linux-musl` (the consumer profile) builds clean.
- [x] End-to-end test in a live LID-addressed group with multiple device types (primary phone + companion device that triggers retry receipt): admin revoke applied uniformly across recipients.
- [x] Existing test `test_decrypt_fail_hide_logic_for_edits` in `wacore/src/types/message.rs` documents the expected rule and matches the new implementation.

Happy to split into three smaller commits if reviewers prefer one fix per commit — kept consolidated here because the three bugs only become observably correct together for the admin-revoke flow.

### DIFF
--- a/src/retry.rs
+++ b/src/retry.rs
@@ -434,6 +434,8 @@ impl Client {
             let _session_guard = session_mutex.lock().await;
             let mut store_adapter = self.signal_adapter().await;
 
+            let edit_attr =
+                wacore::types::message::EditAttribute::infer_from_message(&original_msg);
             let stanza = wacore::send::prepare_group_retry_stanza(
                 &mut store_adapter.session_store,
                 &mut store_adapter.identity_store,
@@ -445,6 +447,7 @@ impl Client {
                 retry_count,
                 device_snapshot.account.as_ref(),
                 addressing_mode,
+                edit_attr,
             )
             .await?;
 
@@ -464,6 +467,8 @@ impl Client {
             let _session_guard = session_mutex.lock().await;
             let mut store_adapter = self.signal_adapter().await;
 
+            let edit_attr =
+                wacore::types::message::EditAttribute::infer_from_message(&original_msg);
             let stanza = wacore::send::prepare_dm_retry_stanza(
                 &mut store_adapter.session_store,
                 &mut store_adapter.identity_store,
@@ -474,6 +479,7 @@ impl Client {
                 message_id,
                 retry_count,
                 device_snapshot.account.as_ref(),
+                edit_attr,
             )
             .await?;
 
--- a/wacore/src/send.rs
+++ b/wacore/src/send.rs
@@ -51,7 +51,12 @@ pub fn extract_ciphertext(msg: CiphertextMessage) -> Option<(&'static str, bool,
 
 /// Unwrap wrapper message types to reach the inner message.
 /// Matches WA Web's getUnwrappedProtobufMessage (EProtoUtils.js:19-35).
-fn unwrap_message(msg: &wa::Message) -> &wa::Message {
+///
+/// Public so the retry-edit-inference path can descend through neutral
+/// wrappers (ephemeral / view_once / device_sent) before checking for
+/// pin/edit/revoke fields. Does **not** unwrap `edited_message` — that field
+/// is itself the signal we're trying to detect.
+pub fn unwrap_message(msg: &wa::Message) -> &wa::Message {
     macro_rules! try_unwrap {
         ($($field:ident),+ $(,)?) => {
             $(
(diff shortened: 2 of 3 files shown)


Salientekill:
Thanks @coderabbitai — both suggestions applied in ef61bd1:

1. **`infer_from_message` now checks top-level `edited_message`** before falling through to `protocol_message`. The new branch order is: `pin_in_chat_message` → top-level `edited_message` (current WA Web format) → `protocol_message.r#type == Revoke` → legacy `protocol_message.edited_message`. Added 5 unit tests in `types::message::tests` covering admin/sender revoke, top-level edit, legacy edit, and plain (returns None).

2. **`hide_decrypt_fail` consolidated** into `should_hide_decrypt_fail_for_send(edit, msg)` in `wacore/src/send.rs`. The three previous inline copies (DM peer fanout, group SKDM, group SKMSG) now call the helper. Same canonical rule documented in the docstring.

Local: `cargo test -p wacore --lib --no-default-features` → **687 passed**, no failures. `cargo clippy -p wacore --no-default-features --tests` → no new warnings.

Salientekill:
Round 2 applied in 5806e55:

1. **`AdminEdit` (edit="3") now covered** in `infer_from_message`. Mirrors the Sender/Admin Revoke logic: when `protocol_message.r#type == MessageEdit`, `key.from_me == true` → `MessageEdit`, `false` → `AdminEdit`. Top-level `edited_message` (no `from_me` flag available) keeps defaulting to `MessageEdit`. Added regression tests `infer_from_message_message_edit_sender` and `infer_from_message_admin_edit`.

2. **`cargo fmt`** applied (the previous push was missing it — that's why Format Check went red). Verified with `cargo +nightly-2026-04-05 fmt --all -- --check` exit 0.

Local: 689 tests passed, no failures, no clippy warnings introduced.

Salientekill:
Round 3 applied in 78b5e6c:

**Neutral wrappers now unwrapped before inference.** `infer_from_message` calls `send::unwrap_message` first, so payloads nested inside `ephemeral_message`, `view_once_message{,_v2,_v2_extension}`, `device_sent_message`, `document_with_caption_message`, `group_mentioned_message`, `bot_invoke_message`, `associated_child_message`, or `poll_creation_option_image_message` are descended through before checking pin/edit/revoke fields.

`unwrap_message` was made `pub` (it intentionally does **not** include `edited_message` since that field is the signal we're trying to detect — calling `get_base_message` from `MessageExt` would have collapsed the edit envelope and broken detection).

Added regression test `infer_from_message_unwraps_neutral_wrappers` covering:
- Admin revoke nested under `ephemeral_message` → returns `AdminRevoke`.
- Pin nested through `device_sent_message → view_once_message` (double wrapping) → returns `PinInChat`.

Local: **690 tests passed**, fmt clean.

PR #633, opened 2026-05-19
fix(client): preserve recipient in <ack>, soften unknown stream:error

zdanysfa:
# Fix: stop spurious stream disconnects on `<stream:error><ack/>` notifications

## Summary

Two independent fixes — one in the ACK encoder, one in the stream-error handler — that together stop the reconnect storm users have been seeing under the JS bridge as `DisconnectReason.badSession (500)`.

1. **Encoder** (`encode_ack_bytes`) now copies the `recipient` attribute from the incoming stanza onto the outgoing `<ack/>`, matching whatsmeow. Without it, server cannot route the ACK back to the originating companion/device and replies with `<stream:error><ack/>…</stream:error>`.
2. **Handler** (`handle_stream_error` catch-all) no longer treats unknown / code-less stream errors as fatal. It logs + dispatches the `StreamError` event and leaves the connection alone, exactly like whatsmeow's default branch (`connectionevents.go:66-69`). The socket layer remains responsible for detecting a real teardown.

Either fix in isolation reduces the symptom; together they remove both the trigger and the over-reaction.

---

## Problem

A bot running this stack repeatedly reconnects with no actionable error. The JS bridge surfaces it as `badSession (500)`, but **the server never sends `<stream:error code="500">`** — that "500" is a bridge-side fallback for `expected_disconnect` without a recognised code.

Across 19 captured Rust debug logs (7 distinct accounts, ~2-hour window), **every** disconnect matched this exact shape:

```
DEBUG <stream:error><ack class="message" type="text" id="2A32F960553696093D99"/></stream:error>
ERROR Unknown stream error: <stream:error><ack class="message" type="text" id="2A32F960553696093D99"/></stream:error>
INFO  Notifying connection shutdown from stream error handler
DEBUG Expected disconnect signaled during frame processing. Exiting message loop.
DEBUG Transport::disconnect called
```

The `<ack>` inside the `<stream:error>` is the server's reference to the message stanza whose ACK it rejected. Immediately after reconnect, that same message is redelivered (`offline="10"`, `offline="12"`), confirming the previous ACK never registered.

Affected stanzas were a mix of:

- Peer / hosted-companion messages with `recipient="…@lid"` and `peer_recipient_pn="…"`.
- LID-mode group messages (`addressing_mode="lid"`) routed through hosted recipients.
- Retry placeholders (`<enc decrypt-fail="hide">` + `<meta tag_reason="rereg_recovery_request">`).

Common denominator: they are stanzas where WhatsApp Web / whatsmeow propagate `recipient` into the ACK. Rust drops it.

---

### Root cause (Fix A)

`whatsmeow/receipt.go:143-177` (`Client.sendAck`):

```go
attrs := waBinary.Attrs{
    "class": node.Tag,
    "id":    node.Attrs["id"],
}
attrs["to"] = node.Attrs["from"]
if participant, ok := node.Attrs["participant"]; ok {
    attrs["participant"] = participant
}
if recipient, ok := node.Attrs["recipient"]; ok {
    attrs["recipient"] = recipient
    // BotJIDMap remap (Meta-bot only) …
}
if receiptType, ok := node.Attrs["type"]; node.Tag != "message" && ok {
    attrs["type"] = receiptType
}
```

The server uses `recipient` to dispatch the ACK to the originating device. Without it the ACK is unroutable; the server signals that by closing the stream with `<stream:error><ack …/></stream:error>`. The `BotJIDMap` remap is for Meta's BotServer and has no analogue in `whatsapp-rust`, so we intentionally do not port it.

### Before

`whatsapp-rust/src/client.rs::encode_ack_bytes` (production path):

```rust
let Some(id_val)   = node.get_attr("id")   else { return Ok(None); };
let Some(from_val) = node.get_attr("from") else { return Ok(None); };
let participant_val = node.get_attr("participant").filter(|p| {
    p.as_str().as_ref() != from_val.as_str().as_ref()
});
// (no recipient handling)

let typ_val = if tag != "message" && !is_encrypt_identity_notification(node) {
    node.get_attr("type")
} else { None };

let attr_count = 3
    + usize::from(include_from)
    + usize::from(participant_val.is_some())
    + usize::from(typ_val.is_some());

struct AckNode<'a> {
    id: &'a ValueRef<'a>,
    from: &'a ValueRef<'a>,
    participant: Option<&'a ValueRef<'a>>,
    typ: Option<&'a ValueRef<'a>>,
    own_pn: Option<&'a Jid>,
    tag_str: &'a str,
    attr_count: usize,
}
// encode_attrs: class, id, to, [from], [participant], [type]
```

### After

```rust
let Some(id_val)   = node.get_attr("id")   else { return Ok(None); };
let Some(from_val) = node.get_attr("from") else { return Ok(None); };
let participant_val = node.get_attr("participant").filter(|p| {
    p.as_str().as_ref() != from_val.as_str().as_ref()
});
// Server expects `recipient` echoed back so it can route the ack to the
// origin companion/device (hosted-companion, peer, LID-routed stanzas).
// Dropping it makes the server close the stream with `<stream:error><ack/>`.
let recipient_val = node.get_attr("recipient");

let typ_val = if tag != "message" && !is_encrypt_identity_notification(node) {
    node.get_attr("type")
} else { None };

let attr_count = 3
    + usize::from(include_from)
    + usize::from(participant_val.is_some())
    + usize::from(recipient_val.is_some())
    + usize::from(typ_val.is_some());

struct AckNode<'a> {
    id: &'a ValueRef<'a>,
    from: &'a ValueRef<'a>,
    participant: Option<&'a ValueRef<'a>>,
    recipient: Option<&'a ValueRef<'a>>,   // new
    typ: Option<&'a ValueRef<'a>>,
    own_pn: Option<&'a Jid>,
    tag_str: &'a str,
    attr_count: usize,
}
// encode_attrs: class, id, to, [from], [participant], [recipient], [type]
```

The `#[cfg(test)]` mirror `build_ack_node` gets the same shape so structural assertions stay possible without touching the wire encoder.

---

### Root cause (Fix B)

`whatsmeow/connectionevents.go:19-69` deliberately ignores stream errors it doesn't recognise:

```go
default:
    cli.Log.Errorf("Unknown stream error: %s", node.XMLString())
    go cli.dispatchEvent(&events.StreamError{Code: code, Raw: node})
}
```

There is no `expectDisconnect()`, no socket close, no reconnect trigger. The design relies on the read loop noticing if the server actually ends the stream (via `xmlstreamend` or a closed socket).

`whatsapp-rust` diverges: its catch-all sets `expected_disconnect = true` and the function unconditionally calls `notify_connection_shutdown()` for **every** branch, including `429`, `503`, and unknown. That converts an informational stanza into a full shutdown + reconnect cycle, which under high traffic snowballs into the loop the logs show.

### Before

`whatsapp-rust/src/client.rs::handle_stream_error` (excerpt):

```rust
match code {
    "515" => { /* … */ should_disconnect = true; }
    "516" | "401" | "409" => { /* … */ should_disconnect = true; }
    "429" => {
        warn!("Got 429 stream error (rate limited). …");
        self.auto_reconnect_errors.fetch_add(5, Ordering::Relaxed);
    }
    "503" => {
        info!("Got 503 service unavailable, will auto-reconnect.");
    }
    _ => {
        error!("Unknown stream error: {}", DisplayableNodeRef(node));
        self.expected_disconnect.store(true, Ordering::Relaxed);     // ← bug
        self.core.event_bus.dispatch(Event::StreamError(/* … */));
    }
}

if should_disconnect {
    let transport_opt = self.transport.lock().await.clone();
    if let Some(transport) = transport_opt {
        self.runtime
            .spawn(Box::pin(async move { transport.disconnect().await; }))
            .detach();
    }
}

info!("Notifying connection shutdown from stream error handler");      // ← always runs
self.notify_connection_shutdown();                                     // ← always runs
```

### After

```rust
match code {
    "515" => { /* … */ should_disconnect = true; }
    "516" | "401" | "409" => { /* … */ should_disconnect = true; }
    "429" => {
        warn!("Got 429 stream error (rate limited). …");
        self.auto_reconnect_errors.fetch_add(5, Ordering::Relaxed);
    }
    "503" => {
        info!("Got 503 service unavailable, will auto-reconnect.");
    }
    _ => {
        // Mirror whatsmeow's default: log + dispatch event, but keep the
        // connection alive. The server wraps per-stanza notifications in
        // <stream:error> without a code attribute (e.g. <ack/> for malformed
        // routing). Proactively disconnecting on those causes reconnect storms
        // under load — whatsmeow lets the noise socket detect a real teardown.
        error!("Unknown stream error: {}", DisplayableNodeRef(node));
        self.core.event_bus.dispatch(Event::StreamError(/* … */));
    }
}

// Only tear down the connection for branches that explicitly opted in.
// 429/503/unknown match whatsmeow: log and let the socket layer notice
// if the server actually closes the stream.
if should_disconnect {
    let transport_opt = self.transport.lock().await.clone();
    if let Some(transport) = transport_opt {
        self.runtime
            .spawn(Box::pin(async move { transport.disconnect().await; }))
            .detach();
    }
    info!("Notifying connection shutdown from stream error handler");
    self.notify_connection_shutdown();
}
```

Two concrete behavior changes:

1. The catch-all no longer flips `expected_disconnect`. If the server really closes the stream afterwards, the existing socket-level path handles it; if it doesn't, the connection lives.
2. `notify_connection_shutdown()` (and the matching `info!` log) is now inside the `should_disconnect` block. The implicit side effect — `429` and `503` were also triggering shutdown — is gone.

This brings Rust's behavior matrix in line with whatsmeow:

| Stream error | whatsmeow | Rust before this PR | Rust after this PR |
|---|---|---|---|
| `code="515"` | Disconnect + reconnect | Disconnect + shutdown | Disconnect + shutdown |
| `code="516"` / `"401"` / `"409"` / conflict | Disconnect, disable reconnect | Disconnect + shutdown, disable reconnect | Disconnect + shutdown, disable reconnect |
| `code="429"` | Log + dispatch (no shutdown) | **Shutdown** | Log + dispatch (no shutdown) |
| `code="503"` | Log + dispatch (no shutdown) | **Shutdown** | Log + dispatch (no shutdown) |
| Unknown / no code (e.g. `<ack/>` body) | Log + dispatch (no shutdown) | **Shutdown + expected_disconnect=true** | Log + dispatch (no shutdown) |

---

## Tests

Five new unit tests in `src/client.rs` (existing test module).

### Encoder (Fix A)

- `test_build_ack_node_for_message_with_recipient_preserves_recipient` — message stanza with `recipient="146991363395800@lid"` produces an ACK that echoes the same value.
- `test_build_ack_node_for_receipt_with_recipient_preserves_recipient` — same for `<receipt>`.
- `test_build_ack_node_for_message_without_recipient_omits_recipient` — regression guard: never synthesise an empty `recipient`.

### Handler (Fix B)

- `test_stream_error_unknown_keeps_connection_alive` — bare `<stream:error/>` (no `code`, no body) leaves both `expected_disconnect` and `enable_auto_reconnect` untouched.
- `test_stream_error_ack_shaped_does_not_force_shutdown` — the exact shape seen in production (`<stream:error><ack class="message" type="text" id="…"/></stream:error>`) does not flip `expected_disconnect`.

All fictitious JIDs, per `AGENTS.md` ("No real PII in tests").

Local results:

```
$ cargo test --lib client::tests::test_build_ack_node
running 9 tests
test result: ok. 9 passed; 0 failed; 0 ignored

$ cargo test --lib client::tests::test_stream_error
running 6 tests
test result: ok. 6 passed; 0 failed; 0 ignored
```

Workspace-wide (`cargo test --workspace --exclude e2e-tests`) passes 511 + sub-crate tests with no regressions. `e2e-tests` is skipped because it requires the mock server (per `AGENTS.md`).

---

## Alignment with `whatsapp-rust/AGENTS.md`

| Rule | This PR |
|---|---|
| *Cross-reference whatsmeow, Baileys, and captured WhatsApp Web JS to verify implementations.* | Both fixes are direct ports of whatsmeow (`receipt.go:143-177` for A, `connectionevents.go:19-69` for B). Each change cites its whatsmeow line range in code comments. |
| *Collapsible if: always use let-chains instead of nested `if let`.* | New code uses the same `if let Some(_) = …` single-bind pattern already present; no nested `if let` introduced. |
| *No real PII in tests.* | All test JIDs are fictitious (`146991363395800@lid`, `155500012345:48@s.whatsapp.net`, etc.). |
| *State: never modify Device state directly. Use `DeviceCommand` + `PersistenceManager::process_command()`.* | Neither fix touches Device state. |
| *Errors: `thiserror` for typed errors, no `.unwrap()` outside tests.* | New code returns early via the existing `Result` paths; no new `.unwrap()` introduced outside tests. |
| *When adding comments to the code, don't be so verbose, also only explain why, not what.* | Both new comment blocks describe the WHY (server routing requirement, whatsmeow parity) — not the WHAT. |
| *Rust style — `cargo fmt --all`, `cargo clippy --all --tests`, `cargo test --all`.* | All three are clean. |

For reference: `NackInvalidHostedCompanionStanza = 493` already exists in `wacore/src/protocol/nack.rs`. That nack family is exactly what the server is signalling with the ack-shaped stream error before this fix — surfaced at the stream layer instead of as a nack.

---

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -3303,8 +3303,12 @@ impl Client {
                     info!("Got 503 service unavailable, will auto-reconnect.");
                 }
                 _ => {
+                    // Mirror whatsmeow's default: log + dispatch event, but keep the
+                    // connection alive. The server wraps per-stanza notifications in
+                    // <stream:error> without a code attribute (e.g. <ack/> for malformed
+                    // routing). Proactively disconnecting on those causes reconnect storms
+                    // under load — whatsmeow lets the noise socket detect a real teardown.
                     error!("Unknown stream error: {}", DisplayableNodeRef(node));
-                    self.expected_disconnect.store(true, Ordering::Relaxed);
                     self.core.event_bus.dispatch(Event::StreamError(
                         crate::types::events::StreamError {
                             code: code.to_string(),
@@ -3315,7 +3319,9 @@ impl Client {
             }
         }
 
-        // Single transport lock acquisition for all branches that need disconnect.
+        // Only tear down the connection for branches that explicitly opted in.
+        // 429/503/unknown match whatsmeow: log and let the socket layer notice
+        // if the server actually closes the stream.
         if should_disconnect {
             let transport_opt = self.transport.lock().await.clone();
             if let Some(transport) = transport_opt {
@@ -3325,10 +3331,9 @@ impl Client {
                     }))
                     .detach();
             }
+            info!("Notifying connection shutdown from stream error handler");
+            self.notify_connection_shutdown();
         }
-
-        info!("Notifying connection shutdown from stream error handler");
-        self.notify_connection_shutdown();
     }
 
     pub(crate) async fn handle_connect_failure(&self, node: &wacore_binary::NodeRef<'_>) {
@@ -3904,6 +3909,10 @@ fn encode_ack_bytes(
         let from_str = from_val.as_str();
         p_str.as_ref() != from_str.as_ref()
     });
+    // Server expects `recipient` echoed back so it can route the ack to the
+    // origin companion/device (hosted-companion, peer, LID-routed stanzas).
+    // Dropping it makes the server close the stream with `<stream:error><ack/>`.
+    let recipient_val = node.get_attr("recipient");
     let tag = node.tag.as_ref();
 
     let typ_val = if tag != "message" && !is_encrypt_identity_notification(node) {
@@ -3914,16 +3923,18 @@ fn encode_ack_bytes(
 
     let include_from = tag == "message" && own_device_pn.is_some();
 
-    // Count attrs: class + id + to + optional(from, participant, type)
+    // Count attrs: class + id + to + optional(from, participant, recipient, type)
     let attr_count = 3
         + usize::from(include_from)
         + usize::from(participant_val.is_some())
+        + usize::from(recipient_val.is_some())
         + usize::from(typ_val.is_some());
 
     struct AckNode<'a> {
         id: &'a wacore_binary::node::ValueRef<'a>,
         from: &'a wacore_binary::node::ValueRef<'a>,
         participant: Option<&'a wacore_binary::node::ValueRef<'a>>,
+        recipient: Option<&'a wacore_binary::node::ValueRef<'a>>,
         typ: Option<&'a wacore_binary::node::ValueRef<'a>>,
         own_pn: Option<&'a Jid>,
         tag_str: &'a str,
@@ -3958,6 +3969,10 @@ fn encode_ack_bytes(
                 enc.write_string("participant")?;
                 p.encode_value(enc)?;
             }
+            if let Some(r) = self.recipient {
+                enc.write_string("recipient")?;
+                r.encode_value(enc)?;
+            }
             if let Some(t) = self.typ {
                 enc.write_string("type")?;
                 t.encode_value(enc)?;
@@ -3976,6 +3991,7 @@ fn encode_ack_bytes(
         id: id_val,
         from: from_val,
         participant: participant_val,
+        recipient: recipient_val,
         typ: typ_val,
         own_pn: if include_from { own_device_pn } else { None },
         tag_str: tag,
@@ -3999,13 +4015,14 @@ fn build_ack_node(node: &wacore_binary::NodeRef<'_>, own_device_pn: Option<&Jid>
         .get_attr("participant")
         .filter(|p| p.as_str().as_ref() != from_ref.as_str().as_ref())
         .map(|v| v.to_node_value());
+    let recipient = node.get_attr("recipient").map(|v| v.to_node_value());
     let tag = node.tag.as_ref();
     let typ = if tag != "message" && !is_encrypt_identity_notification(node) {
         node.get_attr("type").map(|v| v.to_node_value())
     } else {
         None
     };
-    let mut attrs = Attrs::with_capacity(6);
+    let mut attrs = Attrs::with_capacity(7);
     attrs.insert("class", NodeValue::from(tag));
     attrs.insert("id", id);
     attrs.insert("to", from);
@@ -4016,6 +4033,9 @@ fn build_ack_node(node: &wacore_binary::NodeRef<'_>, own_device_pn: Option<&Jid>
     }
     if let Some(p) = participant {
         attrs.insert("participant", p);
+    }
+    if let Some(r) = recipient {
+        attrs.insert("recipient", r);
     }
     if let Some(t) = typ {
         attrs.insert("type", t);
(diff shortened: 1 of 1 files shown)


jlucaso1:
Thanks

PR #635, opened 2026-05-19
fix: self-DM/sibling decryption deadlock (retry shape + session recovery + peer pkmsg identity)

jlucaso1:
## Summary

Multiple layered fixes for the self-DM / linked-device decryption deadlock that pinned `236395184570386@lid.0` in prod to a stale ratchet (`574b6be3…` counter 940+). Confirmed end-to-end recovery in prod within ~3 seconds of deploying the last commit — Android re-synced sessions and the bot's been decrypting cleanly since.

The investigation peeled back ~8 distinct issues. Each commit fixes one and includes a falsifiable regression test.

### 1. DM retry stanza shape (`cda07efd`)
`prepare_dm_retry_stanza` was emitting the fanout-shaped envelope (`<message><participants><to jid=…><enc>`) for single-device retries. Server rejected every resend with `<ack error="479"/>` (SmaxInvalid). WA Web's `WAWebSendMsgCreateDeviceStanza.createUserDeviceMsgStanza` puts `<enc>` directly under `<message>` plus a `recipient` attribute.

### 2. Receipt's `recipient` forwarded verbatim (`62dd6f7e`)
`recipient_jid` on the resend was derived from `info.chat` — wrong on the alt PN↔LID namespace path. WA Web's `WAWebHandleRetryRequest`: `f && (k.recipient = f)` — propagate the receipt's `recipient` attribute as-is.

### 3. `should_recreate_session` whatsmeow parity (`c31f06db`)
When a retry receipt arrives without `<keys>` for a peer we already have a session with, force a session recreate (fresh prekey fetch + `process_prekey_bundle`) when `retryCount ≥ 2` and >1h since the last recreate for that JID. Mirrors whatsmeow's `retry.go:158-175`.

### 4. PN session wins on LID migration conflict (`712f5e49`)
`migrate_signal_sessions_on_lid_discovery` deleted PN when both PN and LID slots had sessions — but the LID side was often the freshly built stub, not the working ratchet. Switched to whatsmeow's `MigratePNToLID` policy: `ON CONFLICT … session=excluded.session` — PN overwrites the LID stub before being dropped.

### 5. Transactional decrypt (`bf691f5a`)
`decrypt_message_with_state` advanced the receiver chain via `set_receiver_chain_key` BEFORE `verify_mac`. Each MAC failure committed the chain step regardless, so a string of failed attempts walked the chain past the peer's actual position; from there the deadlock was permanent. Now clones into a scratch `SessionState`, mutates the scratch, commits `*state = scratch` only after MAC + plaintext both verify.

### 6. PN→LID migration on inbound BadMac (`53c98e05`)
The inbound decrypt path only attempted `try_pn_to_lid_migration_decrypt` on `InvalidPreKeyId`. When a fresh LID-keyed session shadowed the working PN-keyed session, BadMac fired forever without ever migrating the PN blob into the LID slot. Extended the migration attempt to the BadMac/InvalidMessage path too.

### 7. PDO target is LID when LID-migrated (`234b9224`)
PDO peer messages were addressed to PN (matching WA Web's `getMeDevicePnOrThrow_DO_NOT_USE` — the "DO_NOT_USE" suffix flags it as un-migrated tech debt). The phone's post-migration Signal store keys by LID, so PN-addressed pkmsg landed in the wrong slot and never refreshed its outbound chain. Switched to whatsmeow's policy via `Store.GetJID().ToNonAD()` — LID when known.

### 8. Peer pkmsg `<meta>` + `<device-identity>` (`8a54365f`) ← the actual unblock
The fix that finally restored prod. `prepare_peer_stanza` was emitting only `<enc>` inside `<message category="peer">`. Whatsmeow's `preparePeerMessageNode` builds `[<meta appdata="default"/>, <enc>, <device-identity>?]` — the `<device-identity>` element (the proto-encoded `ADVSignedDeviceIdentity`) is what the primary phone uses to verify the linked device's identity before promoting a Signal session on the pkmsg path. Without it the phone XMPP-acks the stanza but the Signal layer silently rejects the pkmsg, its outbound ratchet stays on the diverged chain, and the linked device never catches up.

## Verification

- `cargo fmt --all`, `cargo clippy --workspace --tests` clean.
- `cargo test --workspace --exclude e2e-tests` green (522 whatsapp-rust + 11 session_divergence + 748 wacore + the rest).
- New regression tests cover every layer:
  - `dm_retry_emits_enc_directly_under_message_with_recipient` — retry stanza shape.
  - `resolve_retry_chat_info_forwards_recipient_attribute_verbatim` — receipt `recipient` propagation incl. alt-namespace.
  - `should_recreate_session_matrix` — the four-state transition table for the throttle.
  - `migration_preserves_working_session_when_both_namespaces_present` — PN wins on conflict.
  - `failed_mac_must_not_advance_receiver_chain` — 10 tampered ciphertexts, chain index stays put.
  - `test_badmac_migrates_pn_session_when_lid_shadow_exists` — BadMac→migration→decrypt success.
  - `self_peer_target_prefers_lid_when_present` / `_falls_back_to_pn_without_lid` / `_errors_when_no_identity_known`.
  - `peer_pkmsg_includes_meta_and_device_identity` + `peer_pkmsg_omits_device_identity_when_account_missing`.
- Plus 9 hypothesis tests in `wacore/libsignal/tests/session_divergence.rs` mapping out what does and doesn't recover at the libsignal layer.
- Deployed to prod (image `8a54365f`). Recovery within ~3s of first inbound msg from the previously stuck peer. 57 successful decrypts since, normal message flow restored (reactions + text, fanout including `<device-identity>` on every pkmsg).

## Test plan

- [x] Workspace test suite green excluding e2e (mock server lags behind the new DM retry shape; suite tracked separately).
- [x] Prod redeploy + observed recovery on the originally-stuck peer.
- [x] WA Web vs whatsmeow cross-reference for every behavioral change. Where the two diverge, the commit message names which one we picked and why.

## Co-authorship

Codex helped pin the inbound `BadMac` → `try_pn_to_lid_migration_decrypt` extension (commit `53c98e05`) after I'd burned several rounds chasing the wrong layer.

### DIFF
--- a/examples/benchmark.rs
+++ b/examples/benchmark.rs
@@ -1,6 +1,8 @@
 use chrono::Local;
-use log::{error, info};
+use log::{error, info, warn};
+use std::collections::HashMap;
 use std::sync::Arc;
+use wacore::net::{HttpClient, HttpRequest};
 use wacore::proto_helpers::MessageExt;
 use wacore::store::InMemoryBackend;
 use wacore::types::events::Event;
@@ -9,6 +11,24 @@ use whatsapp_rust::TokioRuntime;
 use whatsapp_rust::bot::{Bot, MessageContext};
 use whatsapp_rust_tokio_transport::TokioWebSocketTransportFactory;
 use whatsapp_rust_ureq_http_client::UreqHttpClient;
+
+/// Derive the mock-server admin scan-qr endpoint from a `ws[s]://host:port/...`
+/// WebSocket URL. Same host/port, scheme `ws`→`http` / `wss`→`https`, path
+/// `/admin/mock-phone/scan-qr`. Mirrors `tests/e2e/src/lib.rs`. Returns `None`
+/// for URLs that don't match the ws scheme — the autoresponder would
+/// no-op on real WhatsApp anyway, but skipping the POST keeps logs clean.
+fn mock_admin_scan_qr_url(ws_url: &str) -> Option<String> {
+    let http_scheme = if ws_url.starts_with("wss://") {
+        "https://"
+    } else if ws_url.starts_with("ws://") {
+        "http://"
+    } else {
+        return None;
+    };
+    let after_scheme = ws_url.split("://").nth(1)?;
+    let host_port = after_scheme.split('/').next()?;
+    Some(format!("{http_scheme}{host_port}/admin/mock-phone/scan-qr"))
+}
 
 fn main() {
     env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("warn"))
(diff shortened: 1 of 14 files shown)


jlucaso1:
@claude review

PR #636, opened 2026-05-21
fix(send): align own devices to LID namespace for LID-addressed DMs

oonid:
Fixes ACK 400 on DMs to LID-addressed recipients.

Root cause: own devices are fetched via the bot's PN JID, so they come back PN addressed (628xxx@s.whatsapp.net). The server rejects a stanza that mixes PN and LID participants, so including a PN own device in a LID stanza gets the whole stanza rejected with ACK error 400.

Fix: when the recipient is LID, convert the remaining own PN devices to their LID equivalents before dedup, so the stanza is uniformly LID addressed. This matches whatsmeow, which switches ownID to LID before GetUserDevices so own devices come back in the right namespace.

### DIFF
--- a/src/client/device_registry.rs
+++ b/src/client/device_registry.rs
@@ -253,7 +253,7 @@ impl Client {
     ///
     /// Removes all device registry cache entries (all LID/PN aliases) so the
     /// next lookup falls through to the database or network.
-    pub async fn invalidate_device_cache(&self, user: &str) {
+    pub(crate) async fn invalidate_device_cache(&self, user: &str) {
         let lookup = self.resolve_lookup_keys(user).await;
 
         for key in lookup.all_keys() {
@@ -363,7 +363,7 @@ impl Client {
     /// Delete Signal sessions for specific device IDs under both LID and PN
     /// addresses, then flush. Shared by `clear_device_record` and
     /// `patch_device_remove`.
-    pub(crate) async fn delete_sessions_for_devices(&self, user: &str, device_ids: &[u16]) {
+    async fn delete_sessions_for_devices(&self, user: &str, device_ids: &[u16]) {
         let lookup = self.resolve_lookup_keys(user).await;
         let servers = [wacore_binary::Server::Lid, wacore_binary::Server::Pn];
         for server in servers {
--- a/src/client/device_registry.rs
+++ b/src/client/device_registry.rs
@@ -253,7 +253,7 @@ impl Client {
     ///
     /// Removes all device registry cache entries (all LID/PN aliases) so the
     /// next lookup falls through to the database or network.
-    pub(crate) async fn invalidate_device_cache(&self, user: &str) {
+    pub async fn invalidate_device_cache(&self, user: &str) {
         let lookup = self.resolve_lookup_keys(user).await;
 
         for key in lookup.all_keys() {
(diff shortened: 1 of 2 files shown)


oonid:
Hey @jlucaso1 , would appreciate a review when you have a moment!

This fixes ACK 400 on DM sends to LID-addressed recipients. 
The root cause: `own_cached` is fetched via the bot's PN JID so it always returns PN-format JIDs, but the server requires all participants in a stanza to share the same namespace.
Including a `628xxx@s.whatsapp.net` participant in a LID-addressed stanza causes the entire stanza to be rejected.

Fix: after the sender-device retain filter, convert any remaining own PN devices to their LID equivalents when `recipient_is_lid`.

jlucaso1:
Thanks for the contribution.

> ACK 400 handler (spawn_phash_validation) When a DM ACK arrives with error="400", clears Signal sessions for own device 0 under both LID and PN namespaces and invalidates the device cache. Does not retry — a blind retry with the same fanout produces an infinite loop. The namespace conversion below is the primary prevention; this handler cleans up residual stale state.

This appears to be hacky, do you have any evidence/log that this is the correct approch? Some whatsmeow or whatsapp web source code?

oonid:
> This appears to be hacky, do you have any evidence/log that this is the correct approch? Some whatsmeow or whatsapp web source code?

Thank you for your time @jlucaso1 .

Fair point, the spawn_phash_validation ACK 400 branch (session clear + cache invalidate) is not backed by whatsmeow or WA Web source code. 
I added it as defensive cleanup reasoning that a 400 on a DM likely means the server rejected the participant list, and stale sessions for own device 0 might be contributing. That reasoning is speculative.

Whatsmeow (send.go:445-446) simply wraps the error and returns it to the caller, no session cleanup at all on ACK 400. There is no captured WA Web JS available to cross-reference either.
The part that is consistent with whatsmeow is the primary prevention: ensuring namespace uniformity before building the stanza. Whatsmeow does this at send.go:346 by switching ownID = cli.getOwnLID() so GetUserDevices returns LID-format JIDs for own devices. The Rust equivalent is the ok_or_else fail-fast when own_lid is None, same intent, different mechanism.

I'm happy to drop the ACK 400 session-clearing block if you'd prefer to keep the diff minimal and evidence-based. The phash mismatch handler below it (cache invalidation) is closer to what whatsmeow does (send.go:460-462) and could stand on its own.

tmigone:
I'm hitting LID <> PN problems with latest from `main`, this PR fixed it for me.

jlucaso1:
Thanks for digging into this, your root cause is right. Own devices come back PN addressed and the server rejects a stanza that mixes PN and LID participants, so the LID conversion is the correct fix and it matches whatsmeow switching ownID to LID before fanout.

I pushed a commit trimming the PR down to just that conversion. I dropped the rest after checking against WA Web and whatsmeow:

- device 0 exclusion: getFanOutList only excludes the current sender device, never the primary. Dropping device 0 stops the DeviceSentMessage from reaching the primary phone, so messages sent from the companion stop syncing there. The LID conversion already makes device 0 a valid participant, so the exclusion is not needed.
- ACK 400 session clearing: no equivalent in whatsmeow or WA Web, and with the namespace fixed the 400 should not happen in the first place.
- tc token changes: the AB prop default is false in WA Web, so defaulting to true diverges. The perpetual 463 looks like a separate problem with ab_props starting empty, better handled on its own.

Build, clippy and the unit tests are green. Let me know if you disagree on any of these.

PR #641, opened 2026-05-27
fix(prekeys): re-upload after re-pair to heal stale server bundle

Salientekill:
## Problem

A long-lived bot that re-pairs (or otherwise reuses a persisted store) can end up with the server holding a prekey bundle that no longer matches local key material. Peers that fetch the stale bundle then BadMac forever.

Root cause: `handle_pair_success` never resets `server_has_prekeys`. After a re-pair the persisted `server_has_prekeys=true` from the prior pairing is still there, so the next connect's `upload_pre_keys_at_login` short-circuits and skips the upload. The server keeps the old bundle.

## Fix

`src/pair.rs`: on pair-success, reset `server_has_prekeys=false` so the next connect re-uploads. This mirrors WA Web, where a freshly registered device always uploads its prekeys (its storage starts with the flag unset). It is a no-op on first pairing since the flag already defaults to false.

`src/prekeys.rs` and `src/handlers/notification.rs`: `validate_digest_key` now takes `prekey_upload_lock` internally instead of the caller holding it. This serializes the digest 404 re-upload with the login, prekey-low and refresh paths for both callers. The background-init path in `client.rs` was calling validate without the lock before, so its 404 re-upload could race on prekey id allocation. The external lock in `handle_digest_key` is removed, since holding it there plus the internal acquire would deadlock (`async_lock::Mutex` is not reentrant).

digestKey behavior stays exactly like WA Web's `WAWebDigestKeyJob.digestKey()`: re-upload only on 404, log and skip on regId / hash / missing-prekey mismatch. For the restore or partial-backup recovery case, `refresh_pre_keys()` already exists as an explicit path.

## Risk

No new IQ types and no peer-facing traffic. Worst case is one extra prekey upload on the connect right after a re-pair, which is the same upload a fresh device already does. The digest path is unchanged from main except for the added lock.

## Test plan

- cargo fmt, cargo clippy --all --tests, and the non-e2e tests are green

### DIFF
--- a/src/handlers/notification.rs
+++ b/src/handlers/notification.rs
@@ -266,7 +266,7 @@ async fn handle_prekey_low(client: &Arc<Client>) {
 ///
 /// Matches WA Web's `WAWebHandleDigestKey`:
 /// Queries server for key bundle digest, validates SHA-1 hash locally,
-/// re-uploads if mismatch or missing.
+/// re-uploads only when the server has no record (404).
 ///
 /// `validate_digest_key` owns `prekey_upload_lock` acquisition internally, so
 /// any upload it triggers stays serialized with `upload_pre_keys_at_login`,
--- a/src/handlers/notification.rs
+++ b/src/handlers/notification.rs
@@ -268,14 +268,16 @@ async fn handle_prekey_low(client: &Arc<Client>) {
 /// Queries server for key bundle digest, validates SHA-1 hash locally,
 /// re-uploads if mismatch or missing.
 ///
-/// Acquires `prekey_upload_lock` to serialize with the count-based upload path,
-/// preventing concurrent uploads that could race on prekey ID allocation.
+/// `validate_digest_key` owns `prekey_upload_lock` acquisition internally, so
+/// any upload it triggers stays serialized with `upload_pre_keys_at_login`,
+/// `handle_prekey_low`, and `refresh_pre_keys` without this caller needing to
+/// (and indeed, holding it here would deadlock — `async_lock::Mutex` is not
+/// reentrant).
 fn handle_digest_key(client: &Arc<Client>) {
     let client_clone = client.clone();
     client
         .runtime
         .spawn(Box::pin(async move {
-            let _guard = client_clone.prekey_upload_lock.lock().await;
             if let Err(e) = client_clone.validate_digest_key().await {
                 warn!("Digest key validation failed: {:?}", e);
             }
--- a/src/pair.rs
+++ b/src/pair.rs
@@ -263,12 +263,10 @@ async fn handle_pair_success<'a>(
                 )))
                 .await;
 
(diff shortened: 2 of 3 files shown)


Salientekill:
Addressed @coderabbitai's race-condition finding in `78b373a` — `validate_digest_key` now acquires `prekey_upload_lock` once at function entry, before the digest IQ. This serializes the function with `upload_pre_keys_at_login`, `handle_prekey_low`, and `refresh_pre_keys`, and also closes the pre-existing gap in the 404 branch that was already there.

Local validation:
- `cargo test --workspace --exclude e2e-tests` → 1.6k+ tests, 0 failed
- `cargo build --release` clean

The two CI failures shown above are unrelated to this PR:
- `claude-review`: external contributor permission gate (`Actor does not have write permissions`)
- `Integration Benchmark (PR)`: workflow YAML error (`Unexpected value ''` at `bench-integration.yml:32`), likely a secret unavailable on fork PRs

Salientekill:
Good catch @coderabbitai — that was a real deadlock. `async_lock::Mutex` isn't reentrant, so my previous commit (`78b373a`) plus the existing outer `let _guard = client_clone.prekey_upload_lock.lock().await` in `handle_digest_key` would have hung the digest-notification path indefinitely.

Fixed in `87fd5f6` by removing the outer acquisition from `handle_digest_key` and consolidating ownership inside `validate_digest_key`. This also fixes the pre-existing imbalance where `bg_client.validate_digest_key()` in `client.rs` (the background-init call site) was *not* holding the lock, leaving its 404 upload path racy.

`cargo test --workspace --exclude e2e-tests` → 1630 tests, 0 failed.

jlucaso1:
Thanks for digging into this. I pushed a commit on top that keeps most of your fix and reverts just the digestKey part to stay aligned with WA Web.

Kept: the pair.rs reset (it matches how WA Web ends up uploading prekeys after a fresh registration, since a freshly registered device has the flag unset) and moving the lock inside validate_digest_key. The lock move also fixes a real race, the background-init path in client.rs was calling validate_digest_key without holding the lock before, so its 404 re-upload could race with the login/prekey-low uploads on prekey id allocation.

Reverted: forcing a re-upload on regId / hash / missing-prekey mismatch. WA Web's digestKey wraps validateLocalKeyBundle in a try/catch and only re-uploads on a 404 (no server record), it deliberately swallows the validation failures and lets RotateKeyJob handle refresh. Since we run validate_digest_key on every connect, forcing an upload on a content mismatch would re-upload 812 fresh keys on every connect whenever the divergence doesn't immediately converge, and it risks pushing local state over a server bundle that might actually be the good one. For the restore or partial-backup case there's already refresh_pre_keys() as an explicit recovery path, which feels like the safer home for that.

So digestKey now logs and skips on those mismatches and uploads only on 404, exactly like WA Web. fmt, clippy and the non-e2e tests are green.

PR #642, opened 2026-05-27
fix(polls): align poll vote encryption/decryption to conversation addressing

jlucaso1:
## Summary
Poll votes always used the own **PN** as the voter JID. But the voter feeds the poll-vote HKDF info + AAD, so it must match the addressing of the poll creator. In LID-addressed conversations the real WhatsApp Web (`WAWebAddonEncryption`) authors the vote under the **LID** — so PN-authored votes fail to decrypt on the host as contacts migrate to LID.

This brings whatsapp-rust in line with WA Web behavior (verified against `docs/captured-js/WAWeb/Addon/Encryption.js`), mirroring the concept of whatsmeow `5d16908e` (encrypt) and `e46d104e`/`1c97bf40` (decrypt fallback). The decrypt path reuses the same LID↔PN fallback pattern the message-edit path (`wacore/src/message_edit.rs`) already implements.

- **encrypt** (`Polls::vote`): pick the voter JID to match the poll creator's namespace — own LID for LID-addressed polls, own PN otherwise — falling back to PN (with a warning) when the own LID is unknown.
- **decrypt** (`Polls::decrypt_vote` / `aggregate_votes`): retry under the opposite addressing (LID↔PN) when a counterpart mapping is known, matching WA Web's `decryptAddOn` homogeneous LID/PN attempts. The fallback pair is built only when **both** the creator and the voter have a known counterpart.
- new `wacore::poll::decrypt_poll_vote_with_fallback` + `PollVoteAddressing`.

### ⚠️ Breaking change
`Polls::decrypt_vote` and `Polls::aggregate_votes` changed from sync associated functions to `&self async` methods. Call sites must use `client.polls().decrypt_vote(...).await`. No internal callers in this repo.

## Test plan
- [x] `cargo fmt --all`
- [x] `cargo clippy --all --tests` (clean)
- [x] `cargo test --workspace --exclude e2e-tests` (0 failures)
- [x] wacore unit tests: fallback recovers across addressing, primary-without-fallback, combined error, `None` propagation
- [x] client tests: voter is PN/own-LID/PN-fallback; decrypt + aggregate recover when fed LID but encrypted under PN; no-mapping must not silently decrypt

### DIFF
--- a/src/features/polls.rs
+++ b/src/features/polls.rs
@@ -119,7 +119,11 @@ impl<'a> Polls<'a> {
             .await
             .ok_or_else(|| anyhow!("Not logged in — cannot determine own JID"))?;
         let my_base = my_jid.to_non_ad();
-        let voter_jid_str = my_base.to_string();
+
+        let voter_jid = self
+            .resolve_voter_jid(poll_creator_jid, &my_base, poll_msg_id)
+            .await;
+        let voter_jid_str = voter_jid.to_string();
         let creator_jid_str = poll_creator_jid.to_non_ad().to_string();
 
         let selected_hashes: Vec<Vec<u8>> = option_names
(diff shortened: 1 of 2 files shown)


jlucaso1:
@claude review

PR #647, opened 2026-05-27
fix(offline): drain offline queue by acking duplicate and undecryptable messages

jlucaso1:
## Summary
Every received message is now acknowledged so the server drains it from the offline queue instead of replaying it on each reconnect (which eventually force-closed the stream).

- Ack duplicates (already-processed) and own-account fan-out. Own non-peer messages get a transport ack (correct `to`=`from`) since their delivery receipt is suppressed.
- Send a transport ack on decrypt failure and on `<unavailable>`; the retry receipt alone does not clear the queue.
- Nack generic group decrypt errors, matching the session catch-all.
- Send the retry receipt, PDO request and transport ack in one ordered, flushed task so a disconnect cannot clear a stanza before its resend request goes out.
- Send delivery receipts as `type="inactive"` unless marked online (presence available or forced), matching whatsmeow's background companion.
- Raise `message_retry_counts` TTL to 1h so the retry cap survives spaced redeliveries.
- Lower `decrypt-fail="hide"` failures to DEBUG.
- Recognize `stream:error` carrying an `ack` child instead of logging it as unknown.

## Test plan
- [ ] `cargo test --workspace --exclude e2e-tests`
- [ ] `cargo clippy --all --tests`
- [ ] `cargo fmt --all --check`

### DIFF
--- a/src/cache_config.rs
+++ b/src/cache_config.rs
@@ -170,7 +170,8 @@ pub struct CacheConfig {
     /// Default: capacity 0 (disabled — DB-only, matching WA Web).
     /// Set capacity > 0 to enable a fast in-memory cache in front of the DB.
     pub recent_messages: CacheEntryConfig,
-    /// Message retry counts (time_to_live). Default: 5m TTL, 500 entries.
+    /// Message retry counts (time_to_live). Default: 1h TTL, 500 entries.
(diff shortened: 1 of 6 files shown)


jlucaso1:
@claude review

jlucaso1:
@claude review

PR #664, opened 2026-05-30
fix(message): ack SKDM-only session decrypts

jlucaso1:
## What

Session decrypt now tracks whether a successfully decrypted plaintext actually reached `dispatch_parsed_message`. SKDM-only session plaintext still processes the sender-key distribution and skips user dispatch, but `process_decrypted_message` now emits the normal message confirmation when that skip would otherwise leave the stanza unacked.

This replaces the anonymous `(bool, bool, bool)` session result with `SessionBatchOutcome` and propagates the dispatch bit through direct decrypt, untrusted-identity retry, and PN→LID migration via `MigrationDecryptOutcome`.

The ack path itself is unchanged: `ack_received_message` still chooses delivery receipt, sender receipt, or transport ack from the message shape.

## Why

A production group stanza (`AC1F6388BDE0B721870652C378CFEDDB`) decrypted successfully as a session `msg`, carried only a Sender Key Distribution Message, and had no `skmsg` sibling. Because SKDM-only messages skipped dispatch, they also skipped the only success-path ack. The server retained the stanza until it later sent `<stream:error><ack class="message" id=...>`, re-delivered it, and the already-advanced Signal ratchet then hit BadMac.

whatsmeow and WA Web acknowledge every successfully decrypted message; the decision is based on decrypt result, not user-visible content.

## Tests

Added regression coverage with real Signal/session and sender-key ciphertexts, decoding captured outbound frames by message id:

- SKDM-only group session decrypt emits exactly one group delivery receipt and no `Event::Message`
- SKDM session + successful `skmsg` sibling emits exactly one receipt and dispatches only the `skmsg` content
- Own-device group SKDM-only session emits exactly one transport `<ack class="message">`, not a `type="sender"` receipt

Validation:

- `cargo fmt --all`
- `cargo build -p whatsapp-rust`
- `cargo clippy --all-targets -- -D warnings`
- `cargo test -p whatsapp-rust` (617 lib + 18 discrepancy + 7 handshake + 5 doctests passed)
- `cargo test -p whatsapp-rust skdm -- --nocapture` (11 passed)

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -70,6 +70,20 @@ pub(crate) struct ClassifiedMessage {
     pub decrypt_fail_mode: crate::types::events::DecryptFailMode,
 }
 
+#[derive(Clone, Copy, Debug, Default)]
+struct SessionBatchOutcome {
+    decrypted: bool,
+    duplicate: bool,
+    undecryptable: bool,
+    dispatched: bool,
+}
+
+#[derive(Clone, Copy, Debug, Default)]
+struct MigrationDecryptOutcome {
+    decrypted: bool,
+    dispatched: bool,
+}
+
 /// Retry count threshold for logging high retry warnings.
 /// WhatsApp Web logs metrics when retry count exceeds this value.
 const HIGH_RETRY_COUNT_THRESHOLD: u8 = 3;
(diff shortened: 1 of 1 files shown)


jlucaso1:
@claude review

jlucaso1:
@claude review

PR #668, opened 2026-05-31
feat(msg-secret)!: bound messageSecret retention by policy and event-time horizon

jlucaso1:
## What

Bounds the per-message `messageSecret` store, which a recent change (#665) made grow without limit: it seeds a secret for every non-forwarded history-sync message and captures one for every live message and every send, and nothing ever prunes them. On a freshly paired headless account that produced ~95k rows accounting for ~99% of the SQLite file.

Retention is now one policy plus a per-row, event-time deadline, all on `CacheConfig`:

- **`msg_secret_policy: MsgSecretPolicy`** — `Managed` (default), `BotOnly`, `Full`, `Disabled` — unifies the three previously-scattered decisions (capture on live receive, seed from history, prune).
- **Per-row `expires_at`** (absolute unix seconds, `0` = never) replaces the old `created_at`-based TTL. `created_at` was insert time and was overwritten on every conflict-update and edit re-persist, so no age-based prune was ever sound. The deadline is computed from the **parent message's own event time** plus a per-add-on-kind horizon, and on conflict the later deadline wins (`0` = never wins) so a redelivery or edit re-persist never shortens a window.
- **`msg_secret_retention`** horizons: text/`MESSAGE_EDIT` 30d, poll/event 90d, bot 30d. An edit is only valid when authored within 20 min of the parent, but the offline queue can deliver that (validly-authored) edit up to ~30 days later, so 30d is the floor that keeps a 30-day-offline receiver able to decrypt. `PollAddOption`/`EventEdit` have no window and ride the parent's secret, so poll/event parents get the longer 90d horizon.
- **Edit-processing window on receive** — the parent's event time is now stored (`message_ts`) so the receive path can drop a secret-encrypted `MESSAGE_EDIT` authored outside the window (`editTs >= parentTs + 1200s`), matching WhatsApp Web's `ProcessEditProtocolMsgs`. The check is on authored time (not "now"), so a validly-authored edit still applies after an offline gap; it is permissive when the parent's time is unknown (resolver-supplied secrets carry none).
- **Age/type-filtered seed** — history secrets whose parent is already past its horizon are dropped at seed time. For a multi-month history this collapses the pairing burst to the recent, still-relevant slice.
- **`seed_msg_secrets_from_history: bool`** (default `true`) — live capture (covering everything received after connect) is independent from history seeding. Seeding only matters for the boundary case: an add-on that arrives live yet references a parent delivered via history sync. Headless consumers that only react to new messages can flip this off.
- **`original_message_resolver`** (+ `msg_secret_resolver_timeout`, default 5s) — an app-supplied fallback consulted when an add-on's parent secret is absent from the store (and its LID/PN alternates), bounded by a timeout because it runs in the per-chat receive lane. This is what lets the `Disabled` policy still decrypt: an app that keeps its own message store can own retention entirely and answer secret lookups on demand. (`Send + Sync` is dropped on wasm32 so a wasm app can supply a `!Send` resolver.)

Newsletters arrive as plaintext and never carry a `secret_encrypted_message`, so there is no newsletter retention class (documented inline to avoid a future dead branch).

## Why

The secrets only unlock add-ons that reference a parent by id — secret-encrypted edits, msmsg bot replies, and poll/event edits — all of which are bounded in time. The send path never reads the store, and poll votes/reactions take an app-supplied secret rather than reading it, so the only real consumers are five inbound add-on kinds with finite useful lifetimes. Keeping a secret for every message forever (and re-seeding the entire history on pairing) is therefore pure storage cost with no decrypt benefit beyond the horizon.

The redesign is grounded in verified behavior: the edit window is 1200s (20 min), enforced on both send and receive as an *authored-time* check (`editTs < parentTs + window`) — not relative to "now" — so an offline-delivered but validly-authored edit still applies, which is exactly why retention is sized to the ~30-day delivery window rather than the 20-minute authoring window. PollAddOption/EventEdit have no window at all, and edits by other users (newsletter admins) still key the lookup on the original author's secret, so the existing keying already covers them.

## Breaking Change

BREAKING CHANGE: `CacheConfig.msg_secret_ttl_secs: u64` is removed. Replace it with `msg_secret_policy: MsgSecretPolicy` (default `Managed`) plus the additive `msg_secret_retention`, `seed_msg_secrets_from_history` (default `true`), `original_message_resolver`, and `msg_secret_resolver_timeout`. The default behavior changes from "seed and keep everything forever" to "seed only the still-relevant slice and prune by event-time horizon".

Migration:

- To keep the previous behavior exactly, set `msg_secret_policy: MsgSecretPolicy::Full`.
- To restore pre-#665 capture (bot contexts only — including group bot invocations), use `MsgSecretPolicy::BotOnly`.
- To persist nothing and own retention in the app, use `MsgSecretPolicy::Disabled` and register an `original_message_resolver`.
- `MsgSecretStore::put_msg_secrets` is now the required method and carries `MsgSecretEntry { expires_at, message_ts }`; the scalar `put_msg_secret` is a never-expiring default wrapper, and `get_msg_secret_with_ts` has a default returning `0` for the parent time. Custom backends that implemented the scalar should move to `put_msg_secrets` and honor `expires_at` (keep the later deadline on conflict, `0` wins) and `message_ts` (keep the non-zero value).

Existing installs are not mass-deleted: the SQLite migration backfills legacy rows to `created_at + 30d`, so the prior pairing-burst seed ages out gradually once pruning runs (which now includes `Disabled`, so switching to it also reclaims legacy rows — they drain as each row hits its backfilled deadline, completing within ~30 days, not all at once on the first sweep).

## Tests

New coverage:

- seed filter: `Managed` drops text past 30d but keeps recent text and polls within 90d; `Full` seeds an ancient secret; `Disabled` seeds nothing; the opt-out flag skips seeding under `Managed`
- seeded rows carry an `expires_at` derived from the parent's `messageTimestamp` — a prune past `msg_ts + 30d` removes the row while an earlier prune keeps it
- edit-processing window: an in-window edit applies, an out-of-window edit is dropped, and an unknown parent timestamp stays permissive; `get_msg_secret_with_ts` round-trips and keeps the parent ts across a 0-ts redelivery
- `BotOnly` keeps a group bot prompt (botMetadata) but skips a plain message; unit cover for `is_bot_context` / `classify_from_flags` / policy predicates
- `OriginalMessageResolver`: a secret-encrypted edit decrypts via the resolver when the store is empty under `Disabled`
- SQLite: conflict keeps the later deadline (`0`/never wins); prune deletes only passed deadlines

Validation:

- `cargo fmt --all`
- `cargo clippy -p wacore -p whatsapp-rust -p whatsapp-rust-sqlite-storage --all-targets -- -D warnings` clean
- `cargo test -p wacore -p whatsapp-rust -p whatsapp-rust-sqlite-storage` — 819 (wacore lib) + 644 (whatsapp-rust lib) + 31 (sqlite) passing

### DIFF
--- a/src/appstate_sync.rs
+++ b/src/appstate_sync.rs
@@ -233,14 +233,11 @@ mod tests {
     #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
     #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
     impl MsgSecretStore for MockBackend {
-        async fn put_msg_secret(
+        async fn put_msg_secrets(
             &self,
-            _chat: &str,
-            _sender: &str,
-            _msg_id: &str,
-            _secret: &[u8],
-        ) -> StoreResult<()> {
-            Ok(())
+            entries: Vec<wacore::store::traits::MsgSecretEntry>,
+        ) -> StoreResult<usize> {
+            Ok(entries.len())
         }
 
         async fn get_msg_secret(
--- a/src/cache_config.rs
+++ b/src/cache_config.rs
@@ -210,6 +210,17 @@ pub struct CacheConfig {
     pub msg_secret_policy: MsgSecretPolicy,
     /// Per-add-on-kind retention horizons applied under `Managed`/`BotOnly`.
     pub msg_secret_retention: MsgSecretRetention,
+    /// Whether to seed `messageSecret`s from history-sync blobs. Default `true`.
+    ///
+    /// Independent of live capture (which `msg_secret_policy` governs): seeding
+    /// only matters for add-ons that arrive live after connect yet reference a
+    /// parent delivered via history sync — edits of just-pre-pairing messages,
+    /// add-options/edits on still-open polls, or replays to a reconnecting
+    /// offline device. Headless consumers that only react to new messages can
+    /// set this to `false` to skip the pairing-time seed entirely. When `true`,
+    /// the policy still filters the seed (age/type under `Managed`, bot-only
+    /// under `BotOnly`, everything under `Full`).
+    pub seed_msg_secrets_from_history: bool,
     /// Optional app-supplied fallback consulted when an add-on's parent secret
     /// is absent from the store (and its LID/PN alternates). Lets an app that
     /// keeps its own message store own secret retention; required for the
@@ -247,6 +258,10 @@ impl std::fmt::Debug for CacheConfig {
             .field("sent_message_ttl_secs", &self.sent_message_ttl_secs)
             .field("msg_secret_policy", &self.msg_secret_policy)
             .field("msg_secret_retention", &self.msg_secret_retention)
+            .field(
+                "seed_msg_secrets_from_history",
+                &self.seed_msg_secrets_from_history,
+            )
             .field(
                 "original_message_resolver",
                 &self.original_message_resolver.is_some(),
@@ -295,6 +310,7 @@ impl Default for CacheConfig {
             // longer accumulates a secret for every message forever.
             msg_secret_policy: MsgSecretPolicy::default(),
             msg_secret_retention: MsgSecretRetention::default(),
+            seed_msg_secrets_from_history: true,
             original_message_resolver: None,
             cache_stores: CacheStores::default(),
         }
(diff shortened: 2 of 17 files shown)


jlucaso1:
@claude review

jlucaso1:
@claude review

PR #672, opened 2026-06-01
perf: implement streaming decompression for history sync processing

jlucaso1:
## What

Cuts the memory cost of history-sync decode along two axes: peak (a new streaming decompress-and-parse path) and total allocations (the per-message and per-secret-record work). All in `wacore::history_sync` plus the pooled decompressor in `wacore_binary::zlib_pool`.

1. **Streaming decompression + incremental parse (peak).** A new `InflateReader` decompresses zlib into a bounded 64 KB window and keeps an accumulation buffer that drops its consumed prefix, so a caller can read length-delimited records as they appear. `process_history_sync_streaming` walks the top-level `HistorySync` fields (conversations, pushnames, nctSalt) as their bytes arrive, extracts from each conversation, and discards it before pulling the next. `process_history_sync` now branches: when the caller does not need the full decompressed blob (`retain_blob == false`), it streams; when a consumer wants it (`Event::HistorySync`), the existing whole-blob path runs unchanged. Peak shifts from the entire decompressed blob to roughly the largest single conversation.

2. **Per-message partial decode (allocations).** `extract_conversation_fields` now walks the conversation protobuf directly and prost-decodes each `HistorySyncMsg` one at a time, extracting its secret record and dropping it, instead of decoding the conversation into a `Vec<HistorySyncMsgInternalFields>` that heap-allocated one element per message. The forwarded / poll / bot detection stays in prost via the typed `MessageInternalFields`, so the security-relevant logic is untouched; only the outer wrapper is read by hand.

3. **Message-secret accumulator pre-sizing (allocations).** The records vector is sized once from a single allocation-free count pass (`count_history_sync_messages`) and records are pushed straight into it, with no per-conversation temp `Vec` and no `extend` copy. This removes the one-element-at-a-time growth (`grow_one`) that dominated the decode's allocations.

4. **zlib output buffer sizing.** The pooled decompressor's output buffer is now pre-sized from the compressed length (`2x`, bounded by the hard cap) instead of a fixed 64 KB clamp, which had forced roughly `log2(size / 64 KB)` doubling reallocations for every multi-MB chunk.

## Why

The history-sync core only extracts message secrets, tctokens, pushname and nctSalt during the protobuf walk; it never needs the fully materialized proto, nor every message in memory at once. Holding the whole decompressed blob, plus one heap box per message, plus a doubling accumulator, was pure cost with no decode benefit. Only an external `Event::HistorySync` consumer actually needs the blob, and on that path nothing changes.

## Results

Measured with `dhat` on synthetic history-sync blobs (varied per-message content, so they compress at a realistic ratio rather than collapsing to near-zero).

- **Total allocated (decode path, 500 conversations x 30 messages, secret on every message):** 147 MB to 84 MB (about -43%), and fewer allocations than the baseline (393k to 375k) despite the same decode work. The `grow_one` churn from the message vector and the secret-record accumulator goes to zero; the 60k per-message boxes go to zero.
- **Peak (streaming, 2000 conversations x 30 messages, ~200 B body each, secret on every 10th):** at the global heap maximum, 31.1 MB to 1.9 MB (about -94%, ~16x); total allocated 48.5 MB to 4.2 MB. This is the no-consumer case, where the blob never has to be materialized.

## Compatibility

- No breaking changes. `InflateReader` is new and additive. `process_history_sync`'s signature is unchanged; the streaming path is selected by the existing `retain_blob` argument and returns the same secrets / tctokens / pushname / nctSalt as the full path (covered by a parity test).
- Streaming runs only when there is no `Event::HistorySync` consumer (the caller passes `retain_blob = has_handlers()`). When a consumer is present, the blob is still produced via the full path, so peak there is unchanged. Broadening the gate (per-event-type interest, so a client with only unrelated handlers can still stream) is a possible follow-up.

## Tests

New coverage:

- Full-vs-streaming parity over a multi-conversation blob that includes a >64 KB conversation spanning multiple decompress windows, a group (no tctoken expected), pushname and nctSalt. Both paths return byte-identical secrets, tctokens, pushname and nctSalt; only the full path retains the blob.
- `InflateReader` units: round-trip read back in tiny odd-sized steps across chunk boundaries, an `ensure()` larger than the 64 KB window, the decompressed-size cap, and one-shot vs streaming equivalence.
- The existing extraction tests (top-level and context secrets, forwarded / nested-forwarded skipped) still pass through the new per-message path.

Validation:

- `cargo fmt --all`
- `cargo clippy -p wacore -p wacore-binary --lib -- -D warnings` clean
- `cargo test -p wacore -p wacore-binary` (history-sync + zlib suites green)

### DIFF
--- a/wacore/binary/src/zlib_pool.rs
+++ b/wacore/binary/src/zlib_pool.rs
@@ -9,24 +9,139 @@ thread_local! {
     ));
 }
 
+/// Streaming zlib reader: decompresses `input` incrementally into a small
+/// accumulation buffer, so a caller can parse length-delimited records as they
+/// become available and discard consumed bytes — peak memory stays ~the largest
+/// single record being buffered, not the whole decompressed blob.
+///
+/// Usage: `ensure(n)` to make ≥ n bytes available, read from `available()`, then
+/// `consume(k)`. The buffer is compacted (consumed prefix dropped) as it grows.
+pub struct InflateReader<'a> {
+    input: &'a [u8],
+    in_pos: usize,
+    decomp: Decompress,
+    buf: Vec<u8>,
+    cursor: usize,
+    total_out: u64,
+    max: u64,
+    eof: bool,
+}
+
+impl<'a> InflateReader<'a> {
+    /// Output decompress window per pump; also the compaction threshold.
+    const CHUNK: usize = 64 * 1024;
+
+    pub fn new(input: &'a [u8], max: u64) -> Self {
+        Self {
+            input,
+            in_pos: 0,
+            decomp: Decompress::new(true),
+            buf: Vec::with_capacity(Self::CHUNK),
+            cursor: 0,
+            total_out: 0,
+            max,
+            eof: false,
+        }
+    }
+
+    /// Unparsed decompressed bytes currently buffered.
+    #[inline]
+    pub fn available(&self) -> &[u8] {
+        &self.buf[self.cursor..]
+    }
+
+    /// Mark `n` already-read bytes as consumed.
+    #[inline]
+    pub fn consume(&mut self, n: usize) {
+        self.cursor = (self.cursor + n).min(self.buf.len());
+    }
+
+    /// Ensure at least `need` unparsed bytes are buffered, decompressing more as
+    /// required. Returns `Ok(false)` if the stream ends before reaching `need`.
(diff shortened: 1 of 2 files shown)


jlucaso1:
@claude review

PR #677, opened 2026-06-01
perf(lid-pn): skip PN->LID session migration for peers with no PN state

jlucaso1:
## What

Skip the per-device PN->LID Signal-session migration when the peer has no PN state to migrate.

`migrate_signal_sessions_on_lid_discovery` runs whenever a new PN<->LID mapping is learned. It loops `MIGRATION_DEVICE_RANGE` (100) device slots, and for each takes two per-address session locks plus a session and identity lookup. For a freshly-resolved peer (every member of a large group on the first send) none of those slots have any state, so the entire 100-slot scan is wasted.

- New `SignalStore::has_signal_state_for_user(user)` (default `true`, so a backend that doesn't implement it keeps the full scan). Implemented for the in-memory backend (key prefix scan) and SQLite (`EXISTS` over sessions and identities, filtered by the owning device id).
- New `SignalStoreCache::has_state_for_user(user, backend)` that checks the in-memory cache first (conservatively: any matching key counts) then the durable backend.
- The migration early-returns when neither has any session or identity for the PN user. The 100-slot loop is unchanged for peers that do have state.

## Why

Profiling a cold group send to an ~800-member group (new `group-send` bench scenario) showed `migrate_signal_sessions_on_lid_discovery` as 95+ percent of all allocation: 800 members x 100 slots = 80k lock+lookup iterations, all finding nothing because the members were freshly resolved. This is invisible in latency logs but is real malloc/CPU churn (and likely part of the report's "~4s session setup").

## Results

dhat on the cold first group send to 800 fresh members:

- `migrate_signal_sessions_on_lid_discovery`: 10.04 GB to 0 (-100 percent), ~5.15M allocations to ~800.
- Total run: 10.07 GB to 0.02 GB (about -99.8 percent), allocations -97.3 percent.
- 12/12 group replies delivered, 0 failed (behavior unchanged).

## Correctness / protocol

The guard only skips when there is provably nothing to migrate, which is identical to what the 100-slot loop did when every slot was empty. For peers with existing PN sessions (the migration's actual purpose, e.g. a prior DM contact who later reveals a LID) the full migration runs unchanged. This matches whatsmeow, which migrates only the sessions that exist for the user rather than a fixed device range; the guard is a conservative subset of that (skip when none).

## Tests

- `has_signal_state_for_user_matches_by_user_prefix` (in-memory): false when empty, true after a device-0 session, no false-positive for a longer user that this one prefixes, identity-only also counts.
- `migrate_skips_when_no_pn_signal_state`: a stateless peer migration creates no LID session.
- `migration_blocks_on_per_address_session_lock` now seeds a PN session so it still exercises the lock the loop takes (the guard would otherwise skip a stateless migration). Existing `migration_preserves_working_session_when_both_namespaces_present` and `migration_lock_dance_*` still pass.
- `cargo clippy --all-targets -- -D warnings` clean; `cargo test -p wacore -p whatsapp-rust -p whatsapp-rust-sqlite-storage` (829 + 658 + 31 passing).

### DIFF
--- a/src/client/lid_pn.rs
+++ b/src/client/lid_pn.rs
@@ -404,6 +404,18 @@ impl Client {
 
         let backend = self.persistence_manager.backend();
 
+        // Nothing to migrate unless the PN side has Signal state. For a freshly
+        // resolved peer (e.g. every member of a large group on first send) this
+        // skips MIGRATION_DEVICE_RANGE lock+lookup iterations that would all
+        // find nothing. On a lookup error, fall through to the full scan.
+        if let Ok(false) = self
+            .signal_cache
+            .has_state_for_user(pn, backend.as_ref())
+            .await
+        {
+            return;
+        }
+
         for device_id in 0..MIGRATION_DEVICE_RANGE {
             // `&str` → `CompactString` is inline for ≤24-byte user parts
             // (all PN/LID identifiers fit), so no String intermediate.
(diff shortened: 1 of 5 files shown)


jlucaso1:
@claude review

PR #680, opened 2026-06-01
perf(usync): batch LID-PN re-learn from device response off the cold group-send path

jlucaso1:
## Context
Continued group-send benchmark profiling against the current main (post #677/#678/#679). The big bottleneck — `migrate_signal_sessions_on_lid_discovery` (10 GB on an 800-member cold send) — is already fixed (#677): the cold send now allocates only ~16.6 MB. Profiling the remainder surfaced one clean, latency-only redundancy.

## Finding
`process_device_list_response` (the usync device-list handler inside `get_user_devices`) learned each LID↔PN mapping by **awaiting `add_lid_pn_mapping` per entry** — a serial `record + DB write + migration` per mapping. On a large **cold group send**, the device usync echoes the same mappings `query_info` already learned moments earlier (via its *batched, detached* `learn_lid_pn_mappings_batch`), so the send blocked on ~N serial awaited DB writes for pairs we already had.

CPU was never the issue (0.15 s total for 40 sends to 800 members); this was wall-clock latency from serial awaited `spawn_blocking` SQLite hops on the send's critical path.

## Change
Route the usync re-learn through `learn_lid_pn_mappings_batch` (the same path `query_info` uses): one detached transaction for genuinely-new pairs, internally guarded against re-learning already-durable ones. Falls back to the per-mapping path if the owning `Arc<Client>` isn't available. No new ordering risk — `query_info` already triggers these migrations detached before `get_user_devices` runs.

## Measurement (group-send bench, 800 members, mock server, single A/B)
| metric | baseline | this PR | Δ |
|---|---|---|---|
| ack_avg | 64.50 ms | 55.48 ms | **−14%** |
| ack_max (cold first send) | 139.33 ms | 95.68 ms | **−31%** |
| pong_avg | 80.56 ms | 73.39 ms | −9% |
| duration | 683 ms | 671 ms | −2% |
| CPU total | 0.150 s | 0.150 s | = |
| dhat total alloc | 16.61 MB | 16.70 MB | = (alloc unchanged; pure latency win) |

Multiple independent metrics moved together, consistent with removing the serial awaited writes from the cold send.

## Scope note
The remaining client-side cost for large groups (per-member device/LID resolution on every send: `get_current_lid`/`resolve_lookup_keys` String churn) is O(members × messages) but CPU-cheap (0.15 s total), so it isn't pursued here. The dominant per-send latency (pong ~73 ms) is server/network fanout + the per-(group,sender) chain-lock serializing the bench flood, not client work.

Tests: clippy `--all-targets -D warnings` clean; whatsapp-rust lib suite (659) green; existing `process_response_preserves_omitted_users` (merge-safety) unaffected.

### DIFF
--- a/src/usync.rs
+++ b/src/usync.rs
@@ -50,13 +50,18 @@ impl Client {
     /// simply absent here, so their cached records are left untouched (the
     /// merge-safe behavior the `device_hash` optimization depends on).
     async fn process_device_list_response(&self, response: &DeviceListResponse) -> Vec<Jid> {
-        // Learn LID↔PN mappings from the response via the batched, guarded path:
-        // one detached DB transaction for genuinely-new pairs (skipping ones
-        // already durably learned, e.g. by query_info's batch learn right before
-        // a group send). The previous per-mapping add_lid_pn_mapping awaited a DB
-        // write + migrations serially, so a large cold group send redid ~N writes
-        // on the critical path for mappings we already had. Falls back to the
-        // per-mapping path if the owning Arc<Client> isn't available.
+        // Learn LID↔PN mappings via the same batched, guarded learner query_info
+        // uses (one detached transaction, skipping already-durable pairs), so
+        // per-mapping DB writes stay off the send's critical path. Falls back to
+        // the per-mapping path if the owning Arc<Client> isn't available.
+        //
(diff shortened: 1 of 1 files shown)


jlucaso1:
@claude review

PR #685, opened 2026-06-02
feat(upload): add encrypted_len() and UploadSource for Bytes

jlucaso1:
## What

Two small, additive helpers for the constant-memory streaming upload path
(`Client::upload_stream` / `encrypt_media_streaming`, from #684):

- `encrypted_len(plaintext_len) -> usize`: the exact size of the encrypted blob
  (PKCS#7-padded up to the next 16-byte block, plus the 10-byte MAC). Callers can
  pre-size the encrypt destination buffer exactly, so a `Vec` neither grows during
  the encryptor's per-block writes nor shrink-reallocates when handed to an
  `UploadSource`.

- `impl UploadSource for bytes::Bytes`: O(1) to build from an owned `Vec`
  (`Bytes::from` adopts the buffer), avoiding the full-ciphertext copy that
  `Arc::<[u8]>::from(vec.into_boxed_slice())` makes (the only in-memory source
  shipped before).

## Why

Profiling an image-upload workload showed the ciphertext buffer (caller-side) was
the single biggest allocator in the upload path: `Vec::new()` grows by doubling
(~14 reallocs per upload) and `Arc::<[u8]>::from` copies the whole blob. With
these helpers a caller does:

```rust
let mut ct = Vec::with_capacity(encrypted_len(plaintext.len()));
encrypt_media_streaming(reader, &mut ct, MediaType::Image)?;
client.upload_stream(Bytes::from(ct), info, MediaType::Image).await?;
```

Measured on a local media-upload benchmark (50 uploads of a ~59 KB image):

| variant | total alloc churn |
| --- | --- |
| baseline (`Vec::new` + `Arc::from`) | 26.70 MB |
| pre-size exact + Bytes (these helpers) | 17.61 MB (-34%) |
| full disk streaming (Bytes / file source) | 14.94 MB (-44%) |

Peak heap and latency are unchanged; the win is pure allocation churn, which
scales with file size (a large video avoids a full-blob copy per attempt).

## Notes

- Fully additive, no breaking changes (the trait gets a new impl, `encrypted_len`
  is a new `const fn`).
- Platform-agnostic: `std::io` + `bytes` only, no filesystem, so it keeps building
  on wasm32 and embedded targets. Filesystem / remote (S3, etc.) sources stay
  caller-provided through the `UploadSource` trait.
- Tests: `encrypted_len` is pinned to the real encryptor output across block
  boundaries (including the aligned case); the `Bytes` source is checked for
  offset reads and byte-for-byte parity with the `Arc<[u8]>` impl, plus a doctest
  on `encrypted_len`.

### DIFF
--- a/wacore/src/upload.rs
+++ b/wacore/src/upload.rs
@@ -8,6 +8,8 @@ use rand::rng;
 use std::io::{Read, Write};
 
 const BLOCK: usize = 16;
+/// Length of the truncated HMAC-SHA256 appended to the ciphertext.
+const MEDIA_MAC_LEN: usize = 10;
 
 /// Streaming sidecar chunk size: one HMAC per 64 KiB of ciphertext.
 const SIDECAR_CHUNK: usize = 64 * 1024;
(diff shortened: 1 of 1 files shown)


jlucaso1:
@claude review

PR #698, opened 2026-06-03
fix(message): emit quote remoteJid only for cross-chat quotes

jlucaso1:
## Problem

`build_quote_context_with_info` always set `ContextInfo.remoteJid` to the quoted message's chat JID. That is not what WhatsApp Web does: the message model's `msgContextInfo` getter emits `remoteJid` only when the quoted message lives in a chat other than the send target (`quotedMsg.remote != targetChat`), and a same-chat reply omits it. The serializer (`WAWebE2EProtoGenerator`) only encodes whatever `quotedRemoteJid` it receives, which is null for a same-chat quote, so the field is dropped on the wire.

Always emitting `remoteJid` on a same-chat reply diverges from WA Web and can make some clients treat a 1:1 reply as a cross-chat reference. Baileys mirrors WA Web here too (`contextInfo.remoteJid` set only when `jid !== quoted.key.remoteJid`); whatsmeow never populates `ContextInfo.RemoteJID` at all. The receive side confirms the omission is lossless: when `remoteJid` is absent, WA Web reconstructs the quoted chat from the enclosing message's `remote` (the current chat).

The old comment claimed `remoteJid` was "required by iOS to scope the quote". That was a misreading of the serializer layer. The actual defense against re-notifying mentions carried inside a quoted message is `prepare_for_quote` (it clears `mentioned_jid`/`group_mentions` from the nested copy), which is independent of `remote_jid` and predates this code.

## Fix

Thread the send target into `build_quote_context_with_info` (`quoted_chat_jid` plus a new `target_chat_jid`) and emit `remote_jid` only for a cross-chat quote:

```rust
let remote_jid = (!quoted_chat_jid.is_same_chat_as(target_chat_jid))
    .then(|| quoted_chat_jid.with_device(0).to_string());
```

The emitted `remote_jid` is the device-less chat form: a chat reference never carries a device, so `with_device(0)` strips it (and keeps the agent that `@bot`/`@interop` JIDs render, which `to_non_ad` would wrongly drop), matching the device-insensitive compare.

`Jid::is_same_chat_as` is a new device-insensitive comparator. It compares `user`/`server`/`integrator`, ignores the multi-device `device`, and compares `agent` **only for servers that render it** (`@bot`/`@interop`) via a new `Server::renders_agent` that is the single source of truth shared with the JID formatter. That keeps the comparison consistent with the JID string we actually emit: Pn/Lid/Hosted suppress the agent in display, so a stray decoded agent byte there must not split the chat, while `@bot`/`@interop` render it, so distinct agents are distinct chats. It allocates nothing, where `to_non_ad() == to_non_ad()` would build two throwaway `Jid`s.

This is equivalent to WA Web's chat comparison for every realistic input (chat JIDs are conversation addresses that carry no device), though it is not a literal port of `Wid.equals` (which is device-sensitive and has no agent concept). Newsletter participant resolution stays decoupled from `remote_jid`: a same-chat newsletter quote still sets `participant = channel`, and a cross-chat one sets both.

The bot reply path (`MessageContext::build_quote_context`) is structurally same-chat (it quotes and sends into `info.source.chat`), so it now omits `remote_jid`, matching WA Web. Cross-chat quotes stay reachable via the low-level `Client::send_message`.

## Tests

`wacore/src/proto_helpers.rs`:

- renamed `quote_context_sets_remote_jid_for_{group,dm}` to `_omits_remote_jid_same_chat_{group,dm}` (now assert `None`)
- added `_emits_remote_jid_cross_chat` (group A into group B), `_status_reply_is_cross_chat` (status@broadcast into a DM), `_device_suffix_treated_as_same_chat`, `_cross_chat_remote_jid_drops_device_suffix` (a device-scoped quoted chat emits a device-less remote_jid)
- newsletter: `_uses_channel_as_participant` (same-chat: participant kept, `remote_jid` None) and `_cross_chat_sets_both` (participant channel AND `remote_jid` emitted)
- `quote_context_strips_mentions_from_quoted_message` unchanged (the mention defense is untouched)

`wacore/binary/src/jid.rs`:

- `is_same_chat_as_matches_rendered_chat_identity`: device ignored; user/server/integrator distinguish chats; a suppressed Pn agent byte does not split the chat; distinct `@bot` agents do

```
cargo fmt --all
cargo clippy --all-targets -- -D warnings   # clean
cargo test -p wacore --lib proto_helpers     # pass (incl. new/renamed quote tests)
cargo test -p wacore-binary --lib            # 92 pass (incl. is_same_chat_as + Display)
```

## Breaking

`build_quote_context_with_info` gains a `target_chat_jid` parameter. The only in-tree caller (`MessageContext::build_quote_context`) is updated. External callers must pass the send target; pass the same JID twice for a same-chat reply.

### DIFF
--- a/src/bot.rs
+++ b/src/bot.rs
@@ -68,10 +68,14 @@ impl MessageContext {
     }
 
     pub fn build_quote_context(&self) -> wa::ContextInfo {
+        // A bot reply is same-chat: quoted chat and send target are both
+        // info.source.chat, so remote_jid is omitted (WA Web parity).
+        let chat = &self.info.source.chat;
         wacore::proto_helpers::build_quote_context_with_info(
             &self.info.id,
             &self.info.source.sender,
-            &self.info.source.chat,
+            chat,
+            chat,
             &self.message,
         )
     }
--- a/wacore/binary/src/jid.rs
+++ b/wacore/binary/src/jid.rs
@@ -224,6 +224,15 @@ impl Server {
     pub fn is_lid_family(self) -> bool {
         matches!(self, Self::Lid | Self::HostedLid)
     }
+
+    /// Whether the `agent` byte is part of the rendered JID for this server.
+    /// AD-capable servers (Pn/Lid/Hosted/HostedLid) carry it as a hidden domain
+    /// byte and suppress it in display; others (e.g. `@bot`/`@interop`) render it.
+    /// Single source of truth shared by the formatter and `Jid::is_same_chat_as`.
+    #[inline]
+    pub fn renders_agent(self) -> bool {
+        !matches!(self, Self::Pn | Self::Lid | Self::Hosted | Self::HostedLid)
+    }
 }
 
 impl fmt::Display for Server {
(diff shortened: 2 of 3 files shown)


jlucaso1:
@claude review, also please check if is possible to evict some unecessary clone or allocation

PR #714, opened 2026-06-04
perf(store): batch session/identity/sender-key flush into one transaction per category

jlucaso1:
## Problem

`SignalStoreCache::flush()` ran at the end of every send and wrote each dirty session, identity, and sender key one at a time via `put_session` / `put_identity` / `put_sender_key`. On the SQLite backend each of those is its own `spawn_blocking` + `db_semaphore` acquire (a single-permit write lane) + a one-row insert in its own implicit transaction.

A cold group send that establishes hundreds-to-thousands of fresh Signal sessions turns a single flush into that many `spawn_blocking` dispatches and that many separate transactions, serialized on the one write lane. The codebase already proves the batched shape is correct and far cheaper (`update_device_lists`, `store_prekeys_batch`), but no `put_*_batch` existed for sessions/identities/sender keys.

## Change

Add `put_sessions_batch`, `put_identities_batch`, and `put_sender_keys_batch` to the `SignalStore` trait, each with a default implementation that loops the singular calls (so in-memory, wasm, and other backends keep working unchanged). The SQLite backend overrides them following the existing `store_prekeys_batch` template: one `with_retry`, one `conn.transaction` looping the upserts, one write-lane acquire.

`flush()` now collects the dirty puts per category and emits one batched call each, collapsing a cold group send's flush from N `spawn_blocking` + N transactions down to about three (one per category). Deletes stay singular (not the hotspot).

Addresses are passed as `Arc<str>` (the cache's own key type) rather than `String`, so building the batch is a refcount bump per entry instead of a string allocation, and the SQLite impl uses the shared key directly as `&str` in the query. The now-unused `flush_encode_buf` was removed; each entry serializes into its own `Bytes` (moved, not copied).

## Benchmark

Wall-clock, in-memory SQLite, best-of-3, writing 500 sessions:

| Path | Time |
|---|---|
| 500x `put_session` (per-entry) | 6.34 ms |
| 1x `put_sessions_batch` | 1.83 ms |

~3.5x for the per-flush transaction reduction. This is single-threaded in-memory, so it captures the `spawn_blocking` + per-statement transaction overhead but not the write-lane contention under concurrency or on-disk WAL, where the win is larger.

## Tests

- `put_signal_batches_persist_and_upsert` (new): all three batches persist and read back, re-batching the same addresses upserts (`on_conflict do_update`), and empty batches short-circuit.
- Full lib suites green; `cargo clippy --all-targets -- -D warnings` clean.

### DIFF
--- a/storages/sqlite-storage/src/sqlite_store.rs
+++ b/storages/sqlite-storage/src/sqlite_store.rs
@@ -263,6 +263,15 @@ impl SqliteStore {
                     if is_retriable_sqlite_error(e) && attempt < MAX_RETRIES =>
                 {
                     let delay_ms = 10u64 * (1u64 << attempt.min(4));
+                    // Skip the first transient blip; warn from the second retry on so
+                    // sustained busy/locked contention doesn't go unobserved.
+                    if attempt >= 1 {
+                        warn!(
+                            "{op_name} busy/locked, retry {}/{} in {delay_ms}ms: {e}",
+                            attempt + 1,
+                            MAX_RETRIES + 1
+                        );
+                    }
                     tokio::time::sleep(tokio::time::Duration::from_millis(delay_ms)).await;
                 }
                 Ok(Err(e)) => return Err(e.into()),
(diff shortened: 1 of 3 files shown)


jlucaso1:
@coderabbitai review

jlucaso1:
@claude review

PR #716, opened 2026-06-04
refactor(appstate): drive syncd actions from a generated schema registry + public generic action API

jlucaso1:
## Problem

App-state (syncd) actions were hand-coded one by one: each `ChatActions` / `Labels` method built its own index, picked a collection, and stamped a fixed `SyncActionData.version = 1`. But every action has its own version (whatsmeow: `mute=2, pin=5, archive=3, markChatAsRead=3, star=2, label=3, setting_pushName=1`), so version 1 was wrong for almost all of them and could be mishandled by linked clients that honor the action version. Each new action also meant re-deriving the collection and index shape by hand.

## Change

Adopt a generated, dependency-free registry of every syncd action (`wacore/appstate/src/schemas.rs`, generated from WhatsApp Web's app-state schemas). Each `Schema` carries its `collection`, `version`, `scope`, value field, and `index_parts` (the index shape), plus `ALL` and `by_name`.

A new helper drives sends from the registry:

```rust
impl Client {
    pub async fn send_app_state_action(
        &self,
        schema: &schemas::Schema,
        index_args: &[&str],
        value: &wa::SyncActionValue,
    ) -> Result<()>;
}
```

It builds the index from `schema.index_parts` + the supplied non-literal args, maps `schema.collection` to `WAPatchName`, and stamps `schema.version`. `ChatActions` (mute/pin/archive/markChatAsRead/deleteChat/deleteMessageForMe/star), `Labels`, and the push-name setting now all go through it, so every action gets its correct per-action version for free. `encode_record` already takes the version (added in #715); this wires the right value into every call site.

## Public generic API

`send_app_state_action` is public and `schemas` is re-exported, so callers can send any of the registry's actions even when there is no dedicated helper, the same way `send_message` takes an auto-generated `Message`:

```rust
use whatsapp_rust::schemas;
use whatsapp_rust::waproto::whatsapp as wa;

let value = wa::SyncActionValue {
    clear_chat_action: Some(Default::default()),
    timestamp: Some(now_ms),
    ..Default::default()
};
client.send_app_state_action(&schemas::CLEAR_CHAT, &["123@s.whatsapp.net"], &value).await?;
```

The typed APIs (`ChatActions`, `Labels`) are thin wrappers over it.

## Allocations / type-safety

- `build_action_index` assembles the index from `&str` references (no string copies); the only allocation is the JSON output.
- The message-key path (`star` / `deleteMessageForMe`) now owns only what must outlive the call (chat + participant JIDs); `messageId` is borrowed and `fromMe` is a `&'static "1"/"0"` (previously a 4-element `String` array).
- Label create/delete allocate nothing for the index (the label id is already a `&str`); single-JID actions allocate only the one unavoidable JID string.
- Index arg counts are validated against the schema at runtime.

## Versions fixed

mute 1->2, pin 1->5, archive 1->3, markChatAsRead 1->3, deleteChat 1->6, deleteMessageForMe 1->3, star 1->2. label_edit/label_jid stay 3, setting_pushName stays 1 (both already correct).

## Tests

- `build_index_matches_legacy_shapes`: the registry-driven index is byte-for-byte identical to the previous hand-coded format for every migrated action, so the migration changes only the version on the wire, not the index.
- `registry_versions_match_whatsmeow`: locks the per-action versions; a regenerated registry that changes one trips this for review.
- `collection_mapping`, arg-count mismatch, and a doctest for the public generic API.

Full lib suites green; `cargo clippy --all-targets -- -D warnings` clean.

### DIFF
--- a/src/features/chat_actions.rs
+++ b/src/features/chat_actions.rs
@@ -9,6 +9,7 @@ use crate::client::Client;
 use anyhow::Result;
 use log::debug;
 use wacore::appstate::patch_decode::WAPatchName;
+use wacore::appstate::schemas::{self, IndexPart, Schema};
 use wacore::types::events::{
     ArchiveUpdate, ContactUpdate, DeleteChatUpdate, DeleteMessageForMeUpdate, Event,
     MarkChatAsReadUpdate, MuteUpdate, PinUpdate, StarUpdate,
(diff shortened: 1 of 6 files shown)


jlucaso1:
@claude review

PR #717, opened 2026-06-04
feat(signal): react to locally-detected peer identity changes

jlucaso1:
## Problem

When a peer reinstalls WhatsApp (or otherwise rotates their identity key), our Signal store already detects it: `save_identity` returns `IdentityChange::ReplacedExisting` when the stored identity key differs from the new one. `is_trusted_identity` intentionally returns `true` (TOFU, matching WA Web's `ProtocolStoreUnifiedApi.isTrustedIdentity`), so an identity change surfaces through `save_identity` rather than by rejecting the message.

Every caller discarded that signal. `process_prekey_bundle` and the `message_decrypt*` family called `save_identity` and threw the `IdentityChange` away, so a locally-detected identity change fired nothing: no event, no device-record cleanup, no tcToken refresh. Only the server's `<notification type="encrypt"><identity/></notification>` push (`WAWebHandleIdentityChange`) produced any reaction. A peer re-registration detected locally (a pkmsg/bundle arriving before the server push, or a bundle served with a divergent identity) was silently accepted.

Separately, our server-push handler (`handle_identity_change`) never re-issued the trusted-contact token that WA Web's `handleE2eIdentityChange` sends, so the tcToken refresh on identity change was effectively missing.

WA Web reacts in both places: `ProtocolStoreUnifiedApi.saveIdentity` calls `handleNewIdentity` when the stored key differs, and the server-push handler does the heavier `handleE2eIdentityChange`.

## Change

Thread the `IdentityChange` up from the Signal layer and react to `ReplacedExisting` on the local detection paths, mirroring WA Web's `saveIdentity -> handleNewIdentity`.

Signal layer (wacore):
- `message_decrypt` / `message_decrypt_prekey` / `message_decrypt_signal` now return `DecryptionResult { plaintext, identity_change }`.
- `process_prekey_bundle` / `process_prekey_bundle_inner` return `Result<IdentityChange>`.
- `process_prekey` additionally signals whether it reused an already-established session.
- Decrypting against a non-fresh session reports `NewOrUnchanged`, because re-saving an older identity for out-of-order delivery is not the peer's current identity changing. This covers both the previous (archived) session in the plain SignalMessage path and a duplicate/out-of-order `PreKeySignalMessage` that matched an existing session.

High-level client:
- New `handle_local_identity_change`. Gated to the peer's primary device (`device == 0`) and skip-self, it runs the lighter subset of WA Web's `handleNewIdentity`: `clear_device_record` (companion devices) + `invalidate_device_cache` + re-issue an active tcToken + dispatch `Event::IdentityChange { implicit: true }`. It deliberately does NOT delete the primary session, rotate the status sender key, or call `ensure_e2e_sessions`. Those belong to the server-push handler, and the new session is being established by the very message that triggered detection. WA Web's local `handleNewIdentity` omits them too.
- The server-push handler `handle_identity_change` now also re-issues an active tcToken, matching WA Web `handleE2eIdentityChange -> sendTcTokenWhenDeviceIdentityChange`. This is the path that fires in practice, so the tcToken refresh now actually happens. It is spawned and no-ops unless a non-expired sender token already exists (so it only fires for a prior relationship, mirroring WA Web's "had old identity" guard).
- The reaction is dispatched off-path via `Client::react_to_local_identity_change` (spawned, detached) so it never runs under the per-sender session lock, and `clear_device_record` only touches companion (non-primary) sessions.
- Triggers: the inbound decrypt batch (`process_session_enc_batch`, deduped once per batch and mirrored in the retry-after-`UntrustedIdentity` branch), outbound session establishment (`ensure_e2e_sessions`, retry-receipt bundle), and the send path (`encrypt_for_devices`) via a new defaulted `SendContextResolver::on_local_identity_change` callback. PN to LID migration decrypt does not react (re-addressing, not a reinstall).
- `Event::IdentityChange` gains an `implicit` field: `false` for the server `<identity/>` push, `true` for local detection.

No loop or self-clobber: the server-push handler deletes and flushes the identity before spawning `ensure_e2e_sessions`, so the subsequent bundle reports `NewOrUnchanged`.

## Tests

- `cargo fmt --all`
- `cargo clippy --all-targets -- -D warnings` (clean)
- `cargo test --all --exclude e2e-tests` (all green)

New:
- libsignal threading: `process_prekey_bundle_reports_identity_change`, `decrypt_prekey_reports_identity_change` (NewOrUnchanged on first contact, ReplacedExisting on a replaced key, plaintext intact).
- reuse suppression: `process_prekey_signals_reuse_for_established_session` (first pkmsg establishes, re-processing the same pkmsg signals reuse).
- gating: `test_local_identity_change_dispatches_implicit_event`, `test_local_identity_change_skips_self`, `test_local_identity_change_skips_companion_device`.
- send path end to end: `encrypt_for_devices_reports_replaced_identity` (a bundle carrying a new identity for a known address fires the resolver callback).

## Performance

No measurable impact, so no benchmark. The decrypt return now carries a 1-byte `Copy` enum (no allocation; `save_identity` already computed it). On the send path the establishment task returns `Option<Jid>`: `None` on the common path (no clone), and only moves the already-owned `Jid` on the rare `ReplacedExisting`. The reaction (and the tcToken IQ, which is conditional on an active token) only runs on an actual identity change.

## Breaking

Yes (pre-1.0):
- `message_decrypt`, `message_decrypt_prekey`, `message_decrypt_signal` return `DecryptionResult` instead of `Vec<u8>`.
- `process_prekey_bundle` returns `Result<IdentityChange>` instead of `Result<()>`.
- `process_prekey` returns `Result<(PreKeysUsed, IdentityToSave, bool)>` (the trailing bool signals session reuse).
- `Event::IdentityChange` gained a non-optional `implicit: bool` field.
- `SendContextResolver` gained a defaulted `on_local_identity_change` method (no-op default, so external implementors are unaffected).

### DIFF
--- a/src/client/context_impl.rs
+++ b/src/client/context_impl.rs
@@ -49,4 +49,8 @@ impl SendContextResolver for Client {
     async fn get_lid_for_phone(&self, phone_user: &str) -> Option<wacore_binary::CompactString> {
         self.lid_pn_cache.get_current_lid(phone_user).await
     }
+
+    fn on_local_identity_change(&self, jid: &Jid) {
+        self.react_to_local_identity_change(jid);
+    }
 }
--- a/src/client/device_registry.rs
+++ b/src/client/device_registry.rs
@@ -415,6 +415,11 @@ impl Client {
     /// Matches WA Web's `clearDeviceRecord()` in `IdentityUpdateDeviceTableApi`:
     /// - Deletes Signal sessions for non-primary devices (stale identity)
     /// - Invalidates sender key device cache so SKDM will be redistributed
+    ///
+    /// The companion-device session wipe is intentionally not per-device locked
+    /// (matches WA Web's single-threaded model). A concurrent encrypt to one of
+    /// those companions can re-store a session right after the wipe, but that is
+    /// self-healing: the next send re-establishes it via `process_prekey_bundle`.
     pub(crate) async fn clear_device_record(
         &self,
         user: &str,
--- a/src/client/device_registry.rs
+++ b/src/client/device_registry.rs
@@ -256,6 +256,29 @@ impl Client {
         Ok(())
     }
 
+    /// Spawn the local identity-change reaction off the current path so it runs
+    /// after any held session lock is released (the reaction acquires its own
+    /// locks and must not deadlock against an in-flight decrypt/encrypt batch).
+    ///
+    /// Triggered from both the inbound decrypt path and the outbound
+    /// session-establishment paths when `save_identity` reports
+    /// [`IdentityChange::ReplacedExisting`](wacore::libsignal::protocol::IdentityChange),
+    /// mirroring WA Web `saveIdentity` -> `handleNewIdentity`. Gating
+    /// (primary-device, skip-self) lives in [`handle_local_identity_change`].
+    ///
+    /// [`handle_local_identity_change`]: crate::handlers::notification::handle_local_identity_change
+    pub(crate) fn react_to_local_identity_change(&self, sender: &Jid) {
+        let Some(client) = self.self_weak.get().and_then(|w| w.upgrade()) else {
+            return;
+        };
+        let sender = sender.clone();
+        self.runtime
+            .spawn(Box::pin(async move {
+                crate::handlers::notification::handle_local_identity_change(&client, sender).await;
+            }))
+            .detach();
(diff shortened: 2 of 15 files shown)


jlucaso1:
@claude review

PR #718, opened 2026-06-04
perf(identity): gate identity-change reset behind had-prior-identity, like WA Web

jlucaso1:
## Problem

`handle_identity_change` (the handler for the server's `<notification type="encrypt"><identity/></notification>` push, fired when a peer reinstalls / rotates their identity key) ran the full heavy reset unconditionally for every non-self primary-device push: clear device record, delete the primary session + identity under a lock, rotate the status sender key, flush, invalidate the device cache, re-issue a tcToken, dispatch an event, and eagerly re-establish the session.

The dominant cost is self-inflicted: the handler deletes the primary session and then spawns `ensure_e2e_sessions`, which sees the just-deleted session missing and therefore always does a `fetch_pre_keys` IQ + X3DH to rebuild it. So every push paid a network round-trip + a key agreement, including for peers we have no prior relationship with (for example a group-only member we have never had a session with), where the rebuild creates a session we may never use.

WA Web (`WAWebHandleIdentityChange.handleE2eIdentityChange`) gates everything after `clearDeviceRecordForIdentityChange` behind `loadIdentityKey(addr) != null` (only react to a real prior relationship), and only re-establishes when not still draining the offline backlog (`C = !isEmpty(offline) && !isResumeFromRestartComplete()` -> `C || ensureE2ESessions`). We had neither gate.

## Change

Mirror WA Web. Keep the eager re-establishment for the cases that benefit from it (so the next send stays fast), but gate it.

- The device-list cleanup (`clear_device_record` for companion sessions + `invalidate_device_cache`) still runs always, matching `clearDeviceRecordForIdentityChange`, which WA Web runs before the gate.
- Read the stored identity non-destructively (`get_identity`) before deleting it, using the same resolved address the delete uses. With no prior identity, skip the whole reset (session/identity delete, status sender-key rotation, tcToken reissue, the change event, and the eager re-establishment) and return. A backend read error is treated as had-prior (fail-safe: run the reset rather than silently skip it).
- For a prior identity, do the reset as before, then re-establish eagerly unless the push arrived while the offline backlog is still draining (`offline` attr present and `offline_sync_completed == false`), matching WA Web's `C || ensureE2ESessions`. After resume completes, offline-tagged pushes re-establish eagerly again.
- Log `had_prior_identity` per push so the skip rate is measurable.

The deferral is safe because every send path re-establishes before encrypting: `ensure_e2e_sessions` in the DM (`send.rs`) and group (`features/signal.rs`) paths, plus `encrypt_for_devices`'s own `has_session` -> `process_prekey_bundle` fallback. A reinstalled peer's first inbound message is a pkmsg, which establishes the session on decrypt.

## Behavior change

`Event::IdentityChange` (the analog of WA Web's `addSecurityCodeChangedNotifications`, which is inside the gate) now fires only for contacts we had a prior identity with. Consumers no longer get it for a no-prior peer (for example a group-only member), where there was no prior security code to "change."

## Tests

- `cargo fmt --all`
- `cargo clippy --all-targets -- -D warnings` (clean)
- `cargo test --all --exclude e2e-tests` (all green)

Existing identity-change tests were updated to seed a stored identity so they exercise the prior-identity path. New `test_identity_change_no_prior_identity_skips_reset`: a push for a peer with no stored identity dispatches no event, still invalidates the device cache, still deletes a seeded companion-device session (the always-on cleanup), and establishes no identity.

## Performance

Removes the per-push `fetch_pre_keys` IQ + X3DH for the no-prior subset (and the eager re-establish burst after a reconnect). For the prior-identity online case the eager re-establish is kept, so next-send latency is unchanged. The gate adds one non-destructive identity read (already cached on the hot path); the redundant `has_session` read that earlier only fed a log line was removed.

## Breaking

None (signatures unchanged). The `Event::IdentityChange` gating above is a behavior change, not an API break.

### DIFF
--- a/src/handlers/notification.rs
+++ b/src/handlers/notification.rs
@@ -349,19 +349,28 @@ async fn handle_identity_change(client: &Arc<Client>, node: &NodeRef<'_>) {
     // path would otherwise eagerly fetch prekeys + X3DH to build a session we may
     // never use.
     //
-    // Check BOTH the resolved (preferred LID-or-PN) address and the original PN
-    // address. They diverge when a PN->LID mapping was learned from offline replay
-    // but the Signal-state migration hasn't run yet: the identity is still under the
-    // PN address while resolve_encryption_jid already points at the LID one. Reading
-    // only the resolved address would then false-negative and skip a real reset.
+    // Check every address the identity could be stored under, because PN/LID
+    // resolution can diverge from where the state actually lives:
+    //   - the resolved (preferred LID-or-PN) address from resolve_encryption_jid;
+    //   - the original PN address (state can still be under PN when a PN->LID
+    //     mapping was learned from offline replay but the migration hasn't run yet);
+    //   - the LID carried by the stanza itself (the local cache may be cold/evicted
+    //     so resolve falls back to PN, yet the state lives under the stanza LID).
+    // Reading only the resolved address would false-negative and skip a real reset.
     let resolved = client.resolve_encryption_jid(&from_jid).await;
-    let addr = resolved.to_protocol_address();
-    let original_addr = from_jid.to_protocol_address();
-    let mut reset_addrs = vec![addr.clone()];
-    if original_addr != addr {
-        reset_addrs.push(original_addr);
-    }
+    let stanza_lid = node.attrs().optional_jid("lid");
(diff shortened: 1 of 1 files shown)


jlucaso1:
@claude review

PR #724, opened 2026-06-05
fix: @call JID decode, receipt participant_pn, profile-pic empty→remove, has_device churn

jlucaso1:
Four small, independent fixes validated against `docs/captured-js/`, one per commit, each with a test.

## `fix(binary): decode @call JIDs` (commit 1)

`read_jid_pair` returned `BinaryError::AttrParse` for any server string not in the `Server` enum, which fails the entire node decode. `@call` is a real WhatsApp Web server JID (call signaling; `WAWebWid` special-cases it and `isGroupCall` checks `server === "call"`), so a frame carrying one was dropped. Adds a `Server::Call` variant; it is not an AD server, so it round-trips through JID_PAIR like the other non-AD servers, matching WA Web's permissive JID decoding.

## `fix(receipt): thread participant_pn into sender_alt` (commit 2)

The receipt parser already extracts `participant_pn` (the LID->PN counterpart of the reader), but both `handle_receipt` paths dropped it. The inbound message path already maps `participant_pn` to `MessageSource.sender_alt` to warm the LID-PN cache; receipts now do the same on both the aggregated and simple shapes, matching `WAWebHandleMsgReceiptParser` which reads `participantPn` in every branch. No event-schema change.

## `fix(profile): route empty profile-picture bytes to removal` (commit 3)

`Profile::set_profile_picture` forwarded the caller's bytes straight into `SetProfilePictureSpec::set_own`, which asserts (panics) on empty input. `WAWebSendProfilePictureJob` treats empty bytes as a removal (`a ? wap("picture",..) : null`), so empty now routes to the remove path instead of panicking, via a new `SetProfilePictureSpec::for_own` (empty -> `remove_own`, else `set_own`).

## `perf(device-registry): avoid Vec<String> churn in has_device` (commit 4)

`has_device` went through `get_lookup_keys`, which re-owns the already-resolved keys into a `Vec<String>` (1-2 heap allocs) on every received message that hits the device check. It now iterates the borrowed `resolve_lookup_keys(..).all_keys()` (bound once and shared by both the cache and backend loops), matching `get_devices_from_registry`. `get_lookup_keys` is now test-only.

## Tests

Each commit adds a test: `@call` JID_PAIR encode/decode round-trip plus parse/display; receipt `participant_pn` to `sender_alt` on both shapes; `for_own` routes empty bytes to removal and non-empty to set; `has_device` finds a device via the fallback lookup key. `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, and the wacore-binary + wacore + whatsapp-rust suites are green.

## Breaking

None.

### DIFF
--- a/src/client/device_registry.rs
+++ b/src/client/device_registry.rs
@@ -90,10 +90,9 @@ impl Client {
         UserLookupKeys::Unknown { user: user.into() }
     }
 
-    /// Get all possible lookup keys for a user (for bidirectional lookup).
-    /// Returns keys in order of preference: [canonical_key, fallback_key].
-    ///
-    /// Note: Prefer `resolve_lookup_keys` when you need type information.
+    /// Owned-key variant of `resolve_lookup_keys`. Test-only: production callers
+    /// use the borrowed `resolve_lookup_keys(..).all_keys()` to avoid the churn.
+    #[cfg(test)]
     pub(crate) async fn get_lookup_keys(&self, user: &str) -> Vec<String> {
         self.resolve_lookup_keys(user)
             .await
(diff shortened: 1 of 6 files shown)


jlucaso1:
@claude review

PR #732, opened 2026-06-05
refactor: split message/client/send monoliths into per-theme modules

jlucaso1:
This is a pure code-movement refactor that breaks up the three largest source files into module directories grouped by theme. There are no logic changes: the only edits that are not relocations are private to pub(crate) visibility bumps for items that are now referenced across modules or from tests, plus the re-export lines that keep every existing path working. No public API changes.

It follows the pattern that already exists in src/client/, where the root .rs file stays the module root and a sibling directory holds the per-theme submodules.

## What moved

src/message.rs (12467 to 167 LOC at the root). The root keeps the helper structs, the free fns and the RetryReason re-export. The impl Client decrypt/receive pipeline is split into message/{dispatch, msg_secret, retry, receive, special}.rs, and the tests move to message/tests.rs.

src/client.rs (7117 to 810 LOC at the root). The root keeps the Client struct, the public types (NodeFilter, ClientError, MemoryDiagnostics) and the ack free fns. The 107 impl Client methods are split into client/{lifecycle, node_io, app_state, iq_ops, adapters, messaging, accessors}.rs, and the tests move to client/tests.rs. The pre-existing client submodules (device_registry, lid_pn, sessions, sender_keys, offline_resume, context_impl) are untouched.

wacore/src/send.rs (5322 to 84 LOC at the root). The root keeps mod stanza, StanzaType and the re-exports that preserve the wacore::send::* paths. The free fns are split into send/{classify, encrypt, dm, peer, group, status}.rs, and the tests move to send/tests.rs.

## How paths stay intact

Each submodule is just `use super::*;` plus the moved block, so it inherits the root module imports through the glob. impl methods are path independent and only needed visibility bumps to pub(crate). Free functions are re-exported from the root via `pub use submod::*` (and a couple of `pub(crate) use` lines for the internal helpers that cross module boundaries), so `wacore::send::prepare_dm_stanza`, `crate::message::RetryReason` and friends resolve exactly as before.

## Verification

cargo fmt --all applied. cargo clippy --all-targets -- -D warnings is clean. cargo test --workspace --exclude e2e-tests passes with 3975 passed and 0 failed. Test counts are preserved: 147 message, 73 client, 89 send, none lost in the move.

The test files are now in dedicated tests.rs modules out of the production files. They are still large (message/tests.rs is around 9.5k lines), so splitting them further by topic is a possible follow-up that I left out of this PR to keep it a clean, low-risk move.


jlucaso1:
@claude review

jlucaso1:
@coderabbitai review

jlucaso1:
@coderabbitai review

PR #734, opened 2026-06-06
feat(metrics): optional metrics layer (off by default, Prometheus/OTLP-ready)

jlucaso1:
## What

Adds an opt-in `metrics` feature that emits `wa_*` metrics through the [`metrics`](https://docs.rs/metrics) facade for rates and latency percentiles, complementing the tracing spans from #733. Together: spans give the trace of one case, metrics give the rates/percentiles for dashboards.

Off by default: with the feature disabled there is no `metrics` dependency and zero overhead - every emit is an inlined no-op and the duration `Timer` is a zero-sized type that reads no clock. The library only emits; the application installs a recorder (Prometheus, OTLP, ...). See `examples/metrics.rs`.

## Design

Same philosophy as the tracing work: the library emits through a facade, the app owns the recorder; no Prometheus/OTLP dependency in the library.

- `metrics = { optional = true }` in `wacore`; feature `metrics = ["wacore/metrics"]` on the main crate, not in `default`.
- `wacore::telemetry` (re-exported as `whatsapp_rust::telemetry`) provides typed helpers (counters/gauges + a record-on-drop `Timer`), so call sites stay clean. Durations use the pluggable `wacore::time::Instant` so WASM/deterministic builds are unaffected.

## PII / cardinality

Labels are strictly low-cardinality categorical values (`outcome`/`kind`/`result`/`reason`). A JID, phone number or message id is never used as a label - it would explode the metrics backend and leak PII. The duration histograms are unlabeled (the matching `_total` counter carries the categorical breakdown).

## Metrics

Counters: `wa_recv_total{outcome}`, `wa_send_total{kind}`, `wa_retry_receipt_total{reason}`, `wa_iq_total{result}`, `wa_reconnect_total`, `wa_stream_error_total`, `wa_connect_total{outcome}`, `wa_appstate_sync_total{outcome}`, `wa_identity_change_total`, `wa_prekey_upload_total{outcome}`.

Histograms: `wa_iq_duration_seconds`, `wa_connect_duration_seconds`, `wa_decrypt_duration_seconds`, `wa_send_duration_seconds`, `wa_appstate_sync_duration_seconds`.

Gauges: `wa_connected` (plus a `wa_pending_retries` helper).

Emitted at the same boundaries as the `wa.*` spans. `RetryReason` gained a stable `as_str()` for the reason label.

## Verification

- `cargo clippy --all-targets -- -D warnings` clean both with and without `--features metrics`.
- `cargo fmt --all -- --check` clean.
- `cargo test --workspace --exclude e2e-tests`: 1992 passed, 0 failed. The instrumentation is additive and cfg-gated, so no behavior change.

## Usage

```
cargo run --example metrics --features metrics
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,10 +118,38 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic-waker"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
+
+[[package]]
 name = "autocfg"
 version = "1.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f2032f911046de80f0a198e0901378627c33f59ea0ac00e363d481118bd70a53"
+
+[[package]]
+name = "aws-lc-rs"
+version = "1.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5ec2f1fc3ec205783a5da9a7e6c1509cc69dedf09a1949e412c1e18469326d00"
+dependencies = [
+ "aws-lc-sys",
+ "zeroize",
+]
+
+[[package]]
+name = "aws-lc-sys"
+version = "0.41.0"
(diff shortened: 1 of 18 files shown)


jlucaso1:
@claude review

PR #748, opened 2026-06-08
fix(appstate): require snapshot MAC when validating (no silent skip)

jlucaso1:
## What

`process_snapshot` gated the entire aggregate snapshot-MAC validation behind `if validate_macs && let (Some(mac), Some(key_id)) = ...` with no `else`. If the server omitted the snapshot's `mac` (or top-level `key_id.id`), the check was silently skipped and all records were decoded and persisted without verifying the aggregate MAC that anchors the snapshot's ltHash. WA Web does the opposite: its anti-tampering compares against the (possibly undefined) mac and fires the recovery path on mismatch, so a missing mac is an integrity failure, not a success.

## Fix

When `validate_macs` is true, a missing `mac`/`key_id.id` now returns `AppStateError::SnapshotMACMismatch` instead of accepting unverified records. The same skip-on-absence is fixed in the standalone `validate_snapshot_mac` helper (it was a latent bypass).

`validate_macs == false` paths (tests, non-validating callers) are unchanged. Per-record content/index MACs were already validated; this restores the aggregate anchor against a malicious server omitting/reordering validly-keyed mutations.

## Verification

- `cargo clippy --all-targets -- -D warnings` clean
- `cargo test -p wacore-appstate` green (new test: missing mac + `validate_macs=true` → `SnapshotMACMismatch`)

Resolves gap-analysis protocol-21.

### DIFF
--- a/wacore/appstate/src/processor.rs
+++ b/wacore/appstate/src/processor.rs
@@ -86,13 +86,17 @@ where
         hex::encode(&initial_state.hash[120..])
     );
 
-    // Validate snapshot MAC if requested
-    if validate_macs
-        && let (Some(mac_expected), Some(key_id)) = (
+    // Validate snapshot MAC if requested. A snapshot that omits `mac`/`key_id` is
+    // treated as a validation FAILURE, not skipped: WA Web's anti-tampering
+    // compares against the (possibly undefined) mac and fires the recovery path on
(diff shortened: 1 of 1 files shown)


jlucaso1:
@coderabbitai review

PR #761, opened 2026-06-08
fix(groups): keep persisted group metadata in sync on membership change

jlucaso1:
Closes the features-32 gap.

A participant add/remove updated only the in-memory group cache, never the persisted metadata blob. The in-memory cache expires after ~1h; the next `query_info` then loads the **stale** persisted blob and sends an out-of-date participant phash, defeating the not-modified fast-path (forcing a needless full re-query) and diverging from the actual membership. `leave()` likewise dropped the in-memory entry but left the persisted blob behind.

The fix keeps the persisted blob in sync:
- Re-serialize and persist the mutated `GroupInfo` after a successful add/remove in **both** the API methods (`Groups::add_participants`/`remove_participants`) **and** the inbound group-notification handler (the more common path), via a shared `Client::persist_group_metadata` (same `serde_json` format `query_info` already uses).
- On `leave()`, delete the persisted blob through a new `Backend::delete_group_metadata` (default no-op; implemented for sqlite + in-memory) — we're no longer in the group, so a stale phash from it would only cause a needless re-query if ever read.

Verified against WA Web (`WAWebQueryGroupJob` persists updated group metadata on membership mutations so the next not-modified comparison stays consistent).

Test: in-memory `delete_group_metadata` round-trip (put → get → delete → get None).

### DIFF
--- a/src/features/groups.rs
+++ b/src/features/groups.rs
@@ -415,6 +415,17 @@ impl<'a> Groups<'a> {
     pub async fn leave(&self, jid: &Jid) -> Result<(), anyhow::Error> {
         self.client.execute(LeaveGroupIq::new(jid)).await?;
         self.client.get_group_cache().await.invalidate(jid).await;
+        // Drop the persisted blob too: we're no longer in the group, so a stale
+        // phash from it would only force a needless full re-query if ever read.
+        if let Err(e) = self
+            .client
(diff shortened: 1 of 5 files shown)


jlucaso1:
@coderabbitai review

PR #781, opened 2026-06-08
feat(send): emit member_label meta attrs (appdata + tag_reason)

jlucaso1:
`infer_stanza_metadata` (our `genMetaNode` equivalent) emitted `polltype` / `event_type` / `view_once`, but not the `<meta>` attributes WA Web's `MsgMetaNode` attaches to a `member_label` protocol message. So `Groups::update_member_label` went to the server without the meta attrs it expects.

Per `MsgMetaNode` functions `d` (appdata) and `p` (tag_reason): a member_label message carries `appdata="member_tag"`, and `tag_reason="user_delete"` when the label is cleared (empty/absent) or `"user_update"` otherwise. This adds that branch to the existing accumulate-onto-one-`<meta>` chain.

Scoped to `member_label`, the one such message we actually build and send (`build_member_label_message` → `update_member_label`): the peer `appdata="default"` case is already emitted on the PDO send path, and we don't send `MESSAGE_HISTORY_NOTICE` (`group_history`), so those branches would be dead code. Verified the mapping against `docs/captured-js/.../MsgMetaNode...js`.

### DIFF
--- a/src/features/groups.rs
+++ b/src/features/groups.rs
@@ -892,6 +892,11 @@ impl<'a> Groups<'a> {
             ));
         }
         let msg = wacore::send::build_member_label_message(label.into(), wacore::time::now_secs());
+        // This low-level send bypasses send_message_with_options (no reporting
+        // token for a protocol message), so compute the <meta> here and pass it
+        // as an extra node — otherwise the member_label appdata/tag_reason attrs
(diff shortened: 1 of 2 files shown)


jlucaso1:
@coderabbitai review

PR #797, opened 2026-06-09
fix(device-list): always keep the primary device after a raw_id mismatch patch

jlucaso1:
## Problem

When a device-add notification carries a `key-index-list` whose `raw_id` differs from the stored one (an identity rotation or relink), `patch_device_add` clears the device list and rebuilds it from the notification. `filter_devices_by_key_index` only keeps the primary (device 0) when it is already present in the input, so after the clear the primary is dropped: the record is persisted as just the notified companion (for example `[{device_id: 19}]`), or as an empty list when the notified device's key index is rejected.

A device record that exists but is missing the primary is a broken state. `get_user_devices` treats any present record as authoritative and only goes to the network when there is no record at all, so the bad list sticks and the self-healing usync re-fetch never fires. Downstream this leaves the user with a device set that does not cover the primary.

## Fix

Mirror WA Web's `WAWebHandleAdvDeviceNotificationApi`. In `handleDeviceAddNotification`, after filtering the existing devices and adding the notified ones, it unconditionally re-adds the primary with `C.push({ id: DEFAULT_DEVICE_ID, keyIndex: 0 })` (and the remove path does the same). The invariant is that the rebuilt device list always contains device 0.

`patch_device_add` now ensures device 0 is present after the rebuild, so a raw_id mismatch can never drop the primary or persist an empty list. whatsmeow keeps the same invariant on its device-list rebuild.

## Tests

`cargo test -p whatsapp-rust --lib client::device_registry::tests` (37 passing), full `cargo test -p whatsapp-rust --lib` (739 passing), `cargo test -p wacore`, and `cargo clippy --all-targets -- -D warnings`.

New tests:
- `test_patch_device_add_raw_id_mismatch_preserves_primary`: a raw_id mismatch clears the stale companion but keeps device 0 and adds the new companion. Guards the `[{device_id: N}]` (no primary) record.
- `test_patch_device_add_raw_id_mismatch_rejected_device_keeps_primary`: a mismatch where the notified key index is rejected rebuilds to just the primary instead of an empty list.
- Updated `test_patch_device_add_deduplicates` to the realistic seed (a record that includes the primary) and to assert the dedupe plus the primary invariant.

## Breaking

None.

### DIFF
--- a/src/client/device_registry.rs
+++ b/src/client/device_registry.rs
@@ -395,10 +395,14 @@ impl Client {
         // only keeps device 0 when it is already in the input, so after a clear the
         // primary would otherwise be lost, leaving a record with no device 0 (or an
         // empty list) that suppresses the usync re-fetch forever.
+        //
+        // The primary's key_index is never read (`filter_devices_by_key_index` keeps
+        // device 0 regardless and `is_key_index_valid` is not applied to it), so store
+        // `None` to match how device 0 is recorded everywhere else.
         if !record.devices.iter().any(|d| d.device_id == 0) {
             record.devices.push(wacore::store::traits::DeviceInfo {
                 device_id: 0,
-                key_index: Some(0),
+                key_index: None,
             });
         }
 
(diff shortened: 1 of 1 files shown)


jlucaso1:
@claude review

jlucaso1:
@coderabbitai review

PR #799, opened 2026-06-09
fix(usync): refetch an empty device record instead of trusting it

jlucaso1:
## Problem

`get_user_devices` only goes to the network when a user is completely absent from the device registry. A record that exists but is empty is returned as-is, and because a present record is treated as authoritative, the usync re-fetch never fires again. An empty device list means the user has no devices at all, so we cannot send to them, and the state never self-heals.

An empty device list is never valid: WA Web always keeps the primary (device 0) in a user's device list, so an empty record can only be local corruption.

## Fix

Treat an empty registry result as a miss. `get_user_devices` now falls through to the usync fetch when the resolved device list is empty, so a corrupted empty record self-heals from the server instead of leaving the user permanently unreachable.

This is intentionally narrow. A record with real devices (including the common single-device `[0]` case) is still served from the registry with no network round-trip, so there is no change to the hot send path. Only the empty case, which is always broken, is refetched.

## Tests

`cargo test -p whatsapp-rust --lib usync::tests` (7 passing), full `cargo test -p whatsapp-rust --lib` (738 passing), and `cargo clippy --all-targets -- -D warnings`.

New test `test_empty_device_record_falls_through_to_network` seeds a present-but-empty record and asserts `get_user_devices` reaches the network path instead of returning an empty result. It guards the exact regression: the old behavior returned `Ok([])` with no fetch.

## Breaking

None.

### DIFF
--- a/src/client/device_registry.rs
+++ b/src/client/device_registry.rs
@@ -636,7 +636,15 @@ impl Client {
         // L1: device_registry_cache (moka, fast)
         for key in lookup.all_keys() {
             if let Some(record) = self.device_registry_cache.get(key).await {
-                return Some(Self::reconstruct_device_jids(jid, &record));
+                let devices = Self::reconstruct_device_jids(jid, &record);
+                // An empty record is never a valid device set — WA Web always keeps
+                // the primary (device 0) — so read it as a miss instead of `Some([])`.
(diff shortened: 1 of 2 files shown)


jlucaso1:
@claude review

PR #802, opened 2026-06-09
fix(retry): resync the device list on a retry from an unknown device

jlucaso1:
## Problem

A retry from a device that is missing from our device registry is dropped (when it carries no `<keys>` bundle) without refreshing the device list. The previous fix recovers a retry that carries a key bundle, but a retry with no bundle is still just dropped, and nothing else updates that user's device list.

This leaves a real loop. A user links a new device after our last sync, our registry does not have it, and the device receives our group `skmsg` but never got a sender key, so it can only fail to decrypt and retry. The reconciliation that runs when a prekey fetch returns 406 during a send never fires for this device, because the device is not in the set we send to, so it never 406s. The device is never learned and retries forever.

## Fix

Treat a retry from an unknown device as the staleness signal it is: refresh that user's device list. `handle_retry_receipt` now schedules a device-list resync whenever the requester is missing from our registry, whether or not the retry carries a bundle to recover this particular message.

This reuses the existing unknown-device sync path (`handle_unknown_device_sync`), extracted into `schedule_unknown_device_sync`. It dedups per user (so a retry storm cannot fan out into a usync storm), then invalidates and re-fetches via usync. WA Web keeps device lists fresh by triggering `syncDeviceListJob` on an unknown device; this is the equivalent trigger on the retry path.

Once the resync learns the device, the next send includes it in the sender-key distribution and the retries stop.

## Tests

`cargo test -p whatsapp-rust --lib` (744 passing), `cargo clippy --all-targets -- -D warnings`.

New test `schedule_unknown_device_sync_batches_and_dedups` locks the dedup so repeated unknown-device retries enqueue a single sync.

## Breaking

None.

### DIFF
--- a/src/message/receive.rs
+++ b/src/message/receive.rs
@@ -1236,14 +1236,26 @@ impl Client {
     /// WA Web: online → `syncDeviceListJob`, offline → `OfflinePendingDeviceCache`.
     #[cfg_attr(feature = "tracing", tracing::instrument(name = "wa.recv.unknown_device_sync", level = "debug", skip_all, fields(sender = %info.source.sender.observe(), msg_id = %info.id)))]
     async fn handle_unknown_device_sync(self: &Arc<Self>, info: &MessageInfo) {
-        let user_jid = info.source.sender.to_non_ad();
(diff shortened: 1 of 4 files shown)


jlucaso1:
@claude review

PR #805, opened 2026-06-09
perf(receive): bound the per-chat message queue

jlucaso1:
## Problem

Each chat lane's mailbox (`src/handlers/message.rs`) is an `async_channel::unbounded`, so a flood into a single chat buffers `Arc&lt;OwnedNodeRef&gt;` entries without limit — and each queued entry keeps its decoded frame buffer alive, so queue depth translates directly into pinned memory. The enqueue path already has the right overflow handling (try_send failure → warn + `cancelled = true` → no ack → server redelivers), but on an unbounded channel that path can only fire for a closed lane; it was designed as backpressure that could never engage.

## Change

`bounded(CHAT_LANE_QUEUE_DEPTH = 2048)`. Sizing rationale: the deepest legitimate burst is offline resume, which is server-paced at 200 stanzas per `&lt;offline_batch&gt;` request (`src/client/offline_resume.rs`), so even a backlog landing entirely in one chat stays a few hundred deep — an order of magnitude under the bound. Only pathological floods trip it, and those now degrade to server redelivery (the stanza isn't acked, so it stays in the server's offline queue) instead of unbounded buffering.

## Trade-off (honest)

If the bound trips, a dropped stanza is redelivered later (offline replay), so a Signal message can be decrypted out of arrival order relative to its chat. The double ratchet handles skipped message keys, and the existing retry-receipt path covers a PreKey message deferred past a session message — same recovery as any out-of-order delivery today. At 10× the legitimate burst depth this is a degraded-flood mode, not a steady-state behavior change. The alternative (awaiting queue capacity in the handler) would block the read loop and head-of-line-block every other chat, which is worse.

## Verification

`cargo fmt` clean; `cargo clippy -p whatsapp-rust --tests` clean; `cargo test -p whatsapp-rust --lib`: 744 passed, 0 failed. The overflow path itself isn't unit-tested: the depth is a const and the lane worker drains concurrently, so filling 2048 entries deterministically would need either an injectable depth or a stalled-worker harness — happy to add either if preferred.

## Breaking

None (internal handler detail; `ChatLane`'s sender type is the same for bounded/unbounded).

---
_Generated by [Claude Code](https://claude.ai/code/session_01LKdomTCXXcp3V44UgbMuJf)_

### DIFF
--- a/src/handlers/message.rs
+++ b/src/handlers/message.rs
@@ -7,6 +7,15 @@ use std::sync::Arc;
 /// WA Web: `WAWebMessageQueue` uses `promiseTimeout(r(), 2e4)` per queued handler.
 const MAX_MESSAGE_DELAY_MS: u64 = 20_000;
 
+/// Per-chat queue depth. The deepest legitimate burst is offline resume,
+/// which is server-paced at 200 stanzas per `<offline_batch>` request
+/// (`offline_resume`), so even a backlog landing entirely in one chat stays
+/// a few hundred deep. Each queued entry keeps its decoded frame buffer
+/// alive, so an unbounded queue turns a flood into unbounded memory; above
+/// this depth the enqueue fails, the ack is cancelled, and the server
+/// redelivers the stanza instead.
+const CHAT_LANE_QUEUE_DEPTH: usize = 2048;
+
 /// Handler for `<message>` stanzas.
 ///
 /// Messages are processed sequentially per-chat using a mailbox pattern to prevent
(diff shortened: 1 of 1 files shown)


jlucaso1:
Closing this — the Codex P2 is correct, and it invalidates the premise of the PR. Verified both of its claims in code:

1. **Offline batching is arrival-driven, not processing-driven.** `on_offline_stanza_arrived` (called from `process_node` on each offline stanza arrival) CAS-elects the first stanza of the just-sent batch and schedules the next 200-stanza `&lt;offline_batch&gt;` request immediately (`src/client/offline_resume.rs`). Nothing waits for the chat lanes to drain.
2. **Processing is globally serialized during replay.** `cleanup`/connect swaps the message semaphore to **1 permit** during offline sync (`src/client/lifecycle.rs:636`, `src/message/receive.rs:307`), so lane drain rate is far below arrival rate by design.

Together: a single-chat offline backlog &gt; 2048 is a *legitimate* state during replay, and this bound would drop those stanzas — no delivery receipt, so they sit in the server's offline store until a later reconnect. That's a user-visible correctness regression to defend against a flood whose realistic ceiling is already bounded server-side (the offline store), which makes the unbounded queue the lesser evil.

The obvious repair — `send().await` when full instead of dropping — blocks the read loop (message handlers run inline), which would stall ping/pong and IQ-response processing and risk keepalive-triggered reconnects mid-replay. Not worth the fragility.

If queue memory during replay ever becomes a measured problem, the proper fix is processing-driven flow control in the offline pull loop (gate `schedule_continuation` on aggregate lane depth) — a different, deliberate divergence from WA Web's arrival-driven `$4` logic, deserving its own PR.

---

PR #817, opened 2026-06-09
fix(protocol): tighten error handling gaps

jlucaso1:
## Summary

This PR tightens several protocol/API paths that were accepting malformed data or hiding errors:

- reject App State snapshot/patch MAC mismatches even when `has_missing_remove` is set
- reject invalid AD-JID domain type bytes instead of mapping them to PN
- make `contacts().is_on_whatsapp()` fail on unsupported JID types instead of silently dropping inputs
- preserve USync subprotocol errors for user-info style queries while making device-list/LID batch queries degrade per returned user, matching WA Web behavior for noisy users in a larger batch
- reject IQ responses whose `type` is neither `result` nor `error`
- abort malformed peer-device retry receipts without `recipient`
- require the one-time `<key>` for regular retry key bundles; only primary `@bot` FBID bot retries may proceed with identity + signed prekey alone
- abort retry resend when a present key bundle is rejected
- downgrade missing ADV account-key fallback logs from warn to debug

## Impact

The main user-visible change is stricter failure signaling where WA Web treats malformed data as fatal: callers now get explicit errors or error metadata instead of ambiguous empty results, skipped inputs, or best-effort fallbacks.

Device-list and LID batch parsing now degrade per user instead of failing the whole batch, so one server-side per-user subprotocol error does not block sends to the rest of a group. Retry recovery follows the WA Web parser split: regular retry receipts need an OPK, while FBID bot retries keep the signed-prekey-only fallback.

The ADV log downgrade reduces noise for a server shape that can legitimately be unverifiable when no account identity is stored.

## Validation

- `cargo +nightly-2026-04-05 clippy --all --tests -- -D warnings`
- `cargo test -p wacore`
- `cargo test -p whatsapp-rust`
- `cargo test -p wacore device_list_`
- `cargo test -p wacore lid_query_lid_error_skips_only_that_user`
- `cargo test -p whatsapp-rust retry_key_bundle`
- `cargo test -p whatsapp-rust resolve_retry_chat_info`
- `cargo test -p whatsapp-rust is_on_whatsapp`
- `cargo test -p wacore-binary test_ad_jid -- --nocapture`
- `cargo test -p wacore-appstate validate_patch_macs_rejects -- --nocapture`

### DIFF
--- a/src/features/contacts.rs
+++ b/src/features/contacts.rs
@@ -90,7 +90,11 @@ impl<'a> Contacts<'a> {
                     known_lid: None,
                 });
             } else {
-                unreachable!("validated is_on_whatsapp JID type at entry");
+                debug_assert!(
+                    jid.is_pn() || jid.is_lid(),
+                    "validated is_on_whatsapp JID type"
+                );
+                continue;
             }
         }
 
--- a/src/features/contacts.rs
+++ b/src/features/contacts.rs
@@ -14,7 +14,7 @@ use wacore_binary::{Jid, JidExt};
 
 // Re-export types from wacore
 pub use wacore::iq::contacts::ProfilePicture;
-pub use wacore::iq::usync::{IsOnWhatsAppResult, UserInfo};
+pub use wacore::iq::usync::{IsOnWhatsAppResult, UserInfo, UsyncSubprotocolError};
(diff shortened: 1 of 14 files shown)


jlucaso1:
@claude review

PR #826, opened 2026-06-10
fix(contacts): use fn items for LID mapping extractors so boxed futures compile

### LINKED ISSUE #825: Lifetime Issue: `is_on_whatsapp()` Fails Within Async Trait + RwLock Context
## Summary

The `is_on_whatsapp()` method works correctly in standalone async functions but fails to compile with HRTB lifetime errors when used within `async_trait` implementations that involve `RwLock` guards. This appears to be a lifetime inference issue specific to complex async contexts rather than a bug in the method itself.

## Status

✅ **Works**: Standalone async functions  
❌ **Fails**: Within `async_trait` + `RwLock<Option<Arc<Client>>>` context

## Error Message

```
error: implementation of `FnOnce` is not general enough
   --> src/client.rs:413:5
    |
413 |     async fn check_phone_numbers(
    |     ^^^^^ implementation of `FnOnce` is not general enough
    |
    = note: closure with signature `fn(&'0 IsOnWhatsAppResult) -> (&wacore_binary::Jid, std::option::Option<&wacore_binary::Jid>)`
            must implement `FnOnce<(&'1 IsOnWhatsAppResult,)>`, for any two lifetimes `'0` and `'1`...
    = note: ...but it actually implements `FnOnce<(&IsOnWhatsAppResult,)>`
```

### ✅ Working: Standalone Example

This compiles and runs successfully:

```rust
use std::sync::Arc;
use wacore_binary::jid::Jid;
use whatsapp_rust::bot::Bot;
use whatsapp_rust::store::SqliteStore;
use whatsapp_rust::TokioRuntime;
use whatsapp_rust_tokio_transport::TokioWebSocketTransportFactory;
use whatsapp_rust_ureq_http_client::UreqHttpClient;

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let backend = Arc::new(SqliteStore::new("test.db").await?);

    let bot = Bot::builder()
        .with_backend(backend)
        .with_transport_factory(TokioWebSocketTransportFactory::new())
        .with_http_client(UreqHttpClient::new())
        .with_runtime(TokioRuntime)
        .build()
        .await?;

    let client = bot.client();

    // ✅ This works perfectly
    let jids = vec![Jid::pn("1234567890")];
    let results = client.contacts().is_on_whatsapp(&jids).await?;

    for result in results {
        println!("{}: {}", result.jid, result.is_registered);
    }

    Ok(())
}
```

### ❌ Failing: Async Trait + RwLock Example

This fails to compile:

```rust
use async_trait::async_trait;
use std::sync::Arc;
use tokio::sync::RwLock;
use wacore_binary::jid::Jid;
use whatsapp_rust::Client;

#[async_trait]
pub trait WhatsAppClient: Send + Sync {
    async fn check_phones(&self, phones: Vec<String>) -> Result<Vec<bool>, Box<dyn std::error::Error>>;
}

pub struct RealClient {
    client: RwLock<Option<Arc<Client>>>,
}

#[async_trait]
impl WhatsAppClient for RealClient {
    async fn check_phones(&self, phones: Vec<String>) -> Result<Vec<bool>, Box<dyn std::error::Error>> {
        // Get client from RwLock
        let client_guard = self.client.read().await;
        let client = client_guard.as_ref().unwrap().clone();
        drop(client_guard);

        // Convert phones to JIDs
        let jids: Vec<Jid> = phones.iter().map(|p| Jid::pn(p.as_str())).collect();

        // ❌ COMPILATION ERROR HERE
        // error: implementation of `FnOnce` is not general enough
        let results = client.contacts().is_on_whatsapp(&jids).await?;

        Ok(results.into_iter().map(|r| r.is_registered).collect())
    }
}
```

## Analysis

The error originates from the `persist_lid_mappings()` calls inside `is_on_whatsapp()`:

```rust
// In src/features/contacts.rs
self.persist_lid_mappings(results.iter().map(|r| (&r.jid, r.lid.as_ref()))).await;

self.persist_lid_mappings(results.iter().filter_map(|r| {
    if r.jid.is_lid() {
        r.pn_jid.as_ref().map(|pn| (pn, Some(&r.jid)))
    } else {
        None
    }
})).await;
```

These closures return references with lifetimes tied to the iterator items. In standalone contexts, Rust can infer the lifetimes correctly. However, when combined with:

1. `async_trait` (which boxes the future)
2. `RwLock` guards (which add lifetime constraints)
3. Complex generic bounds in the trait

The lifetime inference fails, resulting in the HRTB error.

## Root Cause

The issue is **not** a bug in `is_on_whatsapp()` itself, but rather a limitation in Rust's lifetime inference when dealing with:

- Async trait methods (which are desugared into complex futures)
- RwLock guards that must be held across await points
- Closures that capture references with complex lifetime relationships

## Attempted Workarounds (All Failed)

1. ❌ Cloning the `Arc<Client>` and dropping the guard early
2. ❌ Using explicit lifetime annotations
3. ❌ Moving client extraction into a separate scope
4. ❌ Using `into_iter()` instead of `iter()`
5. ❌ Extracting JID creation into a separate function

## Current Workaround

The only solution that works is to avoid calling `is_on_whatsapp()` from within async trait contexts:

```rust
#[async_trait]
impl WhatsAppClient for RealClient {
    async fn check_phones(&self, phones: Vec<String>) -> Result<Vec<bool>, Box<dyn std::error::Error>> {
        // Workaround: Return placeholder data
        tracing::warn!("Using placeholder due to lifetime issue in async trait context");
        Ok(vec![true; phones.len()])
    }
}
```

### Solution 1: Refactor `persist_lid_mappings()` to Accept Owned Values

Change the signature from:

```rust
async fn persist_lid_mappings<I>(&self, mappings: I)
where
    I: IntoIterator<Item = (&Jid, Option<&Jid>)>;
```

To:

```rust
async fn persist_lid_mappings<I>(&self, mappings: I)
where
    I: IntoIterator<Item = (Jid, Option<Jid>)>;
```

And clone the values before passing:

```rust
self.persist_lid_mappings(
    results.iter().map(|r| (r.jid.clone(), r.lid.clone()))
).await;
```

### Solution 2: Add Helper Method That Returns Owned Results

```rust
impl Contacts {
    pub async fn is_on_whatsapp_owned(&self, jids: &[Jid]) -> Result<Vec<IsOnWhatsAppResultOwned>> {
        let results = self.is_on_whatsapp(jids).await?;
        // Convert to owned version without internal closures
        Ok(results.into_iter().map(|r| IsOnWhatsAppResultOwned {
            jid: r.jid,
            lid: r.lid,
            pn_jid: r.pn_jid,
            is_registered: r.is_registered,
            is_business: r.is_business,
            verified_name: r.verified_name,
        }).collect())
    }
}
```

### Solution 3: Document the Limitation

Add documentation noting that `is_on_whatsapp()` may not work in all async contexts and provide the standalone example as a reference.

## Environment

- **whatsapp-rust version**: 0.6.0
- **Rust version**: 1.85.0 (stable)
- **async-trait version**: 0.1.83
- **tokio version**: 1.48
- **OS**: Windows 11

## Impact

This limitation prevents downstream libraries from using `is_on_whatsapp()` within trait-based architectures, which is a common pattern in Rust async codebases. Users must either:

1. Use standalone functions (not ideal for abstraction)
2. Use placeholder implementations (loses functionality)
3. Restructure their entire architecture to avoid async traits

## Additional Context

This appears to be related to known Rust compiler limitations with:

- Higher-Ranked Trait Bounds in async contexts
- Lifetime elision across await points
- Complex interactions between `async_trait` macro expansion and lifetime inference

Similar issues have been reported in other async Rust libraries when combining traits, RwLock, and closures that return references.

## References

- [Rust Issue #63033: Lifetime inference fails in async fn](https://github.com/rust-lang/rust/issues/63033)
- [async-trait Limitations](https://github.com/dtolnay/async-trait#dyn-traits)
- [Tokio RwLock Documentation](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html)

---

**Expected behavior**: `is_on_whatsapp()` should work in all contexts, including within async traits.

**Actual behavior**: Works in standalone functions but fails within async trait + RwLock contexts.

jlucaso1:
## Problem

`is_on_whatsapp()` and `get_user_info()` failed to compile for any consumer that boxes their futures, which is exactly what `#[async_trait]` does:

```
error: implementation of `FnOnce` is not general enough
  = note: closure with signature `fn(&'0 IsOnWhatsAppResult) -> (&Jid, Option<&Jid>)`
          must implement `FnOnce<(&'1 IsOnWhatsAppResult,)>`, for any two lifetimes `'0` and `'1`...
```

Reported in #825 with an excellent minimal repro. The `RwLock` in the report is incidental: the trigger is the `Box<dyn Future + Send>` that `async_trait` produces, which forces the compiler to prove auto traits and higher-ranked bounds for the whole future tree. The real culprit lives in this library: the `persist_lid_mappings` call sites passed closures returning references tied to their argument (`|r| (&r.jid, r.lid.as_ref())`). Rust infers such closures at a concrete lifetime rather than the higher-ranked `for<'r> Fn(&'r _)`, and since the closure types are embedded in the public methods' future types, the unprovable obligation leaks to every boxing consumer. None of the user-side workarounds can fix it, since the problem is inside the library's future type.

## Fix

Replace the three borrowing closures (two in `is_on_whatsapp`, one in `get_user_info`) with fn items, which implement `Fn` for every lifetime by construction. Zero clones, zero public API change, identical behavior.

Swept the rest of the workspace for the same class: the other generic helpers with borrowed-item bounds (`GroupInfo::add_participants`, `DeviceTopology::record`, `DeviceRegistryCache::insert`, `participant_list_hash`) are either synchronous (the iterator never lives inside a future) or only ever receive closure-free iterators (slices/`Chain`), and the `FDownload` closures in appstate return owned values, which generalize fine. The contacts call sites were the only affected surface.

## Tests

New `tests/async_trait_boxed_future_compat.rs`: a compile-time regression guard reproducing the consumer shape from the report (`#[async_trait]` + `RwLock<Option<Arc<Client>>>` calling both methods). This class of error only manifests in the downstream crate's context, so the guard IS the compilation. Verified it catches the regression: with the previous closures restored, the guard fails with six "not general enough" errors; with the fn items it compiles and passes.

- `cargo fmt --all`
- `cargo clippy --all-targets -- -D warnings`
- `cargo test -p whatsapp-rust --lib` (765 passing)
- `cargo test -p whatsapp-rust --test async_trait_boxed_future_compat`

## Breaking

None.

Fixes #825

### DIFF
--- a/src/features/contacts.rs
+++ b/src/features/contacts.rs
@@ -30,10 +30,12 @@ fn ensure_is_on_whatsapp_jids_supported(jids: &[Jid]) -> Result<()> {
 /// that box those futures (`#[async_trait]`, `Box<dyn Future + Send>`) hit
 /// "implementation of `FnOnce` is not general enough" (issue #825). Fn items
 /// implement `Fn` for every lifetime by construction.
+/// PN-primary result -> (PN, LID mapping).
 fn forward_lid_pair(r: &IsOnWhatsAppResult) -> (&Jid, Option<&Jid>) {
     (&r.jid, r.lid.as_ref())
 }
 
+/// LID-primary result inverted to (PN, LID); None when not LID-primary.
 fn reverse_lid_pair(r: &IsOnWhatsAppResult) -> Option<(&Jid, Option<&Jid>)> {
     if r.jid.is_lid() {
         r.pn_jid.as_ref().map(|pn| (pn, Some(&r.jid)))
@@ -42,6 +44,7 @@ fn reverse_lid_pair(r: &IsOnWhatsAppResult) -> Option<(&Jid, Option<&Jid>)> {
     }
 }
 
+/// UserInfo entry -> (queried JID, LID mapping).
 fn user_info_lid_pair(entry: &UserInfo) -> (&Jid, Option<&Jid>) {
     (&entry.jid, entry.lid.as_ref())
 }
@@ -55,6 +58,10 @@ impl<'a> Contacts<'a> {
         Self { client }
     }
 
+    /// Callers must pass `fn` items (e.g. [`forward_lid_pair`]), NOT
+    /// closures: a closure returning borrowed pairs embeds a non-HRTB type in
+    /// the public caller's future and breaks `#[async_trait]` consumers
+    /// (issue #825, guarded by tests/async_trait_boxed_future_compat.rs).
     async fn persist_lid_mappings<'b, I>(&self, entries: I)
     where
         I: IntoIterator<Item = (&'b Jid, Option<&'b Jid>)>,
--- a/src/features/contacts.rs
+++ b/src/features/contacts.rs
@@ -24,6 +24,28 @@ fn ensure_is_on_whatsapp_jids_supported(jids: &[Jid]) -> Result<()> {
     Ok(())
 }
 
+/// Mapping extractors as fn items, NOT closures. A closure returning
+/// references tied to its argument is inferred at a concrete lifetime, and
+/// because its type is embedded in the public methods' future types, callers
+/// that box those futures (`#[async_trait]`, `Box<dyn Future + Send>`) hit
+/// "implementation of `FnOnce` is not general enough" (issue #825). Fn items
+/// implement `Fn` for every lifetime by construction.
+fn forward_lid_pair(r: &IsOnWhatsAppResult) -> (&Jid, Option<&Jid>) {
+    (&r.jid, r.lid.as_ref())
+}
+
+fn reverse_lid_pair(r: &IsOnWhatsAppResult) -> Option<(&Jid, Option<&Jid>)> {
+    if r.jid.is_lid() {
+        r.pn_jid.as_ref().map(|pn| (pn, Some(&r.jid)))
+    } else {
+        None
+    }
+}
+
+fn user_info_lid_pair(entry: &UserInfo) -> (&Jid, Option<&Jid>) {
+    (&entry.jid, entry.lid.as_ref())
+}
+
 pub struct Contacts<'a> {
     client: &'a Client,
 }
@@ -112,16 +134,10 @@ impl<'a> Contacts<'a> {
             results.extend(self.client.execute(spec).await?);
         }
 
-        self.persist_lid_mappings(results.iter().map(|r| (&r.jid, r.lid.as_ref())))
+        self.persist_lid_mappings(results.iter().map(forward_lid_pair))
             .await;
-        self.persist_lid_mappings(results.iter().filter_map(|r| {
-            if r.jid.is_lid() {
-                r.pn_jid.as_ref().map(|pn| (pn, Some(&r.jid)))
-            } else {
-                None
-            }
-        }))
-        .await;
+        self.persist_lid_mappings(results.iter().filter_map(reverse_lid_pair))
+            .await;
 
         Ok(results)
     }
@@ -189,7 +205,7 @@ impl<'a> Contacts<'a> {
         let spec = UserInfoSpec::new(jids.to_vec(), request_id);
 
         let info = self.client.execute(spec).await?;
-        self.persist_lid_mappings(info.values().map(|entry| (&entry.jid, entry.lid.as_ref())))
+        self.persist_lid_mappings(info.values().map(user_info_lid_pair))
             .await;
         Ok(info)
     }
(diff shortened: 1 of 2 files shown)


jlucaso1:
@claude review

PR #839, opened 2026-06-11
perf(libsignal): cache the verify-side Edwards derivations per sender key

jlucaso1:
## Problem

The receive-side twin of #838. The CodSpeed flamegraph for `bench_signature_verification` (pulled via the MCP) shows `MontgomeryPoint::to_edwards` at 267.7 µs and the compression of A at ~120 µs per verification, both built on field inversions, together ~24% of a verify. Both derive solely from the signer's public key and the signature's sign bit, and a given signer always produces the same sign bit. But `group_decrypt` called `signing_key_public()` per message, rebuilding the `PublicKey` from the record's protobuf bytes, so every incoming group message re-derived them.

## Change

- `PreparedVerifyingKey`: a verifying view of a `PublicKey` caching `(-A, A_compressed)` per sign bit behind `OnceLock`s (clones carry initialized entries). Lives in `core::curve`, exported from `protocol`.
- The XEdDSA verify equation is extracted into a shared `verify_signature_prepared` in curve25519.rs; the existing `PrivateKey::verify_signature` now derives its inputs and delegates, so the plain and cached paths share one implementation and cannot drift.
- `SenderKeyState` memoizes the verifier next to the signing-key memo from #838 (same lifecycle: seeded at creation, never persisted, lazily rebuilt after a cold load, persists for the record-cache lifetime), exposed as `signing_key_verifier()` returning a reference (no clone per message).
- `group_decrypt` verifies via `SenderKeyMessage::verify_signature_prepared`. The plain `verify_signature(&PublicKey)` remains for other callers.

## Measurements

2000-message `group_decrypt` loop, core-pinned `perf stat`, A/B against main:

| | main | this PR |
| --- | --- | --- |
| wall per decrypt | 45.0 µs | 33.9 µs (-25%) |
| instructions per decrypt | 682K | 571K (-16.2%) |

Expect `bench_group_decrypt_message` and `bench_group_recv` to drop on the CodSpeed report.

## Tests

- Differential test pinning `prepared == plain` across: valid signatures under BOTH sign bits (corpus asserts both were exercised), corrupted signatures, corrupted messages, flipped sign bits, wrong-length signatures, garbage keys, and warmed clones.
- Memo lifecycle test (seeded at creation, cold after protobuf roundtrip, warm after first use, clones carry it).
- Full workspace suite, strict clippy, wasm32 CI builds pass.

## Notes

The verify hot core (`vartime_double_scalar_mul_basepoint`, ~59% of a verify) is irreducible crypto and untouched. Per-signature data (R, s, the message hash and the final compression of R') is computed per call as before.

### DIFF
--- a/wacore/libsignal/src/core/curve.rs
+++ b/wacore/libsignal/src/core/curve.rs
@@ -237,6 +237,82 @@ use curve25519_dalek::edwards::CompressedEdwardsY;
 use curve25519_dalek::scalar::Scalar;
 use std::sync::OnceLock;
 
+use curve25519_dalek::edwards::EdwardsPoint;
+use curve25519_dalek::montgomery::MontgomeryPoint;
+
+/// A verifying view of a [`PublicKey`] that caches the per-key XEdDSA
+/// derivations: the Montgomery-to-Edwards conversion and its compressed
+/// encoding, per sign bit (the bit travels in each signature, and a given
+/// signer always produces the same one). Repeat verifications of one key,
+/// e.g. every incoming message under a group sender key, skip two field
+/// inversions per signature.
+///
+/// Clones carry initialized entries, so a warmed instance stays warm through
+/// the per-use clones of whatever memoizes it.
+/// `(-A, A_compressed)` for one sign bit of a verifying key.
+type PreparedEdwards = (EdwardsPoint, [u8; 32]);
+
+#[derive(Clone)]
+pub struct PreparedVerifyingKey {
+    mont: [u8; 32],
+    /// Per sign bit; `None` when the key has no Edwards form for that bit
+    /// (such a signature can never verify).
+    cached: [OnceLock<Option<PreparedEdwards>>; 2],
+}
+
+impl PreparedVerifyingKey {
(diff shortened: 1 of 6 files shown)


jlucaso1:
Pushed a06540f6 addressing the first CodSpeed report: the Edwards entries now live behind an Arc shared by all clones (8 bytes per state, one allocation per key, warmth propagates to every holder including the cached master), which removes the ~460 B-per-clone memory cost the instrument flagged across the group benches. Receive-side states also pre-derive both sign-bit entries at SKDM processing (once per sender rotation), so the decrypt benches measure the warm steady state like production does instead of paying the first-use derivation inside their fresh-fixture window. Local 2000-message decrypt loop: 1363.5M (main) -> 1140.8M instructions.

Also investigated going further with dalek's VartimeEdwardsPrecomputation (per-key NAF8 tables for [-A, B]): measured 2.8x SLOWER than the specialized vartime_double_scalar_mul_basepoint (6.5 -> 18.1 us/op), so the specialized routine stays. The remaining verify core is irreducible.

PR #845, opened 2026-06-11
perf(docker): enable -Zshare-generics in the image build

jlucaso1:
## What

Adds `-Zshare-generics=y` to the Docker image's `RUSTFLAGS`.

## Why

The cargo-bloat audit showed that rustc instantiates generics and async state machines in every crate that uses them; the per-crate symbol copies carry distinct instantiating-crate hashes that even fat LTO cannot merge. #842/#843 removed the biggest sources at the library level, but the long tail (moka cache machinery, async send graphs reached from handler closures, drop glue) is inherent to how stable rustc codegens cross-crate. `share-generics` is the upstream switch for exactly this: downstream crates reuse upstream monomorphizations instead of re-codegening them.

## Measured (release bin, current main)

- `.text`: 11.65 MiB -> 11.00 MiB (**-666 KiB, -5.6%**)
- cross-crate duplicate-symbol waste: 1414 KiB -> 475 KiB
- consumer-crate reinstantiation: 1484 KiB -> 531 KiB

Cumulative with #842/#843/#844 the audit took `.text` from 13.03 MiB to 11.00 MiB (-15.6%).

## Safety

- Nightly-only flag, and this image already pins the nightly toolchain through `rust-toolchain.toml`. The stable CI lane and stable consumers are untouched: the flag lives only in the Dockerfile.
- The historical reason share-generics is off in release builds is lost cross-crate inlining of shared instantiations; with `lto = "fat"` that does not apply, since LTO sees all bitcode and re-inlines freely.
- CodSpeed cannot exercise a Dockerfile-only flag, so the functional validation is local: the full wacore + lib suites built and run with the flag pass (1838 tests).

Note for downstream images (e.g. Veloz): the same one-line flag applies to any consumer building on the pinned nightly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,8 @@
 FROM rust:alpine AS chef
 RUN apk add --no-cache musl-dev
 COPY rust-toolchain.toml .
-RUN rustup show && cargo install cargo-chef --locked
+# rust-src feeds -Zbuild-std in the builder stage.
+RUN rustup show && rustup component add rust-src && cargo install cargo-chef --locked
 WORKDIR /app
 
 FROM chef AS planner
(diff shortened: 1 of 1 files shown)


jlucaso1:
@coderabbitai review

jlucaso1:
@claude review

PR #870, opened 2026-06-15
fix(retry): cap outbound resends per (chat, msg, requester)

Salientekill:
## Problem

`handle_retry_receipt` resends a message on **every** incoming retry receipt, with no ceiling independent of the peer-echoed `count`. The existing `MAX_RETRY_COUNT = 5` trusts that echoed count, which a looping device keeps low — so it never fires.

Under WhatsApp's mass PN→LID migration this turns into a resend storm. Observed in production on a bot running this lib:

- **36,374 resends in 66h**, sustained 9–25/min.
- **99.4% to a single group**; one `msg_id` resent **576 times**.
- Ended in `<failure reason="403" location="vll"/>` → **AccountLocked**.

`freshSKDM == resends` 1:1 across every log — each resend is driven by a `Marked … for fresh SKDM … due to retry receipt`. Notably `fix/device-list-preserve-primary` (#797) was already active during the incident and did **not** prevent the storm: the existing fixes reduce churn at the source but none impose a **volume ceiling**.

## Fix

A hard cap (`MAX_RESENDS_PER_MESSAGE = 8`) on the resends we actually perform, keyed by the **existing** processing key (`chat:msg:requester`) via a new `resend_counts` cache (10m TTL, 10k entries):

- Counts real resends, independent of the peer-echoed `count` — the ceiling `MAX_RETRY_COUNT` can't provide.
- **Per-requester** granularity: legitimate resends to distinct failing devices keep working; only a single device looping the same message is capped.
- On exceed: refuses the resend with a `warn!` naming the chat/device (so storm groups are visible).
- `resend_counts` exposed in `ClientStats` for monitoring.

## Test plan

- [x] `cargo test -p whatsapp-rust resend` — new unit tests for the cap arithmetic/threshold (`resend_cap_refuses_after_threshold`, `next_resend_count_saturates`).
- [x] `cargo fmt --all`, `cargo clippy -p whatsapp-rust` clean.
- [x] Downstream bot compiles against the change.

<a href="https://cubic.dev/pr/oxidezap/whatsapp-rust/pull/870?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/cache_config.rs
+++ b/src/cache_config.rs
@@ -195,6 +195,12 @@ pub struct CacheConfig {
     /// Session-recreate throttle history (time_to_live). Default: 1h TTL, 256
     /// entries. Replaces a global `Mutex<HashMap>` scanned O(n) per retry receipt.
     pub session_recreate_history: CacheEntryConfig,
+    /// Outbound resend cap per (chat, msg, requester): a hard ceiling on how
+    /// many times we re-encrypt+resend a message in response to peer retry
+    /// receipts, independent of the peer-echoed `count`. Without it a single
+    /// looping device under PN→LID churn drives a resend storm into
+    /// AccountLocked. Default: 10m TTL, 10000 entries.
+    pub resend_counts: CacheEntryConfig,
 
     // --- Coordination caches (capacity-only, no TTL) ---
     /// Per-device Signal session lock capacity. Default: 10000.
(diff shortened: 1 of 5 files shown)


Salientekill:
Addressed the resend-cap timing issue raised by CodeRabbit and cubic (790b9e0a).

The counter was incremented before a resend was confirmed, so the cache-miss and status-broadcast early returns consumed budget without sending — which could suppress later legitimate resends. Fixed by splitting the **check** (early, read-only) from the **tally** (moved past those early returns, right after the `Resending message` log).

One deliberate choice: the tally counts a *committed attempt*, not only a successful `send_node`. A session-prepare loop that keeps erroring is exactly the churn the cap is meant to bound; counting only successes would leave that loop unbounded. The cache-miss/status early returns (which send nothing) are excluded, which was the actual concern. Cap semantics (8 resends per chat:msg:requester) are unchanged.

jlucaso1:
@Salientekill valeu pelos logs, segui uma abordagem diferente do retry, como mudava muita coisa resolvi criar outro PR

PR #880, opened 2026-06-15
perf(signal): check out sender keys on load instead of deep-cloning

jlucaso1:
## What

`load_sender_key` returned an owned `SenderKeyRecord` by deep-cloning the cached `Arc` — the cache kept its copy, so the `Arc` was never unique and `Arc::unwrap_or_clone` always cloned. `group_decrypt` (and group encrypt) mutate the record and store it back, so every group decrypt deep-cloned the entire `SenderKeyState` backlog (a `VecDeque` of up to `MAX_MESSAGE_KEYS` message keys per state).

This mirrors the checkout that sessions' `load_session` already does:
- `get_sender_key` stays a **read-only `Arc` peek** (refcount bump, used by the send paths that just check existence).
- a new `checkout_sender_key` **takes the record out** of the cache, leaving the slot `CheckedOut`, so the cached `Arc` is unique and the load is a **move, not a clone**. `store_sender_key` puts it back.
- flush and eviction **skip `CheckedOut`** entries (deferred until stored back), exactly like the session path.

## Why it's safe

It's the same `CheckedOut` state machine sessions already use in production (`SessionEntry` / `get_session`). A peek of a checked-out key reports absent — which is safe because every current peek is on the **local sending key** (`send.rs`, `features/signal.rs` SKDM checks) while the load is on a **remote sender's key**, so they never touch the same cache entry; receives of the same key are serialized per-chat by `message_enqueue_locks`. `unwrap_or_clone` still falls back to a clone if a peek's `Arc` is somehow alive, so correctness never depends on uniqueness. No wire/protocol change.

## Measurement

The win is concentrated in the **offline catch-up / out-of-order backlog** path, where the sender-key state is large. A prior bench analysis (#858) put this clone at ~56% of `bench_group_out_of_order_decrypt_worst_case`. Steady-state group messaging (small sender-key state) is unaffected. CodSpeed's `bench_group_out_of_order_decrypt_worst_case` and `bench_group_decrypt` track the delta on this PR.

## Tests

New unit tests in `signal_cache.rs`: checkout lifecycle + peek/second-checkout-during-checkout report absent; cold load from the backend; and flush defers a checked-out key (not persisted while borrowed, persisted once stored back). The existing `warm_sender_key_hit_shares_arc_not_deep_clone` (peek still shares the `Arc`) is unchanged. `cargo test -p wacore` (1031) and `-p whatsapp-rust --lib` (827) pass; `cargo clippy --all-targets` clean.

## Provenance

Found while profiling the messaging hot path with CodSpeed and confirmed against the code — the group analog of the session Arc-move, which sender keys hadn't received.

<a href="https://cubic.dev/pr/oxidezap/whatsapp-rust/pull/880?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/store/signal_adapter.rs
+++ b/src/store/signal_adapter.rs
@@ -285,13 +285,14 @@ impl wacore::libsignal::protocol::SenderKeyStore for SenderKeyAdapter {
     > {
         let device = self.0.device.read().await;
         // group_decrypt mutates the loaded record (catch-up + ratchet) and stores
-        // it back, so the trait needs an owned copy. The cache keeps its `Arc`, so
-        // this clones the inner record (unchanged from the prior behavior).
+        // it back, so the trait needs an owned copy. checkout_sender_key takes the
+        // record out of the cache (leaving it CheckedOut), so the common case is a
+        // move rather than a deep clone of the message-key backlog; store_sender_key
+        // puts it back. Same checkout pattern as sessions' load_session.
         self.0
             .cache
-            .get_sender_key(sender_key_name, &*device.backend)
+            .checkout_sender_key(sender_key_name, &*device.backend)
             .await
-            .map(|opt| opt.map(std::sync::Arc::unwrap_or_clone))
             .map_err(signal_err("backend"))
     }
 
--- a/wacore/src/store/signal_cache.rs
+++ b/wacore/src/store/signal_cache.rs
@@ -169,11 +169,20 @@ impl SessionStoreState {
 
 // === Sender key object cache (same pattern as sessions) ===
 
+/// Mirrors [`SessionEntry`]: the cache holds an `Arc` so a read-only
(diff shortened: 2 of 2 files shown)


jlucaso1:
Closing: the checkout/move approach is unsafe for sender keys (group_decrypt/encrypt have no store-on-error or rollback, and outgoing sends aren't serialized, so a CheckedOut slot leaks on the common NoSenderKeyState error and a local-key peek races a concurrent send into a spurious SKDM). Replacing with a structural-sharing approach (persistent message-key backlog) that keeps the trait/concurrency/error semantics intact.

PR #887, opened 2026-06-17
feat(retry): WA Web log-level parity and retry-flow observability counters

jlucaso1:
## What

Two threads, one commit per change.

Log levels (2 commits): two retry-flow `warn!`s that fire for benign, remote-driven, fully-handled conditions move to the level WA Web uses for them. `Refusing retry #N ... exceeds max attempts` (the MAX_RETRY loop guard) goes to `debug!`, and `Base key collision ... Forcing fresh session` goes to `info!`. The collision line and its `Base key changed` sibling also start carrying the `message_id`.

Observability (4 commits): four low-cardinality, PII-safe, zero-cost-when-off counters on the retry flow, plus the retry count recorded on the existing receipt span.

## Why

This came out of triaging ~14h of production logs from a group bot. The fleet was healthy (0 ERROR, 0 panic, 0 stanza errors), but the entire WARN budget was 25 lines of exactly two kinds, both produced by a single peer device thrashing prekeys during one group send. It was remote-driven and fully contained by the existing defenses (the MAX_RETRY cap, the per-chat resend rate limiter, the SKDM gate, base-key collision detection); the message was delivered and no ban occurred. A cross-model review (Codex/GPT-5.5) verified the oracle citations independently before this PR.

The core finding is that WA Web emits both of those conditions via `WALogger.LOG` (level 2, informational), not `WARN` (level 3):

- `WAWebHandleRetryRequest` logs the `retryCount >= MAX_RETRY` (5) refusal via `WALogger.LOG`, and reserves `WARN` in the same module for real errors (no-requester, device-not-found).
- `WAWebUpdateLocalSignalSession` logs the same-base-key session delete via `WALogger.LOG`.
- `WAWebLogger`: `LOG` is level 2, `WARN` is level 3.

`debug!` (rather than `info!`) for the cap matches the lib's own precedent for a frequent, expected, remote-driven event: the mirror receive-side capped-retry path in `message/retry.rs` already logs at `debug!`, and `decrypt_fail_log_level` downgrades expected fan-out failures to debug to avoid WARN spam. `info!` for the collision matches the three sibling branches of `update_local_signal_session`, which already log at info. The `message_id` is added because the collision is keyed by `(address, message_id)`, so the id makes the event correlatable to a specific message during a storm.

On observability: WA Web emits two dedicated WAM events on this flow that the lib had no aggregate equivalent for. `MessageHighRetryCount` (id 3132, committed at `retryCount >= 5` from the send-receipt path) maps to `wa_high_retry_total`, and `MdRetryFromUnknownDevice` (id 2178, committed when a retry arrives from a device not in our registry, tagged primary vs companion) maps to `wa_retry_unknown_device_total`, placed at the retry call-site rather than inside `schedule_unknown_device_sync` (which the shared inbound path also drives, so it would over-count). Two more counters cover the responder events WA Web only logs and has no dedicated WAM for: `wa_retry_refused_total` and `wa_base_key_collision_total`, so a chronically thrashing peer shows up in aggregate instead of via log scraping. All four are off by default and compile to no-ops without the `metrics` feature. The span change records the retry `count` so storm depth is aggregable per sender even when the cap returns early.

## What this PR deliberately leaves out

The retry logic itself is verbatim parity with the oracle and is untouched: the cap value and comparator, the base-key save/compare/delete, and the recovery path all stay as-is. Two latent parity divergences the review surfaced (peer regId-mismatch without `<keys>`, and offline-gating of the destructive key-bundle path) are intentionally not addressed here: both have zero occurrences in the analyzed logs, carry moderate regression risk on session recovery, and deserve a separate decision.

## Tests

`cargo clippy --all-targets -- -D warnings` is clean both at default (no-op telemetry, tracing off) and with `--features metrics,tracing` (real counters plus the span field).
`cargo test -p wacore --lib` (991) and `cargo test -p whatsapp-rust --lib` (827) pass.
There is no behavior change: every line is a log level, an added log field, a counter increment, or a span field.

<a href="https://cubic.dev/pr/oxidezap/whatsapp-rust/pull/887?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/message/retry.rs
+++ b/src/message/retry.rs
@@ -247,15 +247,16 @@ impl Client {
                 reason
             );
         }
-        if retry_count >= MAX_DECRYPT_RETRIES {
-            // Parity with WA Web's MessageHighRetryCount WAM event (id 3132),
-            // which commits at retryCount >= MAX from the send-receipt path.
-            wacore::telemetry::high_retry(reason.as_str());
-        }
-
         let retry_sent = match self.send_retry_receipt(info, retry_count, reason).await {
             Ok(()) => {
                 wacore::telemetry::retry_receipt(reason.as_str());
+                if retry_count >= MAX_DECRYPT_RETRIES {
+                    // Parity with WA Web's MessageHighRetryCount WAM event (id
+                    // 3132): committed after the retry receipt is sent, not
+                    // before — WAWebHandleMsgSendReceipt awaits sendRetryReceipt
+                    // and only then calls maybePostMessageHighRetryCountMetric.
+                    wacore::telemetry::high_retry(reason.as_str());
+                }
                 debug!(
                     "Sent retry receipt #{} for message {} in chat {} from {} [{:?}]",
                     retry_count,
--- a/src/message/retry.rs
+++ b/src/message/retry.rs
@@ -247,6 +247,11 @@ impl Client {
                 reason
             );
         }
+        if retry_count >= MAX_DECRYPT_RETRIES {
+            // Parity with WA Web's MessageHighRetryCount WAM event (id 3132),
+            // which commits at retryCount >= MAX from the send-receipt path.
+            wacore::telemetry::high_retry(reason.as_str());
+        }
 
         let retry_sent = match self.send_retry_receipt(info, retry_count, reason).await {
             Ok(()) => {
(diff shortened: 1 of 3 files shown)


jlucaso1:
@coderabbitai review

jlucaso1:
@claude review

PR #892, opened 2026-06-18
refactor: move the demo binary to examples/ and make env_logger a dev-dependency

jlucaso1:
## What

The demo app at `src/main.rs` was auto-detected as the package binary, so `env_logger` (and its transitive crates) had to be declared in `[dependencies]` and was pulled onto every library consumer. This moves the demo to `examples/demo.rs` and makes `env_logger` a dev-dependency, so it is only built for examples and tests.

The `whatsapp-rust` binary no longer exists; run the demo with:

```bash
cargo run --example demo                   # QR code only
cargo run --example demo -- -p 15551234567 # Pair code + QR code
```

Two consumers of the old bin are updated accordingly:

- `scripts/ci/measure_binary_size.py` now builds and bloat-analyzes `--example demo` and reads the artifact from `target/release/examples/demo`. The metric series names are unchanged, so the size chart history is preserved.
- `Dockerfile` builds `--example demo` and copies the example artifact from `release/examples/demo` (the runtime entrypoint is unchanged).

`README.md` and `agent_docs/binary_size_ci.md` are updated to match.

## Why

`env_logger` was a binary-only dependency leaking into the public dependency graph purely because the demo lived inside the package. Moving it to `examples/` and downgrading `env_logger` to a dev-dependency removes that cost for downstream library users while keeping the demo runnable via `cargo run --example demo`.

## Verification

- `cargo clippy --all-targets -- -D warnings` is clean (compiles every example, including `demo`, plus tests).
- `cargo test -p whatsapp-rust` passes.
- `python3 -m py_compile scripts/ci/measure_binary_size.py` succeeds.
- `env_logger` no longer appears under `[dependencies]` (only `[dev-dependencies]`).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -138,7 +138,6 @@ async-trait = { workspace = true }
 base64 = { workspace = true }
 bytes = { workspace = true }
 chrono = { workspace = true, features = ["clock"] }
-env_logger = { workspace = true }
 event-listener = { workspace = true }
 futures = { workspace = true, features = ["std"] }
 hex = { workspace = true }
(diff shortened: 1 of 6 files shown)


jlucaso1:
@claude review

PR #895, opened 2026-06-18
test(binary): add decoder roundtrip property tests and unmarshal fuzz target

jlucaso1:
## What
Adds the first property test and fuzz target for the `wacore-binary` crate
(the WhatsApp WABinary wire format).

- **Roundtrip proptest** (`wacore/binary/tests/roundtrip_proptest.rs`): a
  `proptest` strategy builds depth-bounded random `Node` trees (depth ≤4,
  ≤4 children/attrs, ≤64 content bytes) and asserts
  `unmarshal(marshal(node)) == node` over 2048 cases. Adds `proptest` to
  `[dev-dependencies]`.
- **Fuzz target** (`wacore/binary/fuzz/`): a `cargo-fuzz` / libFuzzer target
  that feeds arbitrary `&[u8]` into `unmarshal_ref` — the largest
  hostile-input attack surface — and discards the `Result`.

## Why
There was previously no property testing and no fuzzing anywhere in the
workspace. The marshal/unmarshal path is security- and correctness-critical
(it parses bytes straight off the wire), so it deserves both randomized
roundtrip coverage and a panic/overflow harness.

The encoder canonicalizes losslessly (token compression, nibble/hex packing,
JID packing), but two transforms are lossy at the *type* level, so the
generator is constrained to avoid them rather than weakening the assertion
(documented inline):
- `@`-bearing strings ≤48 bytes are parsed as JIDs (ambiguous), so no
  generated string contains `@`;
- non-token, non-packable `String` *content* re-decodes as `Bytes`, so
  `String` content is limited to the non-empty nibble alphabet, while
  arbitrary payloads go through the always-lossless `Bytes` variant.

The fuzz crate has its own `[workspace]` table so it stays out of the main
cargo workspace and cannot affect `cargo build --workspace`, clippy, or the
binary-size CI. It is intentionally **not** wired into gating CI (it needs
nightly + libFuzzer + time).

## Verification
- `cargo fmt --all --check` — clean
- `cargo build --workspace --exclude e2e-tests --all-targets` — ok
- `cargo build --workspace --exclude e2e-tests --all-features --all-targets` — ok
- `cargo clippy --workspace --exclude e2e-tests --all-features --all-targets -- -D warnings` — clean
- `cargo test -p wacore-binary` — pass
- `cargo test -p wacore-binary --no-default-features --lib` — pass (roundtrip proptest also verified to pass under `--no-default-features`, exercising the non-SIMD decode path)
- `cargo metadata` confirms `wacore-binary-fuzz` is **not** a member of the main workspace
- The fuzz crate and target build/link cleanly via `cargo build --manifest-path wacore/binary/fuzz/Cargo.toml`

## Running the fuzzer
`cargo +nightly fuzz run unmarshal_ref` (from `wacore/binary/`, with `cargo install cargo-fuzz`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ __pycache__
 .codex
 dhat-heap*.json
 /size-out
+# proptest persists failure seeds next to the test on failure; these are
+# environment-local replay caches, not source.
+*.proptest-regressions
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -224,6 +224,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "bit-set"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08807e080ed7f9d5433fa9b275196cfc35414f66a0c79d864dc51a0d825231a3"
+dependencies = [
+ "bit-vec",
+]
+
+[[package]]
+name = "bit-vec"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
+
+[[package]]
 name = "bitflags"
 version = "2.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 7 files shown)


jlucaso1:
@coderabbitai review

PR #918, opened 2026-06-23
VoIP 1:1 calling

jlucaso1:
WhatsApp-rust can now place and answer real, end-to-end encrypted **1:1 voice calls**, interoperating with the actual WhatsApp app on the other end. As far as I know this is the first Rust WhatsApp library to carry the full call **media** path, not just the signaling: the proprietary audio codec, the SRTP/SFrame crypto, the relay transport, and a portable engine that drives all of it.

## What works today
- Incoming and outgoing 1:1 audio calls, against the official WhatsApp app.
- The complete media plane: mic capture, encode, encrypt, relay, decrypt, decode, playout.
- Multi-device reality: a companion device answering is handled (recv-key rekey to the device that actually picks up), plus sibling dismiss, offline missed-call surfacing, and clean teardown.

## The parts I'm proud of
- **A pure-Rust MLow codec.** WhatsApp's voice codec ("MLow", a heavily modified Opus) is hand-ported to Rust from the reverse-engineered reference, byte for byte, pinned by a golden roundtrip test. There is no codec FFI: the encoder and decoder are plain Rust, so they compile to wasm and to embedded targets like everything else.
- **End-to-end encryption, the WhatsApp way.** The call key arrives over the Signal session; from it the library derives the E2E-SRTP keys (HKDF plus the libsrtp AES-CM KDF), protects and unprotects audio with AES-128-CTR and the WARP integrity tag, and handles the SFrame layer. The relay never sees plaintext audio.
- **Sans-IO, bring-your-own-everything.** The `CallEngine` owns no socket, clock, or thread. You feed it relay packets, mic frames, and timeouts; it emits transmits, playout PCM, call events, and the next timeout. A portable `run_call` driver bridges it to any async runtime and transport. That is exactly why the same core runs on Tokio (servers), in the browser (wasm), and on an esp32 (a phone call from a microcontroller).

## Using it
```rust
// Answer an incoming call (you supply the mic source + speaker sink):
let handle = client.voip()
    .accept(&incoming)
    .audio(mic_source, speaker_sink)
    .start()
    .await?;

// Or place one:
let handle = client.voip()
    .call(peer)
    .audio(mic_source, speaker_sink)
    .start()
    .await?;

handle.wait_ended().await;   // resolves when either side hangs up
// handle.hangup().await;    // or end it yourself
```
You bring the audio I/O (any mic source and speaker sink); the bundled example wires up cpal/PipeWire. The library owns the call key, the relay handshake, the crypto, and the codec.

## Scope and safety
- Entirely behind the optional `voip` feature. The default build, and anyone who does not enable voip, pulls none of the webrtc/opus dependencies and is byte-for-byte unaffected.
- ~57k lines, the large majority of it the codec and its test vectors.
- The decoder is intentionally hardened to a single operating point (16 kHz wideband, 60 ms) with fail-loud guards that drop anything off-point rather than desyncing, so a malformed or out-of-spec frame degrades gracefully.

## Validation
- Live-tested end to end against the real WhatsApp app over 3G/WiFi/5G, audio clean in both directions.
- Byte-exact golden roundtrip test for the codec, plus known-answer-test vectors for the crypto, plus an in-tree loopback DTLS/SCTP transport e2e test.

## Roadmap / follow-ups
This 1:1 audio path is the foundation, not the finish line. Natural next steps:
- **Group calls.** The signaling and the SFrame/SRTP key management generalize to multi-party, and the engine is already participant-aware.
- **Video calls.** The transport (DTLS/SCTP/SRTP) and the relay path are codec-agnostic; video is a second media stream plus a VP8/H.264 path.
- **Deeper codec coverage.** Inband FEC, PLC, CNG, and the low-bitrate operating points (the guards already tell us, with data, when a real peer would need them).
- **Calling from embedded.** The sans-IO core already builds for esp32; a reference firmware demo that rings is the fun milestone.


Salientekill:
Eca, ta ruim, melhore kskskskkskskkskskksks

purpshell:
Please work on incorporating wacrg elements into the code as I have discussed with you personally. Let's keep track of that task here before merge.

purpshell:
Also, I request that the commits made on this branch not be squashed, rather use a merge commit to preserve changes.

Salientekill:
> Also, I request that the commits made on this branch not be squashed, rather use a merge commit to preserve changes.

🫪

datfooldive:
video call when?

jlucaso1:
> video call when?

@purpshell is cooking

Salientekill:
> video call when?

Finalmente irei conseguir fazer GF com o bot

PR #924, opened 2026-06-28
feat(retry): bound inbound retry-receipt rate per sender

Salientekill:
## Problem

A chronically undecryptable sender — a broken Signal session the peer never re-establishes, or a draining offline queue of messages that can no longer decrypt — fails **every** message it sends, and we emit a retry receipt for each one.

The per-message cap (`MAX_DECRYPT_RETRIES`) bounds receipts **per message id, not per sender**, so a sender flooding distinct ids is never gated and the aggregate outbound retry-receipt rate climbs into `AccountLocked`. This is the same anti-abuse class the per-chat resend limiter (#871) already bounds for outbound *resends*, but on the inbound decrypt-failure path.

Observed in production logs: **~25k retry receipts from ~15 senders over two days** (each sender failing 700–1300×), correlating with an `AccountLocked` incident.

## Solution

A per-sender token-bucket limiter (`retry_receipt_limiter`) mirroring `resend_rate_limiter`, keyed by sender. When a sender exhausts its budget in `handle_decrypt_failure`, **ack-and-drop** the stanza (send the transport ack to drain the offline queue so the server stops redelivering) instead of asking for a resend forever.

Key properties:
- **Self-healing:** the bucket refills lazily and any successful decrypt / fresh handshake recovers the sender. A transient desync (a handful of retries under the burst) is never penalized.
- **Non-invasive:** the `UndecryptableMessage` event still fires, so any app-side recovery (e.g. local session reset) is unaffected — only the *outbound retry receipt* is bounded.
- **Tunable live** via the public `Client` API, and drops surface on a counter for observability.

`TokenBucket` is now shared with `resend_rate_limiter` (made `pub(crate)`).

## API

- `Client::set_retry_receipt_rate_limit(burst, refill_per_min)` — retune live; `burst = 0` disables.
- `Client::retry_receipts_throttled_total()` — dropped-receipt counter.
- `MemoryDiagnostics`: `retry_receipt_limiter_senders`, `retry_receipts_throttled_total`.

Defaults: **burst 10, refill 2/min, capacity 4096 senders**.

## Testing

- Unit tests for the per-sender bucket (burst/refill, per-sender isolation, disabled, concurrency) mirroring the resend-limiter suite.
- A `Client`-level wiring test (`client_retry_receipt_limiter_is_wired_and_tunable`).
- Full lib suite green (`cargo test -p whatsapp-rust --lib`: 843 passed), `cargo clippy --tests` clean.

<a href="https://cubic.dev/pr/oxidezap/whatsapp-rust/pull/924?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/cache_config.rs
+++ b/src/cache_config.rs
@@ -208,6 +208,11 @@ pub struct CacheConfig {
     /// Default: 4096.
     pub resend_rate_limiter_capacity: u64,
 
+    /// Per-sender retry-receipt rate-limiter capacity: one token-bucket entry per
+    /// sender recently driving decrypt-failure retries. Same fail-open eviction
+    /// semantics as `resend_rate_limiter_capacity`. Default: 4096.
+    pub retry_receipt_limiter_capacity: u64,
+
     // --- Sent message DB cleanup ---
     /// TTL in seconds for sent messages in DB before periodic cleanup. Must
     /// outlive retry receipts (which can arrive well after a send) or the retry
@@ -283,6 +288,10 @@ impl std::fmt::Debug for CacheConfig {
                 "resend_rate_limiter_capacity",
                 &self.resend_rate_limiter_capacity,
             )
+            .field(
+                "retry_receipt_limiter_capacity",
+                &self.retry_receipt_limiter_capacity,
+            )
             .field("sent_message_ttl_secs", &self.sent_message_ttl_secs)
             .field("msg_secret_policy", &self.msg_secret_policy)
             .field("msg_secret_retention", &self.msg_secret_retention)
@@ -340,6 +349,7 @@ impl Default for CacheConfig {
             session_locks_capacity: 10_000,
             chat_lanes_capacity: 5_000,
             resend_rate_limiter_capacity: 4_096,
+            retry_receipt_limiter_capacity: 4_096,
             sent_message_ttl_secs: 7200,
             // Bounded by default: seed only the still-relevant slice of history
             // and prune by per-add-on-kind event-time horizons, so the store no
--- a/src/client.rs
+++ b/src/client.rs
@@ -199,6 +199,9 @@ pub struct MemoryDiagnostics {
     pub resend_rate_limiter_chats: u64,
     /// Total outbound resends dropped by the per-chat rate limiter since start.
     pub resends_throttled_total: u64,
+    pub retry_receipt_limiter_senders: u64,
+    /// Total inbound retry receipts dropped by the per-sender rate limiter since start.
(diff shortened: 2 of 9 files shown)


Salientekill:
Thanks — addressed the cubic P2: `MemoryDiagnostics`' `Display` impl in `src/client.rs` now formats the two new fields (`retry_receipt_limiter_senders`, `retry_receipts_throttled_total`), mirroring the resend-limiter lines. Verified with `cargo check --features debug-diagnostics` + `cargo fmt --check`. Amended into the existing commit.

Salientekill:
Addressed both CodeRabbit comments:

1. **Major (atomic snapshot):** `burst` + `refill_per_min` are now packed into a single `AtomicU64` (`pack_rate`/`unpack_rate`), so `set_rate` publishes both as one atomic snapshot and `try_acquire` reads a single load — no torn (new burst, old refill) pair on a concurrent retune.
2. **Minor (docs):** `handle_decrypt_failure`'s doc comment now documents the ack-and-drop exception (the limiter-exhausted path clears the stanza without a retry).

Validated: `cargo clippy --all-targets -- -D warnings` + `cargo fmt --check` clean, limiter tests green.

Salientekill:
Closing this in light of #985.

This PR bounded the *outbound* direction — the retry receipts we emit in `handle_decrypt_failure` when inbound messages fail to decrypt (the June AccountLocked incident: a cohort of broken senders flooding us with undecryptable messages → mass receipts → lock). That's a different class from the inbound group storm, so #985 doesn't technically supersede it.

But #985 established the principle that settles it either way: WA Web has no volume throttle on retry receipts in any direction, so the SDK shouldn't impose one by default — it belongs behind an opt-in operator hook, not baked into the core path. This PR is exactly a built-in, on-by-default throttle, so the same reasoning that superseded #982 applies here.

The outbound class it guarded is currently mitigated by the upstream per-chat cap plus our app-side `session_recovery` give-up tier (no AccountLocked since June). If the outbound storm resurfaces, the consistent fix is a `DecryptFailureAdmission` hook mirroring @jlucaso1's `RetryAdmission` in #985, owned by the operator — not this. Closing to keep the record clean; happy to propose that hook if the need returns.

PR #934, opened 2026-07-01
fix(receive): skip PDO placeholder-resend for view-once, ack instead

blaueeiner:
## Problem

A message that arrives with only an `<unavailable type="view_once">` child (no `<enc>`) currently triggers a PDO placeholder-resend to our own phone (`run_pdo_request`). But WhatsApp never fans view-once media out to companion/linked devices, so that request **always comes back empty** — the content is unrecoverable by design (the code already documents this expectation elsewhere).

Worse, the peer round-trip is surfaced by the phone as a spurious **"Finished syncing with WhatsApp on \<device\>"** notification — once for every view-once message the companion receives. This is a visible, repeated annoyance for anyone running a linked/companion client.

## Fix

In the `<unavailable>`-child branch of `handle_incoming_message`, when the type is `ViewOnce`:

- **Skip** `run_pdo_request` entirely (it can only return empty and only causes the notification).
- Still **dispatch** the `UndecryptableMessage` event, so consumers keep seeing the failure.
- Still **ack** the stanza directly. Previously the transport ack was gated on the PDO going out (`pdo_sent`); with the PDO skipped we ack unconditionally (outside status-broadcast) so the offline queue drains and the server does not redeliver the stanza.

`Unknown` unavailables keep the existing PDO recovery path completely unchanged — only view-once, which is provably unrecoverable, is short-circuited.

## Test

Adds `view_once_stub_acks_without_pdo`, which drives a bare view-once stub through `handle_incoming_message` on the capturing-transport harness and asserts both that the `ViewOnce` `UndecryptableMessage` still fires **and** that an `<ack class="message">` is sent. Also updates the now-stale comment on `test_unavailable_without_enc_dispatches_view_once_event` (it claimed view-once "relays via PDO").

## Validation

- `cargo test -p whatsapp-rust --lib` — new test passes; the full view-once/PDO/unavailable suite (24 tests) is green, no regressions.
- `cargo fmt --check` clean.
- `cargo clippy -p whatsapp-rust --lib -- -D warnings` clean.

Rebased on current `main`.

<a href="https://cubic.dev/pr/oxidezap/whatsapp-rust/pull/934?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/message/receive.rs
+++ b/src/message/receive.rs
@@ -93,11 +93,31 @@ impl Client {
                 Some("view_once") => crate::types::events::UnavailableType::ViewOnce,
                 _ => crate::types::events::UnavailableType::Unknown,
             };
-            log::info!(
-                "[msg:{}] Message has <unavailable> child (type: {:?}), requesting from phone via PDO",
-                info.id,
-                unavailable_type
+            // View-once media is never fanned out to companion/linked devices,
+            // so a PDO placeholder-resend to our own phone always comes back
+            // empty — the content is unrecoverable by design. Worse, that peer
+            // round-trip is surfaced by the phone as a spurious "Finished
+            // syncing with WhatsApp on <device>" notification for every
+            // view-once message received. Skip the PDO for view-once; still
+            // dispatch the event (so consumers see the failure) and still ack
+            // so the offline queue is cleared and the stanza is not redelivered.
+            let is_view_once = matches!(
+                unavailable_type,
+                crate::types::events::UnavailableType::ViewOnce
             );
+            if is_view_once {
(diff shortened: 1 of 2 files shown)


jlucaso1:
Thanks @blaueeiner <3

PR #940, opened 2026-07-02
feat(client): opt-in force_pn_addressing to keep outbound DMs on PN JIDs

juanlotito:
## Summary

Adds an opt-in `Client::set_force_pn_addressing(true)` that keeps outbound DM addressing on the namespace the caller provided instead of upgrading PN → LID. The flag is scoped to the four addressing decisions the outbound path makes — recipient fanout keys (`send_message_impl`), session creation (`resolve_lid_mappings` via `ensure_e2e_sessions`), stanza prep (`SendContextResolver::get_lid_for_phone`) and session lock keys (`build_session_lock_keys`, kept aligned with the namespace encryption actually uses so the per-session lock invariant against concurrent inbound decrypt holds) — while inbound decrypt addressing (`resolve_encryption_jid` on sender JIDs), LID↔PN mapping learning, and inbound session migration stay untouched. Mirrors the existing `skip_history_sync` runtime flag pattern.

## Motivation

On at least one companion registration (personal account, client linked as a companion device), DMs sent with LID addressing are accepted by the server but never delivered: no `Delivered` receipt ever arrives and the recipient never sees the message. The same message sent with pre-0.6 PN addressing is delivered within ~2 seconds (verified side by side against 0.5.0).

With 0.6's unified addressing there is currently no way to work around this from the caller side:

- passing a PN JID to `send_message` doesn't help, because the DM path re-upgrades it internally as soon as the LID mapping is in the cache, and
- the first DM to an unmapped PN triggers the usync LID query that learns and persists the mapping — so the *first* message to a fresh PN is delivered and every subsequent one is silently dropped, which made the failure mode quite confusing to diagnose.

Timeline observed live (numbers redacted):

1. fresh client start → `send_message(pn_jid)` → delivered ✔
2. inbound message arrives from that user (mapping learned/warmed)
3. `send_message(pn_jid)` → accepted by server, never delivered, no receipt ✘
4. same on every retry; rolling back to 0.5.0 (PN addressing) restores delivery immediately, receipts arrive in ~2s again ✔

I don't know how widespread this server behavior is (possibly related to account type or registration age), so an opt-in escape hatch seemed like the minimal-surface fix rather than changing default behavior. Happy to rework into a `BotBuilder` option or config field if you prefer.

## Why the flag gates four call sites instead of just `resolve_encryption_jid`

A first iteration gated only `resolve_encryption_jid`. That turned out to be both insufficient — `prepare_dm_stanza`/`encrypt_for_devices` upgrade PN → LID through a separate path (`SendContextResolver::get_lid_for_phone`), so participants and enc nodes still went out LID-addressed — and unsafe: session lock keys would have diverged from the namespace encryption actually used, breaking the documented per-session lock invariant (concurrent outbound encrypt and inbound decrypt of the same peer would no longer serialize on the same mutex). The final shape keeps every outbound addressing decision consistent on PN while leaving the inbound half of the pipeline stock.

## Validation

- `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings` (0 warnings), `cargo test --workspace --exclude e2e-tests` (all suites green) — same commands as the `Rust CI` workflow.
- New unit test `test_force_pn_addressing_scopes_to_outbound_paths` asserting the outbound paths keep PN with the flag on, that inbound `resolve_encryption_jid` is deliberately NOT gated, and that turning the flag off restores stock behavior.
- Live validation on the affected deployment: with the flag, repeated DMs (warm mapping cache — the previously failing condition) all produced `Delivered` receipts within ~2s, including full inbound → reply round-trips; without it, silent drops as described above.

Default behavior is unchanged — the flag is off unless a caller opts in.

<a href="https://cubic.dev/pr/oxidezap/whatsapp-rust/pull/940?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -503,6 +503,14 @@ pub struct Client {
     /// or processed. Set via `BotBuilder::skip_history_sync()`.
     pub(crate) skip_history_sync: AtomicBool,
 
+    /// When true, outbound DM addressing (device fanout, session creation,
+    /// stanza prep and session locks) keeps the PN namespace instead of
+    /// upgrading to LID. Inbound decrypt addressing is not affected. Escape
+    /// hatch for companion registrations whose LID-addressed DMs are accepted
+    /// by the server but never delivered (no Delivered receipt); PN
+    /// addressing matches pre-0.6 behavior.
+    pub(crate) force_pn_addressing: AtomicBool,
+
     /// Cache configuration for TTL and capacity of all caches.
     /// Stored for use by lazily-initialized caches (group_cache).
     pub(crate) cache_config: CacheConfig,
@@ -626,6 +634,18 @@ impl Client {
     /// notifications but will not download or process the data.
     pub fn set_skip_history_sync(&self, enabled: bool) {
         self.skip_history_sync.store(enabled, Ordering::Relaxed);
+    }
+
+    /// Keep DM addressing on phone-number JIDs even when a LID mapping is
+    /// known. Some companion registrations observe LID-addressed DMs being
+    /// accepted but never delivered; this restores pre-0.6 PN addressing.
+    pub fn set_force_pn_addressing(&self, enabled: bool) {
+        self.force_pn_addressing.store(enabled, Ordering::Relaxed);
+    }
+
+    /// Returns `true` if DM addressing is pinned to phone-number JIDs.
+    pub fn force_pn_addressing_enabled(&self) -> bool {
+        self.force_pn_addressing.load(Ordering::Relaxed)
     }
 
     /// Override `DeviceProps` fields before the initial pairing. Only fields
@@ -841,6 +861,7 @@ impl Client {
             http_client,
             override_version,
             skip_history_sync: AtomicBool::new(false),
+            force_pn_addressing: AtomicBool::new(false),
             cache_config,
             self_weak: std::sync::OnceLock::new(),
             saver_handle: std::sync::OnceLock::new(),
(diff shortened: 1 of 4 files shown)


juanlotito:
**Update — new evidence, and rebase incoming.**

I cherry-picked #731 onto v0.6.0 plus this branch's flag to isolate variables, and re-tested on the affected companion registration with the ack logging #731 added:

- **Stock consistent-LID addressing (post-#731 stanza shape): the server rejects the DM with `ack error="400"`** — reproducible on every send, message never delivered. So on this registration the drops are not (only) the pre-#731 mixed PN-to/LID-participants bug: even a fully LID-consistent stanza is 400-nacked.
- **With `force_pn_addressing` on (same build, same peer): delivered within ~2s** (Delivered receipts), repeatedly.
- Side observation: in PN mode each send also draws a `479 (SmaxInvalid)` nack (reporting token?) while still delivering — happy to file separately if useful.

This PR currently conflicts with `main` after the send-path restructuring; I'll rebase and reimplement the flag against current `main` shortly. Given the 400s above, an opt-in PN addressing escape hatch still seems necessary for registrations like this one — but if you'd rather approach it differently (e.g. automatic fallback to PN on 400-nack), happy to rework in that direction.

PR #972, opened 2026-07-03
feat(bot): configurable initial app-state key-share wait

gergesh:
## What

Makes the initial critical app-state sync's wait for the app-state key-share configurable, defaulting to the current hard-coded 5s.

- `BotBuilder::with_app_state_key_wait(Duration)`
- `Client::set_app_state_key_wait(Duration)` / `Client::app_state_key_wait()`

## Why

On a fresh pairing, `run_initial_sync` (node_io.rs) waits up to a hard-coded `Duration::from_secs(5)` for `initial_keys_synced_notifier` before it proceeds to fetch the critical (`critical_unblock_low` etc.) app-state snapshot. The key-share is an encrypted peer message; when a large concurrent history sync is saturating the stream at pairing time, it can arrive *after* that 5s window. The snapshot is then fetched before the key is stored, and critical sync fails with `didn't find app state key` (and the follow-up `patch snapshot MAC mismatch`). For a companion client that reconstructs the address book from `critical_unblock_low`, that means the account's **saved contact names never sync**.

`listen()` still returns the instant the keys arrive, so a larger value only extends the worst case (no key-share) — it doesn't slow down a healthy pairing. A consumer with a deliberately deep history-sync window can raise this to, say, 30s to let the key-share win the race.

## How

Stored as an `AtomicU64` (ms) on `Client`, initialized to `5_000`, read where the fixed `5s` was used. Wired through the builder exactly like the existing `wanted_pre_key_count` / `resend_rate_limit` tuning. Default is unchanged, so this is fully backward-compatible.

## Tests

- `test_bot_builder_app_state_key_wait` — override applies to the client
- `test_bot_builder_default_app_state_key_wait` — default is 5s

`cargo fmt --all`, `cargo clippy -p whatsapp-rust --lib --tests`, and the new tests all pass.

### DIFF
--- a/src/bot.rs
+++ b/src/bot.rs
@@ -15,6 +15,7 @@ use std::future::Future;
 use std::marker::PhantomData;
 use std::pin::Pin;
 use std::sync::Arc;
+use std::time::Duration;
 use thiserror::Error;
 use wacore::proto_helpers::MessageBuilderExt;
 use wacore::runtime::Runtime;
@@ -565,6 +566,7 @@ pub struct BotBuilder<
     cache_config: CacheConfig,
     wanted_pre_key_count: Option<usize>,
     resend_rate_limit: Option<(u32, u32)>,
+    app_state_key_wait: Option<Duration>,
     task_instrument: Option<Arc<dyn wacore::stats::TaskInstrument>>,
     alloc_meter: Option<Arc<wacore::stats::AllocMeter>>,
     _marker: PhantomData<(B, T, H, R)>,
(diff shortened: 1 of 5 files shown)


jlucaso1:
Hi @gergesh I'll investigate that, but I'm unable to reproduce this locally. Can you provide logs with debug mode and with PII? Also what version or commit hash of the library are you trying?

jlucaso1:
Also, can you check if this another solution works for you? https://github.com/oxidezap/whatsapp-rust/pull/974

PR #982, opened 2026-07-04
feat(retry): quarantine per-(chat, requester) inbound retry receipts

Salientekill:
## Problem

In large groups, a cohort of members whose pairwise sessions never establish (dead registrations, exhausted prekeys, re-registered LIDs) sends a retry receipt for **every single group message**. Each receipt pays `markForgetSenderKey` (a DB write + a whole-group sender-key cache invalidation), bundle processing and possibly a resend — all **upstream** of the per-chat resend cap (#871), which only bounds the resend itself.

Production evidence (bot in a 1012-participant group, 3.5 days of debug logs):
- **58,554** `Marked <jid> for fresh SKDM ... due to retry receipt` lines — 99.7% from this one group, coming from **468 distinct members** (~68 marks/member/day ≈ one per bot message: they fail on *every* message, i.e. their sessions never repair).
- The per-chat cap blocked ~75% of the resends (23k+19k+1k throttled) — but none of the mark/bundle work, which runs before the cap check in `handle_retry_receipt`.
- Related root cause on the send side: after the server ACK the whole SKDM target set is marked `has_key=true`, including devices whose distribution failed (406 / no bundle) — `wacore/src/send/group.rs` — relying on the retry-receipt path to repair, which produces a permanent warm/cold oscillation for this cohort. (A follow-up PR could add a per-device re-target cooldown; this PR bounds the receipt side.)

## Mechanism

`RetryMarkQuarantine`: a token bucket keyed by **(chat user, requester user)** — burst **2**, refill **2/day**. One mark is enough to repair a healthy member (the next send carries the SKDM), so past the burst, further receipts from the same pair are dropped **before** any repair work; the refill keeps genuine recovery possible. Device is excluded from the key on purpose: WA Web re-targets the whole user when the primary goes cold, so all devices of a broken account share one budget.

With the observed numbers this cuts the repair-path work from ~16.7k/day to ~470×3/day worst case (~97%), while leaving first-time repairs untouched.

## Implementation

- Reuses the existing `TokenBucket` + capacity-only `Cache` machinery from the resend limiter (lazy monotonic refill, no timers, allocation only on miss).
- Gate placed in `handle_retry_receipt` right before `update_local_signal_session`, after the cheap early-outs and the concurrency guard; early return releases the pending-retry scopeguard normally.
- `burst = 0` disables; live-tunable via `Client::set_retry_mark_quarantine(burst, refill_per_day)`.
- Observability: `StatsSnapshot.retry_receipts_quarantined` + `retry_mark_quarantine_pairs` in `MemoryDiagnostics`.
- Unit test covers per-pair bounding, pair isolation and the disable knob.

Relation to #924: that PR bounds **outbound** retry receipts (our own decrypt failures); this one bounds the **inbound** direction, which is where the group-storm cost actually lands.

### DIFF
--- a/src/cache_config.rs
+++ b/src/cache_config.rs
@@ -213,6 +213,15 @@ pub struct CacheConfig {
     /// Default: 4096.
     pub resend_rate_limiter_capacity: u64,
 
+    /// Max (chat, requester) pairs tracked by the inbound retry-receipt
+    /// quarantine. Unlike the resend limiter (one entry per chat), the
+    /// quarantine keyspace is O(groups x broken members) — a single storming
+    /// 1012-participant group holds ~470 pairs — and evicting an ACTIVE pair
+    /// refunds its full burst on the next receipt, weakening the budget. Pairs
+    /// are tiny (two short strings + one bucket), so size generously.
+    /// Default: 32768.
+    pub retry_mark_quarantine_capacity: u64,
+
     // --- Sent message DB cleanup ---
     /// TTL in seconds for sent messages in DB before periodic cleanup. Must
     /// outlive retry receipts (which can arrive well after a send) or the retry
@@ -292,6 +301,10 @@ impl std::fmt::Debug for CacheConfig {
                 "resend_rate_limiter_capacity",
                 &self.resend_rate_limiter_capacity,
             )
+            .field(
+                "retry_mark_quarantine_capacity",
+                &self.retry_mark_quarantine_capacity,
+            )
             .field("sent_message_ttl_secs", &self.sent_message_ttl_secs)
             .field("msg_secret_policy", &self.msg_secret_policy)
             .field("msg_secret_retention", &self.msg_secret_retention)
@@ -350,6 +363,7 @@ impl Default for CacheConfig {
             chat_lanes_capacity: 5_000,
             group_distribution_locks_capacity: 512,
             resend_rate_limiter_capacity: 4_096,
+            retry_mark_quarantine_capacity: 32_768,
             sent_message_ttl_secs: 7200,
             // Bounded by default: seed only the still-relevant slice of history
             // and prune by per-add-on-kind event-time horizons, so the store no
--- a/src/client.rs
+++ b/src/client.rs
@@ -213,6 +213,7 @@ pub struct MemoryReport {
     pub session_locks: u64,
     pub chat_lanes: u64,
     pub resend_rate_limiter_chats: u64,
+    pub retry_mark_quarantine_pairs: u64,
     // -- Unbounded collections --
     pub response_waiters: usize,
     pub node_waiters: usize,
@@ -286,6 +287,11 @@ impl std::fmt::Display for MemoryReport {
             f,
             "  resend_rl_chats:        {}",
             self.resend_rate_limiter_chats
+        )?;
+        writeln!(
+            f,
+            "  retry_quarantine_pairs: {}",
+            self.retry_mark_quarantine_pairs
         )?;
         writeln!(f, "--- Unbounded collections ---")?;
         writeln!(f, "  response_waiters:       {}", self.response_waiters)?;
@@ -612,6 +618,7 @@ pub struct Client {
     /// to a chat (the anti-abuse signal) so a PN to LID fan-out cannot storm into
     /// AccountLocked. Throttled devices still recover via the fresh-SKDM mark.
     pub(crate) resend_rate_limiter: crate::resend_rate_limiter::ResendRateLimiter,
+    pub(crate) retry_mark_quarantine: crate::resend_rate_limiter::RetryMarkQuarantine,
 
     /// Dispatch-once gate for `UndecryptableMessage`: a server resend of a
     /// failed id re-enters the failure path and would otherwise fire a
--- a/src/client/accessors.rs
+++ b/src/client/accessors.rs
@@ -83,6 +83,12 @@ impl Client {
         self.resend_rate_limiter.set_rate(burst, refill_per_min);
     }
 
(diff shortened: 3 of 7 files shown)


Salientekill:
Addressed the review in the latest commit:

- **Capacity (P1)**: added a dedicated `cache_config.retry_mark_quarantine_capacity` (default 32768). Agreed the keyspace is O(groups × broken members) and that evicting an *active* pair refunds its burst — the doc on the field now states exactly that.
- **rotateKey before the gate (P2)**: the gate now sits right after the message-cache lookup and **before** the group-info fetch and the rotateKey block, so a quarantined pair no longer pays `query_info` nor the unknown-sender rotateKey (own sender-key deletion + whole-group invalidation) at full rate. Kept it after the cache lookup so receipts for expired messages (cheap no-op today) don't burn a pair's repair budget.
- **Key allocation (P2)**: documented the tradeoff at the call site — two small strings once per retry receipt (not per message), and when it quarantines it replaces a DB write + whole-group cache invalidation. No `Borrow<(&str, &str)>` for `(String, String)` makes a by-ref lookup impractical without a custom key type; happy to revisit if you prefer one.

Salientekill:
Second round — thanks @coderabbitai and @cubic-dev-ai:

- **Concurrency test (coderabbit `resend_rate_limiter.rs:256`, cubic P3)**: added `quarantine_concurrent_acquires_for_one_pair_do_not_exceed_burst` — 40 concurrent receipts for the same pair, asserts exactly `burst` pass and the rest are throttled, mirroring the existing `ResendRateLimiter` contention test. Agreed this needs to be proven, not assumed.
- **Capacity (cubic P2)**: same fix as greptile's P1 — dedicated `retry_mark_quarantine_capacity` (default 32768) landed in the previous commit.
- **Key allocation (cubic P2)**: kept as documented rather than resolved, and here's the reasoning I'd like a maintainer to weigh in on. This runs once per *retry receipt*, not per message. Even at the pathological volume that motivated the PR (58k marks / 3.5 days) that's ~0.19 receipts/s → two small-string allocations at that rate, and only on the path that *replaces* a DB write + whole-group cache invalidation. A borrowed-key lookup isn't drop-in (`(String,String): !Borrow<(&str,&str)>`); a custom key newtype would add real complexity for allocator traffic that's orders of magnitude below the hot path. Happy to add the newtype if you'd prefer the consistency with `ResendRateLimiter`, but my read is it's not worth it here. Open to your call.

Salientekill:
Good catch @greptile-apps — that was a real regression. Fixed in 32056d63: added `&& !is_peer` to the gate so our own companion devices are exempt from the quarantine. A secondary of our own account that was offline or rotated keys can legitimately need many repair cycles to rebuild its group session; the storm this guards is third-party members, never our own devices. `is_peer` was already computed just above the gate, so it's a one-line exemption.

jlucaso1:
Thanks for the very thorough work here @Salientekill, and especially for the production data. The diagnosis is correct and the engineering (token bucket reuse, the contention test, the `is_peer` exemption, the observability) is clean.

I want to be transparent about where I landed. I cross-checked the whole retry path against the captured WA Web bundle, and the blocker for me is parity rather than code quality:

- **WA Web has no volume-based throttle on inbound retry receipts.** `handleRetryRequest` serializes per chat, refuses past `MAX_RETRY`, and otherwise processes every one. Its only gates are semantic (`getMsgIfAuthorized` → `isRetryEligible`: `ALREADY_DELIVERED`, `CHANGED_IDENTITY`, `DEVICE_NOT_RECIPIENT`, `HIGH_RETRY_COUNT`, …), never "this member is asking too often."
- The send side is faithful too (`markHasSenderKey` marks the whole `skDistribList`, failed devices included), so the "broken member retries every message" loop and the per-receipt session-rebuild cost are both WA Web's own design. WA Web only survives it because a single-user client never sends at bot volume.

So a per-(chat, requester) quarantine that drops eligible repair requests is a deliberate divergence from WA Web. That's a perfectly reasonable trade-off for a bot at your scale, but not something I want the library doing for everyone by default.

Rather than reject the idea, I opened **#985** as a compliant, decoupled version: an opt-in `RetryAdmission` hook (mirroring the existing `InboundDurabilityHook` pattern) with **zero overhead when unused** (one `OnceLock::get` on the receive path). Your quarantine keeps everything that made it good — the token bucket, burst/refill tuning, the `is_peer` exemption, your own counters — and moves into operator code. I ported it as `examples/retry_quarantine.rs` so nothing is lost and the next person with a 1000-member group has a ready blueprint. The core stays byte-for-byte WA Web on the default path, and the gate sits exactly where yours did (before `markForgetSenderKey`, `rotateKey`, key-bundle processing and the resend).

I also folded in two things from reviewing your PR: the per-receipt `Marked … for fresh SKDM` log is demoted to `debug` (that's the source of your 58k lines), and I noted two compliant follow-ups in #985 — an in-place `sender_key_device_cache` update (WA Web mutates the participant record per device instead of invalidating the whole group), and closing the `isRetryEligible` parity gap.

Would you be up for trying #985 in your setup? Register `RetryQuarantine` from the example via `client.set_retry_admission(...)` with your burst 2 / refill 2/day, and confirm it bounds the storm the way your version did. If it does, I'll get it merged and credit this PR. Really appreciate you pushing on this — the fix wouldn't have been obvious without your diagnosis.

---

Salientekill:
Superseded by #985 — @jlucaso1's opt-in `RetryAdmission` hook is the better shape: it keeps the SDK default byte-for-byte WA Web while letting the operator own the exact quarantine mechanism, and it adds the in-place sender-key-device cache forget that attacks the per-mark cost my quarantine left on the table. I've wired my mechanism onto his hook on our side and it builds/tests clean; will report production numbers on #985. Closing in favor of it — thanks all for the reviews here.

PR #985, opened 2026-07-05
feat(retry): opt-in RetryAdmission hook (compliant supersede of #982)

jlucaso1:
## Summary

A WA-Web-compliant, decoupled alternative to the inbound retry-receipt quarantine in #982 (thanks @Salientekill for the diagnosis and production data that made this possible), plus the compliant efficiency win that helps every send by default.

Two things:
1. An opt-in `RetryAdmission` hook so a quarantine can live in operator code, keeping the SDK's default path byte-for-byte WA Web.
2. `markForgetSenderKey` now updates the sender-key-device cache in place, per device, matching WA Web, instead of invalidating the whole group. A generation stamp keeps the warm-send memo correct without any cross-cache invalidation.

## Why not land #982 as-is

Cross-checked against the captured WA Web bundle:

- **WA Web has no volume-based throttle on inbound retry receipts.** `WAWebHandleRetryRequest` serializes per chat (`onMessageQueue` + `sendMsgQueueMap`), refuses past `MAX_RETRY`, and otherwise processes **every** receipt. Its only gates are *semantic* (`getMsgIfAuthorized` → `isRetryEligible`: `ALREADY_DELIVERED`, `CHANGED_IDENTITY`, `RECORD_MISSING`, `DEVICE_NOT_RECIPIENT`, `HIGH_RETRY_COUNT`, `MESSAGE_EXPIRED`), never "this member is asking too often."
- The send side is faithful too: `GroupSkmsgJob` marks `markHasSenderKey(x, M)` over the whole `skDistribList` (failed devices included). So the "broken member retries every message" loop is WA Web's own design, and the per-receipt session rebuild is work WA Web does too. WA Web survives it only because a single-user client never sends at bot volume.

A per-(chat, requester) quarantine that drops eligible repair requests is a deliberate divergence, fine for a bot operator but not a default the library should impose.

## 1. `RetryAdmission` hook (opt-in, zero overhead when unused)

- **Trait** (`src/types/retry_admission.rs`) mirroring the existing `InboundDurabilityHook` idiom (`OnceLock`, builder-less `Client::set_retry_admission`). Object-safe, WASM-safe.
- **Synchronous** `admit(chat, requester, retry_count) -> bool`: a gate never needs to wait, and a sync verdict cannot stall retry processing while the `pending_retries` scopeguard is held.
- **Gate in `handle_retry_receipt`**, placed exactly where #982 put its gate (after `is_peer`, before the group-info fetch, the unknown-device `rotateKey` block, `markForgetSenderKey`, key-bundle processing, and the resend). Own companion devices (`is_peer`) and all DMs never reach the policy.
- **Zero overhead when unused:** with no policy registered the check is a single `OnceLock::get()`.
- **Log demotion:** the per-receipt `Marked … for fresh SKDM` line drops from `info` to `debug` (a broken cohort emits tens of thousands/day; WA Web logs it at its verbose level).
- **`examples/retry_quarantine.rs`** — a token-bucket quarantine keyed by (chat, requester), burst 2 / refill 2/day, built on the hook. This is @Salientekill's mechanism from #982, now operator-owned and tunable.

## 2. In-place sender-key-device cache forget (compliant perf win, default)

WA Web's `markForgetSenderKey` flips one device's flag in the participant record in place. Ours invalidated the whole-group `sender_key_device_cache`, forcing the next send to re-read and re-parse every device row from the DB (a 1000+ device map re-materialized per send during a storm).

- `SenderKeyDeviceMap` stores `has_key` as an `AtomicBool` and carries an `AtomicU64` `generation`. `SenderKeyDeviceCache::mark_forgotten` flips only the touched devices cold in a live cache entry (no invalidation, no re-read), and bumps the generation once — only when a flag actually transitions warm→cold (`swap(false)`), so a retry storm of duplicate cold marks never churns it.
- The warm-marking path still invalidates the group, since marking a device warm can introduce (user, device) pairs the cached map doesn't have yet.
- **The warm-send memo stays correct without a second invalidation.** `skdm_warm_memo` lets a warm repeat send skip the O(devices) `filter_skdm_targets` scan; it now stores the map's `generation` alongside the two `Weak` Arcs `(Weak<ResolvedGroupDevices>, Weak<SenderKeyDeviceMap>, u64)`. An in-place cold flip keeps the same `Arc`, so pointer identity alone can't see it — the generation can. The send path loads the generation *before* filtering (Acquire), so a flip racing in after that stamps the memo as already stale and the next send re-runs the filter. A single Acquire/Release generation pair carries the freshness (the same contract the device-registry generation gives membership), so there is no separate memo invalidation and therefore no cross-cache ordering window.
- The DB write stays the source of truth; a cache miss rebuilds from it.

## What stays out of the core (vs #982)

`retry_mark_quarantine_capacity`, the `StatsSnapshot` and `MemoryReport` fields, the lifecycle wiring, and any on-by-default behavior.

## Remaining follow-up (separate PR)

**Close the `isRetryEligible` parity gap** — WA Web tracks per-(message, recipient-device) delivery state and rejects `ALREADY_DELIVERED` / `CHANGED_IDENTITY`. We don't, so we can't. `handle_retry_receipt` now documents which reject reasons we already enforce (`HIGH_RETRY_COUNT`, `MESSAGE_EXPIRED` / `RECORD_MISSING`, `DEVICE_NOT_IN_DATABASE`) and which stay a known gap. Closing the rest needs a per-recipient receipt store (schema + hot-path writes per recipient), which is a substantial standalone feature and out of scope here.

## Tests

Both the happy and the failure/edge paths are covered:

- **Cache generation** — in-place flip advances the generation while keeping the same `Arc`; a duplicate-cold or absent-device mark must *not* advance it; a mixed batch flips every present device and bumps the generation exactly once; a cache miss is a no-op.
- **WA Web warm gate** — a device is warm only when it and its primary (device 0) both hold the key; a cold or absent primary makes the whole user cold.
- **Malformed input** — `from_db_rows` skips corrupt / partially-migrated JID rows without dropping the valid devices around them.
- **Own-device exclusion** — a cold mark that names our own device plus a broken member flips only the member (WA Web `!isMeDevice`), and is a full no-op (no DB write, no generation bump) when every named device is our own.
- **Quarantine example** — per-pair bounding and isolation, `burst = 0` disable, fail-open past capacity, gradual token refill (recovery, not a reset), and the idle clamp that stops an idle pair banking an unbounded reserve.

`cargo fmt --all --check` clean; `cargo clippy -p whatsapp-rust --tests --examples` clean; 942 lib tests pass; `examples/retry_quarantine.rs` builds and its 5 unit tests pass.

@Salientekill — would you try this in your 1000-member group? Register `RetryQuarantine` from the example via `client.set_retry_admission(...)` with burst 2 / refill 2/day and confirm it bounds the storm the way your version did.

### DIFF
--- a/examples/retry_quarantine.rs
+++ b/examples/retry_quarantine.rs
@@ -96,17 +96,16 @@ impl RetryQuarantine {
     }
 }
 
-#[whatsapp_rust::async_trait]
 impl RetryAdmission for RetryQuarantine {
-    async fn admit(&self, chat: &Jid, requester: &Jid, _retry_count: u8) -> bool {
+    fn admit(&self, chat: &Jid, requester: &Jid, _retry_count: u8) -> bool {
         if self.burst == 0.0 {
             return true;
         }
         let now = Instant::now();
         let key = (chat.user.to_string(), requester.user.to_string());
 
-        // std Mutex, never held across an await: the critical section is a map
-        // op plus the bucket arithmetic.
+        // Synchronous, so the receive path never awaits a policy: the critical
+        // section is a map lookup plus the bucket arithmetic.
         let mut buckets = self.buckets.lock().expect("quarantine mutex poisoned");
         // Fail open if we would exceed the cap with a brand-new pair, so the
         // guard never blocks repair for a pair it isn't already tracking.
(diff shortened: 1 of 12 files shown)


Salientekill:
Strong +1 on this direction — it's the right call, and better than #982 on two counts.

**On the opt-in vs default question**: you're right. I cross-read your WA Web bundle notes against my reasoning and agree the volume throttle is a deliberate divergence — correct for a bot operator to own, wrong for the SDK to impose. The `RetryAdmission` hook keeps the default path faithful while letting us keep the exact mechanism. Moving the token bucket into `examples/retry_quarantine.rs` (operator-owned, tunable) is the clean seam.

**On the in-place cache forget**: this is the part I under-solved. My quarantine cut the *frequency* of the marks (~97%), but every mark that still got through paid the whole-group `sender_key_device_cache` invalidation — re-materializing the 1000+ device map on the next send. Making `has_key` an `AtomicBool` and flipping one device cold in place attacks the *per-event* cost, which helps every send, not just during a storm. Together they cover both halves. `Ordering::Relaxed` is right here — it's a presence flag, the DB stays the source of truth, and the warm-marking path correctly still invalidates since it can add devices.

**Field data that backs the mechanism**: I shipped #982 to the 1012-member group before this landed. Over a 52-minute window: `Marked fresh SKDM` dropped from a ~11.6/min baseline to **0.54/min** (~95%), with **1,175 receipts quarantined** (97.7% of the 1,203 that reached the gate) and the per-chat resend cap firing **0 times** (it was blocking ~14k/day before). So the burst-2 / refill-2-per-day bucket you carried over is proven to bound the storm in production — I expect the same behavior through the hook, plus the extra per-send win from part 2.

Happy to close #982 in favor of this. I'll wire the hook into our bot, run it against that group, and report the numbers back here.

jlucaso1:
@coderabbitai review

PR #1024, opened 2026-07-13
VoIP 1:1 video calls

jlucaso1:
## Summary

Adds interoperable **1:1 video calls** to the existing VoIP implementation. Calls can start with video, upgrade from audio to video, return to audio, and accept the same transitions initiated by an official WhatsApp client.

The library remains codec-neutral: consumers provide and receive complete H.264 Annex-B access units through `VideoSource` / `VideoSink`. Encoding, decoding, capture, and display stay outside the core crates; the CLI example uses `ffmpeg` and `ffplay` without adding codec dependencies to the library.

This is additive to the existing 1:1 voice path. Bidirectional audio continues over its established pipeline while video is independently signaled, demultiplexed, encrypted, paced, and recovered.

## WhatsApp interoperability

The implementation covers the media and signaling profile verified against WhatsApp Web/Android captures and live Android calls:

- WhatsApp SSRC derivation, including the video participant slot and dynamic STUN `StreamDescriptors` for live audio/video SSRCs;
- PT-97 RTP with the captured extension layout, marker/timestamp behavior, WARP authentication, E2E-SRTP, and answering-device rekeying;
- directional/shared relay SRTCP derivation, native RTCP CNAME/SDES, and periodic Sender Reports at the advertised 1.5 s cadence using NTP wall time and actual sender/reception statistics;
- authenticated peer SR/RR, compact reports, NACK, PLI, and FIR parsing, including WhatsApp's feedback profile bit;
- per-sender SRTCP replay windows that reject duplicate/stale authenticated indices, accept bounded reordering and 31-bit wrap, and never commit state for forged packets;
- offer, preaccept, accept, typed video acknowledgements, orientation, upgrade/downgrade, and the standalone `Enabled` transition used to complete accepted upgrades.

The Android peer reports reception of the Rust participant's video SSRC and renders outbound video from both synthetic input and a real V4L2 webcam.

## Empirically confirmed webcam recovery

The webcam-only failure was isolated by replaying the exact captured webcam Annex-B stream and inspecting authenticated peer RTCP. The H.264 was valid and reached Android, but Android repeatedly sent PLI for the local video SSRC. Continuing to send dependent frames before the next IDR kept the official decoder outside a valid recovery point.

The engine now treats authenticated PLI/FIR as a decoder resynchronization boundary: dependent access units are withheld without consuming RTP/SRTP sequence state until the next IDR, then normal transmission resumes. The live V4L2 test changed from PLI every ~440 ms with no picture to one recovery request followed by continuous rendered webcam video.

The same invariant protects from-start video, reactivation, upgrade ungating, and relay backpressure. The recovery gate clears only after an IDR packetizes successfully. Disable purges queued, unstarted video access units while preserving any batch already on the transport, so reactivation resumes at a complete IDR without truncating an in-flight frame; audio and control traffic stay live.

## API and lifecycle

```rust
let handle = client
    .voip()
    .call(peer)
    .audio(mic, speaker)
    .video(h264_source, h264_sink)
    .start()
    .await?;

handle.start_video(new_source, new_sink).await?;
handle.stop_video().await?;
```

- `VideoSource` exposes its RTP timestamp stride, so 15 fps, 20 fps, 30 fps, and other valid 90 kHz cadences do not drift; zero stride is rejected at construction.
- Upgrade video remains send-gated until peer acceptance. Downgrade drains stale source access units without resetting RTP/SRTP state.
- Accepted peer upgrades become locally visible and ungate media only after the standalone `Enabled` send succeeds. A failed second handshake send emits phase-tagged diagnostics, tears down endpoints, and clears local video state in both initiator and acceptor paths, leaving the official peer to perform its normal incomplete-upgrade timeout.
- Incoming video-state events preserve routed `participant`/`recipient` metadata in typed acknowledgements, bind transitions to the call generation, and hold serialization through every post-ack effect and accepted-upgrade `Enabled` send. Unacknowledged or stale same-ID transitions cannot mutate a replacement call, and a committed state supersedes an older queued event under backpressure.
- Signaling state derives its wire markers from `VideoState`, preventing invalid marker/state combinations.
- Failed setup before a peer-visible transition rolls back endpoints and explicitly terminates an already-preaccepted call. Final accept is sent only after camera/ffmpeg has produced a decodable SPS/PPS/IDR.
- ffmpeg reader tasks are cancelled on readiness failure, timeout, source replacement, terminal call removal, and source drop. Video feeds retain source ownership for exactly the call lifetime.
- Peer termination during media preparation is tombstoned so a dead call cannot finish startup later.
- Video teardown hooks are generation-bound, consumed exactly once off-lock, and rearmed on every activation; stop/restart and all terminal paths release the correct endpoints without touching a same-call-id replacement.
- Video state and cadence controls use a lossless FIFO, so a `Disable` purge always precedes a later `Enable`; orientation is isolated in a one-slot latest-value mailbox and cannot evict negotiated transitions.

## H.264 media plane

- Single-NAL and FU-A packetization; single-NAL, STAP-A, and FU-A reassembly.
- SPS/PPS/IDR ordering compatible with the official receiver; encoder-only AUDs are omitted from RTP.
- Timestamp-boundary recovery when markers are lost, sequence-aware fragment-loss handling, allocation caps, forged-tag rejection, and immediate ordered draining when one packet completes multiple access units.
- Complete access units are the backpressure unit, so overload cannot leave half an IDR on the wire.
- From-start video, upgrades/downgrades, receiver rekey, RTP/RTCP wrap, malformed input, authenticated feedback, replay rejection, and keyframe recovery have deterministic tests/KATs.

## CLI example

Every call/listen mode accepts `--video`; `v` toggles video during a call and `q` performs a signaled hangup. Sources can be an OS webcam, a file/URL, or `WA_VIDEO_INPUT=testsrc`; sinks can be an `ffplay` window, raw `.h264`, or discard mode.

The example asynchronously classifies input paths, probes V4L2 capabilities, selects an exact supported capture mode, normalizes pixel format/geometry, and waits for a decodable SPS/PPS/IDR before accepting the call. Its default encoder contract matches the captured WhatsApp Web high-quality tier: H.264 Constrained Baseline Level 3.1, 1280×720 at 20 fps, about 1.98 Mbps, one slice per frame, repeated SPS/PPS, and a 60-frame GOP. `WA_VIDEO_SIZE`, `WA_VIDEO_FPS`, and `WA_VIDEO_BITRATE_KBPS` remain explicit overrides.

The preview uses arrival-time timestamps, bounded low-latency queues, and inverse `device_orientation` correction. This prevents long-run delay accumulation and correctly displays Android portrait video. Official clients still adapt quality dynamically; the example matches a captured high-quality tier but does not claim to implement WhatsApp's full bandwidth estimator.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --all --tests -- -D warnings`
- `cargo test --workspace --exclude e2e-tests`
  - `wacore`: 1,403 passed, 1 ignored
  - `whatsapp-rust`: 1,068 passed, 1 ignored
  - `whatsapp-rust-voip-cli`: 16 passed
- `cargo test -p wacore --features voip voip::`: 297 passed, 1 diagnostic ignored
- `cargo test -p whatsapp-rust --features voip voip::`: 53 passed
- `cargo test -p whatsapp-rust --features voip handlers::call::tests::`: 24 passed
- `cargo build -p whatsapp-rust-voip-cli --release`
- live Android interoperability: stable bidirectional audio, inbound video decode, outbound synthetic and V4L2 webcam video rendered on the phone, portrait correction on desktop, and sustained low-latency preview

`cargo test --all` additionally requires the documented E2E mock server; GitHub Actions supplies that environment.

## Scope

- 1:1 calls only; group video is not part of this PR.
- H.264 Annex-B transport only; pixel formats and codec implementations remain consumer-owned.
- VoIP feature scope only, with no new runtime codec dependency in `wacore` or `whatsapp-rust`.

### DIFF
--- a/examples/voip-cli/src/main.rs
+++ b/examples/voip-cli/src/main.rs
@@ -953,6 +953,7 @@ fn spawn_call_event_listener(
     tokio::spawn(async move {
         let mut opus = WaOpusDecoder::new().ok();
         let mut peer_video_receiver_confirmed = false;
+        let mut peer_keyframe_request_logged = false;
         while let Ok(ev) = events.recv().await {
             match ev {
                 CallEvent::RelayAllocated => {
@@ -1026,8 +1027,9 @@ fn spawn_call_event_listener(
                         );
                         peer_video_receiver_confirmed = true;
                     }
-                    if requests_keyframe {
+                    if requests_keyframe && !peer_keyframe_request_logged {
                         info!("🎥 peer requests an outbound keyframe: [{feedback}]");
+                        peer_keyframe_request_logged = true;
                     }
                     debug!(
                         "📊 peer RTCP PT={packet_types:?} sender={sender_ssrc:#010x} refs={referenced_ssrcs:x?} reports_audio={reports_audio} reports_video={reports_video} blocks=[{blocks}] feedback=[{feedback}]"
--- a/examples/voip-cli/src/main.rs
+++ b/examples/voip-cli/src/main.rs
@@ -32,7 +32,7 @@ use std::collections::{HashMap, HashSet};
 use std::sync::atomic::{AtomicBool, AtomicI32, Ordering};
 use std::sync::{Arc, Mutex};
 
-use anyhow::{Result, anyhow};
+use anyhow::{Result, anyhow, bail};
 use log::{debug, error, info, warn};
 use portable_atomic::AtomicU64;
 use wacore::stanza::call::{self as stanza, CAPABILITY_OFFER};
@@ -108,7 +108,7 @@ async fn main() -> Result<()> {
     }
     match command {
         CliCommand::Loopback { video: true } => {
-            video::run_video_loopback(&VideoOpts::from_env()?).await
+            video::run_video_loopback(&VideoOpts::from_env().await?).await
         }
         CliCommand::Loopback { video: false } => run_loopback().await,
         CliCommand::Listen { accept, video } => run_bot(Mode::Listen { accept, video }).await,
@@ -167,8 +167,10 @@ fn video_source_is_ignored(command: &CliCommand, video_input_is_set: bool) -> bo
         && matches!(
             command,
             CliCommand::Loopback { video: false }
-                | CliCommand::Listen { video: false, .. }
-                | CliCommand::Call { video: false, .. }
+                | CliCommand::Listen {
+                    accept: false,
+                    video: false,
+                }
         )
 }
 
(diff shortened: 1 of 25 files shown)


jlucaso1:
@codex review

nizarfadlan:
That's really interesting—I'll try testing the VoIP video feature.

PR #1047, opened 2026-07-16
ci: reclaim runner disk before tests

### LINKED ISSUE #1046: CI: reclaim unused Android SDK space before Rust tests
## Problem

The Build & Test job has started exhausting the ubuntu-latest disk before the test suite can run. This is reproducible outside the current code change:

- main run 29506617008 failed while linking discrepancy_pocs.
- PR run 29509725550 failed twice, first while compiling waproto and then while linking alloc_tracking.
- Every failure ends with No space left on device. The other Rust jobs pass.

The last successful run used Ubuntu image 20260705.232.1. Main and both failed attempts used the newer 20260714.240.1 image. Re-running alone is therefore not a reliable fix.

## Proposed fix

Remove /usr/local/lib/android at the start of Build & Test and print disk usage before and after. This job does not use the Android SDK, while the hosted image ships several SDK and NDK versions there. Keep the target cache, debug information, and test matrix unchanged.

This should restore several GB of headroom, keep the cache latency gains, and stop unrelated PRs from failing late in the job.

jlucaso1:
Closes #1046.

## Why

Build & Test started running out of disk on the Ubuntu 20260714.240.1 image. Main failed once and PR #1045 failed twice at different rustc targets, always with No space left on device. The previous image completed the same matrix.

## Change

Remove the unused hosted Android SDK at the start of the two disk-heavy Rust jobs and print disk usage before and after. This leaves the Rust target cache, debug information, commands, and coverage unchanged.

## Validation

- Cleanup increased free space from 89 GiB to 99 GiB in about 20 seconds.
- Build & Test passed in 10m17s; Build & Lint all-features passed in 8m41s.
- All 14 executable checks passed, including E2E, WASM, Binary Size, and CodSpeed.
- Binary size and LLVM-lines metrics are unchanged.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -65,6 +65,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
+      - name: Reclaim unused Android SDK space
+        run: |
+          df -h /
+          sudo rm -rf --one-file-system /usr/local/lib/android
+          df -h /
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: nightly-2026-06-16
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -109,6 +109,11 @@ jobs:
     # workspace recompile with voip on, not the deps. The heavy relay e2e tests stay out of `--lib`.
     steps:
       - uses: actions/checkout@v6
+      - name: Reclaim unused Android SDK space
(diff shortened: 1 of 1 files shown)


jlucaso1:
Addressed in 405d95f5. Build & Lint had remained green on the new image, but it is the other disk-heavy job and does not use Android either, so it now gets the same headroom without changing its build or test coverage.

PR #1051, opened 2026-07-17
fix(voip): prevent stale audio-to-video upgrades

jlucaso1:
## Summary

- model self and peer video directions independently, including the complete native video-state set
- serialize peer signaling, user actions, and the five-second upgrade timeout per call generation
- bind `accept_video` to an exact peer request token so cancelled or superseded upgrades cannot attach camera endpoints
- auto-accept video upgrades in the CLI even when the accepted call started as audio, while preserving the encoded-audio/Opus pipeline from #1050

## Root cause

Camera preparation waits for the first decodable IDR outside the signaling path. In the reported trace that preparation crossed the peer timeout, so the peer cancellation overtook the user action. The old code then completed the stale accept and retained `/dev/video0`, making the next upgrade fail with `Device or resource busy`.

The native APK/WASM behavior guards acceptance on the peer still being in Request state, serializes state transitions through a stream mutex, and expires local requests after five seconds. This change mirrors those invariants with a per-generation transition lock plus monotonic request epochs.

## Verification

- `cargo fmt --all`
- `cargo clippy --all --tests --features voip -- -D warnings`
- `cargo test --workspace --exclude e2e-tests`
- `cargo test --features voip voip::facade::tests -- --nocapture`
- `cargo test --features voip handlers::call::tests -- --nocapture`
- `cargo test -p wacore --features voip voip::registry -- --nocapture`
- `cargo test -p whatsapp-rust-voip-cli`

`cargo test --all` was also attempted; only the external E2E crate failed because its documented mock server was not running (`Connection refused`).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -221,7 +221,7 @@ hmac = { workspace = true }
 libc = "0.2"
 metrics-exporter-prometheus = "0.18"
 sha2 = { workspace = true }
-tokio = { workspace = true, features = ["full"] }
+tokio = { workspace = true, features = ["full", "test-util"] }
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 uuid = { workspace = true, features = ["v4"] }
 wacore = { workspace = true, features = ["test-util"] }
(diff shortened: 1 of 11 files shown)


jlucaso1:
@coderabbitai review

PR #1056, opened 2026-07-19
perf(size): token lookup via static hash table — −29 KiB, faster than hashify on hits and marshal, dep dropped

jlucaso1:
Follow-up to #1055, resolving the documented size-audit leftovers under one constraint: **no performance sacrifice, with measurements to prove it**.

**Result: −29.1 KiB stripped (10,070,744 → 10,040,952 on the local reference build); `.text` −56.2 KiB (~+27 KiB `.rodata`). The marshal-heavy CodSpeed workload runs *faster* than main (137.5M vs 131.6M simulated instructions is +4.5%, but wall-clock is −5%), token hits are ~3.5× faster, and the `hashify` dependency is removed.**

## The change: token lookup, generated code → one static hash table

The `index_of_token` probe ran on `hashify::tiny_map!` output — ~55 KiB of length-bucketed compare chains, the single largest cold-ish symbol left in the binary. It now runs on a build-generated open-addressing table (~25 KiB of `.rodata`):

- **Keys of length ≤ 4** (the bulk of probe traffic: the fat 3–4-byte numeric groups plus the hottest tags) are matched **exactly** by their NUL-padded little-endian `u32` — tokens are NUL-free so the padding can't alias real bytes, and a length tag in the meta word rejects `"a"` vs `"a\0"`.
- **Longer keys** match a two-load signature (`first4 ⊕ rot(last4) ⊕ len·φ`) and byte-verify against a length-prefixed blob only on signature hits.
- **Misses of any shape stop at the first empty slot** — one multiply, one or two loads, no full-key hash or compare. This preserves the property that killed the full-key-PHF idea from #1055 (the encode path probes every stanza string and most miss); that follow-up stays **rejected**, but for the *hash-the-whole-key* reason, not against hash tables per se.
- Power-of-two size at ≤ 75% load; the multiplier is chosen deterministically at build time so the worst probe chain stays ≤ 8.

## How this design was reached (the honest version)

The first iteration of this PR used length-bucketed discriminator tables mirroring hashify's probe shape. CodSpeed correctly flagged it: **−26% on `bench_marshal_auto_many_children`**, reproduced locally (212 → 256 µs) and pinned by callgrind to the per-probe cost — ~68 instructions vs hashify's ~15, dominated by linear scans over 100+-entry numeric runs, `memcmp` calls for 4-byte compares, and `partition_point` machinery. Two rounds of measurement-guided fixes (binary search, then a short-token-only hash table) got to +18% and +4.5% respectively; unifying everything into the single table landed at:

| workload | main (hashify) | this PR | Δ |
|---|---:|---:|---:|
| `marshal_auto_many_children`, wall | 212.4 µs | 201.2 µs | **−5%** |
| same, callgrind instructions | 131.6M | 137.5M | +4.5% |
| token hits, per probe (divan micro) | 24.4 ns | 6.8 ns | **−72%** |

The new `token_lookup_benchmark` pins hits and misses **separately** so a future lookup change can't trade one for the other unnoticed; per a Codex review finding, the miss corpus is filtered through `index_of_token` so dictionary tokens (`"0"`, `"37"`, `"407"`…) can't contaminate the miss metric.

## Correctness

- The pre-existing byte-mutation fuzz (`lookup_matches_reference_under_byte_mutation`) checks the table against a reference map across every single-byte mutation of all 1259 tokens — millions of probes, including NUL and 0xFF injections that exercise the padding and empty-slot edge cases.
- Full token roundtrip both directions; `cargo test -p wacore-binary -p wacore`: 1235 passed. Clippy clean; wasm32 lib build checked.
- Build-time guards (review findings): double-byte dict indices are asserted < 256 before packing (the old generated code failed to compile on overflow; the packed encoding needs an explicit check), tokens are asserted NUL-free, and no key may equal the empty-slot sentinel.

## Other follow-ups from #1055, dispositioned

- **PortableCache value-erasure**: measured at 131 symbols / 118 KiB total, largest 3.6 KiB — death by many small cuts, several caches adjacent to the message path. **Declined.**
- **`-Zbuild-std`**: perf-neutral but operationally fragile (CI/local divergence, toolchain-bump coupling). **Declined.**

---
_Generated by [Claude Code](https://claude.ai/code/session_01AkAW3KSG7GySNMgswxFyUG)_

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1709,18 +1709,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ed5909b6e89a2db4456e54cd5f673791d7eca6732202bbf2a9cc504fe2f9b84a"
 
 [[package]]
-name = "hashify"
-version = "0.2.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dd1246c0e5493286aeb2dde35b1f4eb9c4ce00e628641210a5e553fc001a1f26"
-dependencies = [
- "indexmap",
- "proc-macro2",
- "quote",
- "syn",
-]
-
-[[package]]
 name = "heck"
 version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4007,7 +3995,6 @@ dependencies = [
  "codspeed-divan-compat",
  "compact_str",
  "flate2",
- "hashify",
  "itoa",
  "proptest",
  "serde",
--- a/wacore/Cargo.toml
+++ b/wacore/Cargo.toml
@@ -107,6 +107,10 @@ name = "send_receive_benchmark"
 harness = false
 
 [[bench]]
+name = "token_lookup_benchmark"
+harness = false
+
+[[bench]]
 name = "history_sync_benchmark"
 harness = false
 
--- a/wacore/benches/token_lookup_benchmark.rs
+++ b/wacore/benches/token_lookup_benchmark.rs
@@ -1,0 +1,75 @@
+//! Token-table probe benchmarks: `index_of_token` runs once per string in
+//! every outgoing stanza (tags, attr names, attr values), and most probes
+//! miss. Hit and miss workloads are pinned separately so a lookup-structure
+//! change can't trade one for the other unnoticed.
+
+use divan::black_box;
+use wacore_binary::token::{get_double_token, get_single_token, index_of_token};
+
+fn main() {
+    divan::main();
+}
+
+/// Every dictionary token once — the exhaustive hit workload.
+fn all_tokens() -> Vec<&'static str> {
+    let mut toks = Vec::new();
+    for i in 0..=u8::MAX {
+        if let Some(t) = get_single_token(i)
+            && !t.is_empty()
+        {
+            toks.push(t);
+        }
+    }
+    for dict in 0..4u8 {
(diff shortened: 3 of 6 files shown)


jlucaso1:
The CodSpeed regression on `bench_marshal_auto_many_children` was real and is fixed in 6e5fbac by redesigning the lookup, not by acknowledging the report.

**Root cause** (callgrind, 50 marshals): the discriminator-table probe cost ~68 instructions vs hashify's ~15 — the workload's short numeric attr values ("0".."2047") walk the fat 3–4-byte numeric buckets, and hits paid linear scans over 100+-entry runs plus a libc `memcmp` per compare. My original micro miss corpus (long IDs/phone numbers) never touched that path, which is why the PR's benches missed it.

**Fix**: one open-addressing table over all 1259 tokens. Length ≤ 4 keys match exactly by padded u32; longer keys match a two-load signature and byte-verify on hit; misses stop at the first empty slot. On the regressing workload: **131.6M (main) → 137.5M instructions (+4.5%, was +49%), wall-clock 201 vs 212 µs (−5%)**. Micro hits went 24.4 → 6.8 ns. Local size delta vs main is now −29 KiB stripped (the table costs ~17 KiB of the earlier −46 KiB, spent deliberately to clear the perf gate).

The next CodSpeed run on this head is the arbiter, same as before.

---

PR #1084, opened 2026-07-24
feat(messaging): add edit_message_with_stanza_id to override the outer stanza id

Salientekill:
## What

Adds `Client::edit_message_with_stanza_id`, an edit-path counterpart to the existing `SendOptions::message_id` override.

`edit_message()` deliberately uses a **fresh** outer stanza id — reusing the original id makes the server dedupe against it and silently drop the edit (as its own doc comment notes). That default is correct for normal edits, but there is no public way to edit a message you own while pinning the *outer* stanza id to a caller-chosen value.

This method exposes exactly that: it builds the same `MessageEdit` container as `edit_message`, but sets `SendPipelineOptions.request_id` to the caller-provided `outer_stanza_id`.

## Why

Some callers need to control the outgoing stanza id on an edit — for example to collide it with an existing message so clients re-render that slot. This mirrors the already-public `SendOptions::message_id` escape hatch, just for the edit path.

Whether a collision is honored is server- and client-dependent (the server may dedupe against the outer id), so the API is documented as **best-effort / non-guaranteed** rather than a reliable primitive.

## Notes

- No behavior change to `edit_message` — this is purely additive.
- Follows the existing `edit_message_inner` structure (participant resolution, `build_edit_message`, `send_message_impl`).
- `cargo fmt --all` clean.

### DIFF
--- a/src/client/messaging.rs
+++ b/src/client/messaging.rs
@@ -64,16 +64,49 @@ impl Client {
         original_id: impl Into<String>,
         new_content: wa::Message,
     ) -> Result<String, crate::send::SendError> {
-        self.edit_message_inner(to.into(), original_id.into(), new_content)
+        self.edit_message_inner(to.into(), original_id.into(), new_content, None)
             .await
     }
 
+    /// Edits a message you own (`original_id`) with caller-supplied
+    /// [`crate::send::EditOptions`]. The edit-path counterpart of
+    /// [`crate::send::SendOptions::message_id`] (which overrides the stanza id
+    /// for plain sends): `stanza_id` lets callers control the outer stanza id —
+    /// for example to collide it with an existing message so clients re-render
+    /// that slot.
+    ///
+    /// When `stanza_id` is set, no id-keyed local state is bound to the borrowed
+    /// id (the edit skips outbound-secret and retry-cache persistence, leaving
+    /// the original message's state intact), and whether the collision is
+    /// honored is server/client dependent — treat it as best-effort. See
+    /// [`crate::send::EditOptions::stanza_id`].
+    pub async fn edit_message_with_options(
+        &self,
+        to: impl Into<Jid>,
+        original_id: impl Into<String>,
+        new_content: wa::Message,
+        options: crate::send::EditOptions,
+    ) -> Result<String, crate::send::SendError> {
+        self.edit_message_inner(
+            to.into(),
+            original_id.into(),
+            new_content,
+            options.stanza_id,
+        )
+        .await
+    }
+
+    /// Shared edit-send flow for [`Self::edit_message`] and
+    /// [`Self::edit_message_with_options`]. `request_id` overrides the outer
+    /// stanza id when `Some`; when `None` a fresh one is generated (the default,
+    /// safe behavior — see below).
     #[cfg_attr(feature = "tracing", tracing::instrument(name = "wa.send.edit", level = "debug", skip_all, fields(to = %to.observe()), err(Debug)))]
     async fn edit_message_inner(
         &self,
         to: Jid,
         original_id: String,
         new_content: wa::Message,
+        request_id: Option<String>,
     ) -> Result<String, crate::send::SendError> {
         // WhatsApp Web uses getMeUserLidOrJidForChat(chat, EditMessage) which
         // returns LID for LID-addressing groups and PN otherwise.
@@ -100,16 +133,22 @@ impl Client {
             wacore::time::now_millis(),
         );
 
-        // Use a new stanza ID instead of reusing the original message ID.
-        // The original message ID is already embedded in protocolMessage.key.id
-        // inside the encrypted payload. Reusing it as the outer stanza ID causes
-        // the server to deduplicate against the original message and silently
-        // drop the edit.
+        // Default (`request_id = None`) uses a fresh stanza ID instead of
+        // reusing the original message ID: the original ID is already embedded
+        // in protocolMessage.key.id inside the encrypted payload, and reusing it
+        // as the outer stanza ID makes the server deduplicate against the
+        // original message and silently drop the edit. Callers that intentionally
+        // want to pin the outer stanza id pass it via `request_id`; that id is
+        // borrowed from another message, so id-keyed state (retry cache, outbound
+        // secret) must not be bound to it.
+        let borrowed_message_id = request_id.is_some();
         self.send_message_impl(
             to,
             &edit_container_message,
             crate::send::SendPipelineOptions {
                 edit: Some(crate::types::message::EditAttribute::MessageEdit),
+                request_id,
+                borrowed_message_id,
                 ..Default::default()
             },
         )
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -139,7 +139,7 @@ pub mod runtime_impl;
 pub use runtime_impl::TokioRuntime;
 pub use wacore::runtime::Runtime;
 pub mod send;
-pub use send::{PinDuration, RevokeType, SendError, SendOptions, SendResult};
+pub use send::{EditOptions, PinDuration, RevokeType, SendError, SendOptions, SendResult};
 pub use wacore::send::StanzaType;
 pub mod media;
 pub mod session;
@@ -215,7 +215,7 @@ pub mod prelude {
     pub use crate::request::IqError;
     #[cfg(feature = "tokio-runtime")]
     pub use crate::runtime_impl::TokioRuntime;
-    pub use crate::send::{SendError, SendOptions, SendResult};
+    pub use crate::send::{EditOptions, SendError, SendOptions, SendResult};
     #[cfg(feature = "signal")]
     pub use crate::shutdown::shutdown_signal;
     #[cfg(feature = "sqlite-storage")]
--- a/src/send/mod.rs
+++ b/src/send/mod.rs
@@ -165,6 +165,7 @@ struct GroupBranchRequest<'a> {
     extra_stanza_nodes: &'a [Node],
     group_metadata_freshness: crate::cache::Freshness,
     device_freshness: crate::cache::Freshness,
+    borrowed_message_id: bool,
 }
 
 struct DmBranchRequest<'a> {
@@ -175,6 +176,7 @@ struct DmBranchRequest<'a> {
     extra_stanza_nodes: Vec<Node>,
     is_status_addon: bool,
     device_freshness: crate::cache::Freshness,
+    borrowed_message_id: bool,
 }
 
 enum GroupDeviceSnapshot {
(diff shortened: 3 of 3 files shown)


Salientekill:
Addressed: extracted the shared flow into `edit_message_inner`, which now takes a `request_id: Option<String>`. Both `edit_message` (passes `None`) and `edit_message_with_stanza_id` (passes `Some(outer_stanza_id)`) delegate to it, so no edit-flow logic is duplicated. `cargo fmt --all` clean.

src/client/messaging.rs
```diff
@@ -100,16 +132,18 @@ impl Client {
...
+        // in protocolMessage.key.id inside the encrypted payload, and reusing it
+        // as the outer stanza ID makes the server deduplicate against the
+        // original message and silently drop the edit. Callers that intentionally
+        // want to pin the outer stanza id pass it via `request_id`.
         self.send_message_impl(
             to,
             &edit_container_message,
             crate::send::SendPipelineOptions {
                 edit: Some(crate::types::message::EditAttribute::MessageEdit),
+                request_id,
```
jlucaso1, line 150: Colliding the outer id with an existing own message has a destructive side effect beyond server dedupe: edits generate a reporting token (`should_include_reporting_token` only excludes reactions, poll votes and keep-in-chat, so the edit container qualifies and mints a fresh `message_secret`), and after the wire send `persist_outbound_msg_secret` stores that secret under `(chat, sender, outer_id)`. The sqlite store upserts with `secret = excluded.secret`, so the collision **overwrites the original message's outbound msg_secret** — any future `enc_reaction` or poll vote targeting the original message will fail to decrypt from then on.

At minimum this needs to be spelled out in the doc comment (the current text only covers the harmless "server may dedupe" outcome). Ideally the override path would skip, or at least not clobber, secret persistence for the colliding id.

src/client/messaging.rs
```diff
@@ -100,16 +132,18 @@ impl Client {
...
-        // inside the encrypted payload. Reusing it as the outer stanza ID causes
-        // the server to deduplicate against the original message and silently
-        // drop the edit.
+        // Default (`request_id = None`) uses a fresh stanza ID instead of
+        // reusing the original message ID: the original ID is already embedded
+        // in protocolMessage.key.id inside the encrypted payload, and reusing it
+        // as the outer stanza ID makes the server deduplicate against the
+        // original message and silently drop the edit. Callers that intentionally
+        // want to pin the outer stanza id pass it via `request_id`.
         self.send_message_impl(
```
jlucaso1, line 145: Second casualty of the same collision: the send pipeline calls `add_recent_message(&to, &request_id, ...)` keyed by `(chat, id)` (L1 cache insert + DB upsert), so the retry-cache entry for the original message gets replaced by the edit container. A peer sending a retry receipt for the original id will then receive the edit protocolMessage instead of the original content.

src/client/messaging.rs
```diff
@@ -64,16 +64,48 @@ impl Client {
...
+    /// outgoing stanza id — for example to collide it with an existing message
+    /// so clients re-render that slot.
+    ///
+    /// Whether a collision is honored is server- and client-dependent (the
+    /// server may dedupe against the outer id); callers should treat this as a
+    /// best-effort, non-guaranteed operation.
+    pub async fn edit_message_with_stanza_id(
+        &self,
+        to: impl Into<Jid>,
+        edit_target_id: impl Into<String>,
```
jlucaso1: `edit_target_id` and `outer_stanza_id` are adjacent `impl Into<String>` parameters — swapping them at a call site compiles fine and fails silently (the best-effort semantics will mask the caller's mistake: it would send an edit *of* the wrong message *under* the wrong id).

Longer term, the `_with_stanza_id` suffix also opens a combinatorial axis: we already have `edit_message`, `edit_message_encrypted` and `newsletter().edit_message`, and there's no encrypted+override variant if someone needs one. An `EditOptions` struct (mirroring what `SendOptions::message_id` does for plain sends) would remove the swap hazard and scale without new method permutations.

src/client/messaging.rs
```diff
@@ -64,16 +64,48 @@ impl Client {
...
     ) -> Result<String, crate::send::SendError> {
-        self.edit_message_inner(to.into(), original_id.into(), new_content)
+        self.edit_message_inner(to.into(), original_id.into(), new_content, None)
             .await
     }
 
+    /// Edits a message you own (`edit_target_id`) while overriding the OUTER
+    /// stanza id with `outer_stanza_id`, instead of the fresh id
+    /// [`Self::edit_message`] generates. This is the edit-path counterpart of
+    /// [`SendOptions::message_id`] (which overrides the stanza id for plain
```
jlucaso1: `[`SendOptions::message_id`]` doesn't resolve from this module — `messaging.rs` only has `use super::*` and the parent doesn't import `SendOptions` — so `cargo doc` will emit a `broken_intra_doc_links` warning. Use the full path: `[`crate::send::SendOptions::message_id`]`.

jlucaso1 (changes requested):
Thanks for the writeup, and for already folding the flow into `edit_message_inner` with the instrumented span — that resolved the structural concerns I had with the first revision. What's left is mostly about the advertised use case itself: colliding the outer id with an existing message has two destructive side effects on *local* state that the docs don't mention, so I don't think this can land as-is. Details inline; the short version:

1. **Colliding overwrites the original message's outbound `msg_secret`** — future `enc_reaction`/poll-vote decryption for that message breaks.
2. **Colliding overwrites the retry cache entry** — a retry receipt for the original id re-sends the edit container instead of the original content.
3. API shape: two adjacent `impl Into<String>` params are a silent swap hazard, and the `_with_stanza_id` suffix starts a combinatorial axis of edit variants.
4. Broken intra-doc link in the new docs.

Also: no tests. At minimum I'd want a happy-path test asserting the outgoing stanza carries the caller-supplied id, plus the failure cases (empty id → `InvalidRequest`, DM while logged out → `NotLoggedIn`).

For the record on parity: WA Web always mints a fresh id for the edit's outer key (`WAWebMsgKey.newId_DEPRECATED()` in `docs/captured-js/WAWeb/Send/MessageEditAction.js`), so there is no official path that pins the outer id — this API is off-protocol by construction, which is fine for an escape hatch but worth stating in the docs alongside the dedupe caveat.

On the plus side: empty ids and newsletter JIDs are already rejected downstream in `send_message_impl`, and the participant resolution / `NotLoggedIn` handling faithfully mirrors the existing edit flow.

Salientekill:
On the regression-test suggestion: the stanza-id override isn't new wire logic — `edit_message_with_stanza_id` just sets `SendPipelineOptions.request_id`, which flows through the same `send_message_impl` path the already-public `SendOptions::message_id` override uses. A dedicated transport-level test for the edit path needs full Signal session / sender-key setup (the edit has to actually encrypt and reach the wire for the outgoing stanza id to be captured via `capturing_client`), which feels disproportionate for routing an existing option through a thin delegating wrapper. Happy to add one if you'd prefer it in this PR.

The earlier duplication and tracing comments are resolved by the refactor: both public methods now delegate to the instrumented `edit_message_inner`, so there's a single edit flow and the custom-id path is traced too.

jlucaso1:
Fair point on the plumbing — verified: `SendOptions::message_id` is indeed converted to `SendPipelineOptions.request_id` and goes through the same `send_message_impl` (src/send/mod.rs:837), so the wire path is shared. And agreed that a full transport-level assert is disproportionate here: the `capturing_client` tests only exercise receipt/ack handling, nothing does a full E2E DM send in unit tests, and the e2e crate has no edit coverage either. Happy to drop that ask.

Two caveats though:

1. "Routes through an existing path" isn't quite "routes through a tested path" — as far as I can tell nothing anywhere asserts that a caller-supplied id actually reaches the outgoing stanza, not even for the already-public `SendOptions::message_id`. So the shared plumbing doesn't carry existing coverage with it.

2. The failure-path tests I asked for need no Signal setup at all, and they're the ones I'd still like in this PR:
   - `edit_message_with_stanza_id(jid, "ORIG", "", msg)` → `InvalidRequest`: the empty-id check fires at the top of `send_message_impl` (src/send/mod.rs:1554), before any crypto or branch dispatch, so a bare test client works — same class of cheap validation test as `newsletter_edit_revoke_reject_empty_message_id`. Bonus: this only passes if the *third* argument is the one that lands in `request_id` (an empty `edit_target_id` is never validated — it goes straight into `protocolMessage.key.id`), so it doubles as a guard against the adjacent `impl Into<String>` swap hazard from the review.
   - DM while logged out → `NotLoggedIn`.

With those two in, I'm fine on the testing front. The msg_secret / retry-cache collision side effects from the review are the remaining blockers.

Salientekill:
Thanks for the deep review @jlucaso1 — all four addressed:

1. **msg_secret clobber** & 2. **retry-cache clobber**: `send_message_impl` now threads a `borrowed_message_id` flag (set whenever an edit pins the outer id). When set it skips both `persist_outbound_msg_secret` and `add_recent_message` for the borrowed id, so the original message's outbound secret and retry content stay intact. Normal sends/edits are unaffected (flag defaults `false`).
3. **swap hazard / scaling**: replaced the two-positional-`impl Into<String>` method with an `EditOptions` struct (sibling of `SendOptions`); `edit_message_with_options(to, original_id, content, EditOptions { stanza_id })`. Both edit methods delegate to `edit_message_inner`, so no edit-flow duplication.
4. **doc link**: fixed to `[crate::send::SendOptions::message_id]`.

The doc comment now spells out that a pinned id binds no id-keyed state and the wire collision is best-effort. `cargo fmt --all` clean; `cargo check` green.

Salientekill:
Pushed the phash fix — good catch, same class as the secret/retry clobber: a borrowed id also skips `register_ack_waiter` now (the waiter map is keyed by outer stanza id, so it would overwrite the original send's waiter). All three id-keyed side effects (outbound secret, retry cache, ack waiter) are now guarded by `borrowed_message_id`.

On the failure-path tests (empty `stanza_id` -> `InvalidRequest`, logged-out -> `NotLoggedIn`): both hit early validation before any encryption, so they're cheap to add — happy to include them if you'd like them in this PR.

jlucaso1:
Verified the new revision — nice work, this addressed everything:

- `borrowed_message_id` correctly guards all three id-keyed side effects, in both branches (including the status-addon path of `add_recent_message`), and the retry/resend path keeps the flag off, which is right — the re-bind there is intentional.
- `EditOptions` kills the swap hazard and the method-permutation axis, and it's stylistically consistent with `SendOptions`.
- On CodeRabbit's "also reject a borrowed id with an active waiter": I'm fine with skip-only. The residual (an in-flight original waiter consuming the edit's ack) needs a collision with a message sent milliseconds earlier — not worth extra plumbing for an escape hatch.

Two things left:

1. **Yes, please add the two failure-path tests in this PR** (empty `stanza_id` → `InvalidRequest`, logged-out DM → `NotLoggedIn`) — per my earlier comment, the empty-id one doubles as a guard that `EditOptions::stanza_id` is what actually lands in `request_id`.
2. Small irony: the new doc comment introduces two links of the same class we just fixed — `[`EditOptions`]` and `[`EditOptions::stanza_id`]` in `messaging.rs` don't resolve (nothing from `crate::send` is in scope there; the signature uses the full path). Use `[`crate::send::EditOptions`]` / `[`crate::send::EditOptions::stanza_id`]`.

With those two in I'll flip this to approve.

Salientekill:
Both done, thanks @jlucaso1:

1. Added the two failure-path tests (`edit_message_logged_out_dm_returns_not_logged_in`, `edit_message_with_empty_stanza_id_returns_invalid_request`) — both pass. The empty-id one asserts the whole path, so it doubles as the guard that `stanza_id` reaches `request_id`.
2. Fixed the doc links to `[crate::send::EditOptions]` / `[crate::send::EditOptions::stanza_id]` (nice catch — exactly the class we'd just fixed).

Ready for another look.

Salientekill:
Lint fixed — `EditOptions` has a single field, so `..Default::default()` in the new test tripped `clippy::needless_update`. Dropped it (and in the caller). `cargo clippy -p whatsapp-rust --all-targets -- -D warnings` clean now.

PR #1099, opened 2026-07-24
fix(chat-store): key receipts and contacts by the peer's bare identity

### LINKED ISSUE #1095: chat-store: receipts from a peer's companion device (device-suffixed JID) never advance message status; the #1083 alt-key fallback is device-gated
### Summary

Delivery/read receipts sent by a peer's **companion device** (WhatsApp Web/Desktop, i.e. a device-suffixed JID like `10203040506070:48@lid`) never advance `messages.status` in the chat-store.
Three facts compose into the bug:

1. `src/receipt.rs` keeps the raw `from` attr as `Receipt.source.chat` for 1:1 receipts (device and agent intact), while inbound *messages* normalize (`chat: from.to_non_ad()` in `wacore/src/messages.rs`). Receipts are the only event feeding AD-form chat keys into the store.
2. `storages/chat-store/src/store.rs::apply_receipt` derives its SQL key with `receipt.source.chat.to_string()` and matches `messages.chat_jid` by exact string. The row lives under `10203040506070@lid`; the receipt key renders `10203040506070:48@lid`; the UPDATE matches 0 rows.
3. The alternate-key fallback added by #1083 never runs: `lid.rs::user_chat` is gated on `jid.device == 0`, so `counterpart_chat_key` returns `None` for a device-suffixed key and the miss is silently discarded.

Net effect: **ticks only ever advance when the peer's primary phone (device 0) acks.**
For reads it is worse than a delay: in multi-device, the read receipt is emitted once, by whichever device the user read on first. If that was a companion, the primary never re-sends it, so the message is stuck below `Read` **permanently**.
Official semantics are any-device ("delivered to the recipient's phone **or any of their linked devices**", [WhatsApp FAQ](https://faq.whatsapp.com/665923838265756)), and the status column is already a monotonic max, so matching companion receipts is sufficient for parity.

### Environment

Reproduced at `00771869` (the #1083 merge) and confirmed unchanged at current `main` (`040ca136`); none of the commits in between touch this path.

### Observed wire payload (redacted, real session)

A serialized `Event::Receipt` after the peer's linked device delivered an outgoing DM while their phone was offline:

```json
{
  "message_ids": ["3EB0824E3930E4FA0C152B"],
  "offline": false,
  "source": {
    "chat":   { "user": "10203040506070", "server": "lid", "agent": 1, "device": 48, "integrator": 0 },
    "sender": { "user": "10203040506070", "server": "lid", "agent": 1, "device": 48, "integrator": 0 },
    "is_from_me": false,
    "is_group": false
  },
  "timestamp": "2026-07-24T19:09:01Z",
  "type": "Delivered"
}
```

(The `agent: 1` is the binary decoder's LID domain-type byte, suppressed by `to_string()`; only `device` matters here.)

### Exact trace

| step | value |
|---|---|
| receipt `source.chat` | `Jid { user: "10203040506070", server: Lid, agent: 1, device: 48 }` |
| `store.rs` `apply_receipt`: `chat.to_string()` | `"10203040506070:48@lid"` |
| stored row `messages.chat_jid` (from `route_chat_key` on the send/inbound path) | `"10203040506070@lid"` |
| direct UPDATE (`message_row`, exact `chat_jid` equality) | 0 rows, id pushed to `missed` |
| `counterpart_chat_key(...)` -> `user_chat("10203040506070:48@lid")` | `device == 48 != 0` -> `None` |
| alt-key retry | never runs |
| result | status stays at `ServerAck`; a `Read` from a companion is lost forever |

### Collateral from the same root (device-suffixed keys entering the store)

- **`read-self` / `played-self` receipts create phantom chats.** That branch routes through `route_chat_key`, which also starts with the device-gated `user_chat`, so a device-suffixed self receipt falls through unchanged and `ensure_chat` materializes a stray `user:48@lid` chat row; the unread recount lands on the phantom instead of the real thread.
- **Group per-user receipt rows split per device.** The aggregated `<participants>` parser sets `sender: user.jid` (device-suffixed), and the `message_receipts` upsert keys on `sender.to_string()`, so one participant reading from phone and Web produces two rows.
- **Contact names miss for companion senders.** `apply_inbound` upserts push names and business verified names keyed by `info.source.sender.to_string()`; messages keep the device on `sender` by design, so a peer texting from Web writes `contacts` rows under `user:48@lid` that `ChatStore::contact()` (exact JID lookup) never finds.
- **`reconcile_chat` is a silent no-op for AD input**, since it funnels into the same gated resolver.

### Why the fix belongs at the store boundary, not the parser

Stripping the device inside the receipt parser would be wrong: the retry pipeline and the receipt echo need the full device JID (see the `buildBaseReceipt` comment in `src/receipt.rs`: the `to` must echo `from` verbatim or the LID server rejects it for multi-device DMs).
WA Web itself normalizes at match time (`fixMsgKeysWithPnMapping` / alternate msg keys operate on bare chat keys), which is the model #1083 followed.

Suggested minimal shape:

- `apply_receipt`: `let chat = receipt.source.chat.to_non_ad_string();` and `let user = receipt.source.sender.to_non_ad_string();` (the ReadSelf branch then feeds `route_chat_key` a bare key for free). `Jid::to_non_ad_string()` already exists and is single-allocation.
- `lid.rs::user_chat`: normalize (`device`/`agent` zeroed) instead of rejecting, so `counterpart_chat_key` / `chat_key_candidates` / `route_chat_key` / `reconcile_chat` become device-tolerant even if another AD form slips in later.
- `apply_inbound`: key the push-name/business-name upserts by `sender.to_non_ad_string()`.
- Possibly a heal for rows already written under AD keys by the current code (phantom chats, split `message_receipts`, orphan contacts); your call whether that is worth a migration or a doc note.

### Failing test (drops into `storages/chat-store/tests/chat_store_test.rs`)

Note it deliberately has **no** `lid_pn_mapping` row: the direct match must succeed on the normalized key, since a fresh LID-only thread has no counterpart to fall back to.

```rust
#[tokio::test]
async fn companion_device_receipt_advances_status() {
    let (_store, chat_store) = test_store().await;
    let chat = jid("10203040506070@lid");

    chat_store
        .record_outgoing(
            &chat,
            "OUT-AD-1",
            &wa::Message::text("hello"),
            Utc.timestamp_opt(1_700_000_100, 0).unwrap(),
        )
        .unwrap();
    chat_store.flush().await.unwrap();

    // Receipt from the peer's linked device: the binary decoder yields
    // user:48@lid with agent=1 (the LID domain-type byte).
    let companion = Jid {
        user: "10203040506070".into(),
        server: wacore_binary::Server::Lid,
        agent: 1,
        device: 48,
        integrator: 0,
    };
    feed(
        &chat_store,
        [Event::Receipt(
            Receipt::builder()
                .source(MessageSource {
                    chat: companion.clone(),
                    sender: companion,
                    ..Default::default()
                })
                .message_ids(vec!["OUT-AD-1".to_string()])
                .timestamp(Utc.timestamp_opt(1_700_000_200, 0).unwrap())
                .r#type(ReceiptType::Delivered)
                .offline(false)
                .build(),
        )],
    )
    .await;

    let msg = chat_store.message(&chat, "OUT-AD-1").await.unwrap().unwrap();
    // Today this is still MessageStatus::Pending: the receipt was dropped.
    assert_eq!(msg.status, MessageStatus::Delivered);
}
```

Every existing receipt test in the suite builds `MessageSource` with bare JIDs, which is why the (otherwise thorough) #1083 coverage did not catch this.
Variants worth covering alongside: `Read` from a companion after `Delivered` from device 0, a device-suffixed receipt resolving across a PN/LID mapping (alt-key retry path), a device-suffixed `read-self` (unread recount on the real thread, no phantom chat row), and a companion-sent inbound message landing its push name on the bare contact key.

Happy to provide more traces or test the fix against our live sessions.

jlucaso1:
## Summary

Delivery and read receipts sent by a peer's **companion device** (WhatsApp Web/Desktop, a device-suffixed JID like `10203040506070:48@lid`) never advanced `messages.status`. Reproduced on `main`; the issue's repro test is included verbatim.

`apply_receipt` derived its SQL key from `receipt.source.chat.to_string()`, and receipts are the one event that reaches the store with the peer's wire identity intact — `src/receipt.rs` keeps the device on `chat` because the retry pipeline and the receipt echo need the full JID, while inbound messages normalize (`chat: from.to_non_ad()` in `wacore/src/messages.rs`). So the row lived under `10203040506070@lid`, the receipt rendered `10203040506070:48@lid`, and the UPDATE matched zero rows. The #1083 alternate-key fallback could not rescue it either: `lid::user_chat` was gated on `device == 0`, so `counterpart_chat_key` returned `None` for a device-suffixed key and the miss was silently discarded.

For reads that is permanent rather than delayed. Multi-device emits the read receipt once, from whichever device the user read on first; the primary never re-sends it, so a message read on Web stays below `Read` forever. Official semantics are any-device ("delivered to the recipient's phone **or any of their linked devices**"), and `messages.status` is already a monotonic max, so matching companion receipts is sufficient for parity.

The normalization belongs at the store boundary, not in the parser: `buildBaseReceipt` needs the `to` of an echoed receipt to repeat `from` verbatim or the LID server rejects it for multi-device DMs, and `resolve_retry_chat_info` reads the device off `receipt.source.chat`. WA Web normalizes at match time for the same reason (`fixMsgKeysWithPnMapping` / `getAlternateMsgKey` operate on bare chat keys), which is the model #1083 followed.

Three more defects fall out of the same root, all confirmed before the fix:

- a device-suffixed `read-self` materialized a stray `user:48@lid` chat row, and the unread recount landed on the phantom instead of the real thread;
- one group participant reading on phone and on Web produced two `message_receipts` rows, since the upsert keyed on the device-suffixed `sender`;
- a peer texting from Web wrote its push name under `user:48@lid`, which `ChatStore::contact` (exact lookup) never finds.

## Changes

- **`apply_receipt` keys rows bare.** `chat` and the per-participant `user` both go through `Jid::to_non_ad_string()`, so the direct match, the `ReadSelf` routing and the read-by rows all land on the identity the rows are stored under. Read-by is now one row per member, not per device.
- **`lid::user_chat` normalizes instead of rejecting.** Dropping the `device == 0` gate makes `counterpart_chat_key`, `chat_key_candidates`, `route_chat_key` and `reconcile_chat` device-tolerant even if another AD form slips in later. `counterpart_of` is split out so callers that need the normalized key themselves don't parse twice.
- **`route_chat_key` returns the normalized key even with no counterpart.** Normalizing inside `user_chat` alone is not enough: the no-mapping path returned `wire_chat` verbatim, so a device-suffixed key with no `lid_pn_mapping` row still reached `ensure_chat`. This is what stops the phantom thread.
- **Contacts agree on the bare key on both sides.** A `contact_key` helper normalizes in the three `upsert_contact_*` writers (covering the history-sync call sites, which pass raw strings), and `ChatStore::contact` looks up with `to_non_ad_string()`. The helper only allocates when the JID actually carries a device or agent.
- **New migration `2026-07-24-000000_bare_identity_keys`** folds what the old writers left behind, since none of it self-heals: phantom chats are deleted (guarded by `NOT EXISTS` on `messages`, so a device-keyed row that somehow owns messages is left alone rather than orphaned), device-keyed contacts fold onto the bare key with an existing bare row winning, and split read-by rows collapse keeping the highest receipt type — the same monotonic rule `merge_split_chat` already applies. `down.sql` documents that the fold is irreversible: the device that keyed each row is recorded nowhere.

## Coverage

Seven tests in `storages/chat-store/tests/chat_store_test.rs`. Every pre-existing receipt test builds `MessageSource` with bare JIDs, which is why the otherwise thorough #1083 coverage did not catch this.

- `companion_device_receipt_advances_status` — the issue's repro, deliberately with no `lid_pn_mapping` row: a fresh LID-only thread has no counterpart, so the direct match must succeed on the normalized key.
- `companion_read_advances_past_primary_delivered` — `Read` from a companion after `Delivered` from device 0.
- `companion_receipt_resolves_across_pn_lid_mapping` — device-suffixed *and* addressed by the other identity, which only works if normalization happens before the alternate-key retry.
- `companion_read_self_recounts_the_real_chat` — unread recount on the real thread, no phantom row.
- `companion_sender_push_name_lands_on_the_bare_contact` — including that a caller holding the message's `sender` still finds the row.
- `group_receipts_from_two_devices_keep_one_row_per_participant`.
- `migration_folds_device_suffixed_rows` — seeds the artifacts and replays `up.sql` through `batch_execute` (the file is run exactly as the migration harness runs it), covering the guarded phantom chat, the bare-row-wins contact fold, and both the bare-plus-device and device-only receipt splits.

## Validation

```
cargo fmt --all --check
cargo clippy -p whatsapp-rust-chat-store --all-targets --all-features -- -D warnings
cargo test -p whatsapp-rust-chat-store --all-features   # 69 passed
RUSTDOCFLAGS="-D warnings" cargo doc -p whatsapp-rust-chat-store --no-deps
```

Full matrix left to CI.

Fixes #1095.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ur24DTyLiy2NuR1HhTJ9AE)_

### DIFF
--- a/src/voip/facade.rs
+++ b/src/voip/facade.rs
@@ -807,7 +807,7 @@ fn client_weak(client: &Client) -> std::sync::Weak<Client> {
 }
 
 /// Time to wait for the server's `<ack type=offer>` carrying the relay before giving up.
-const OFFER_ACK_RELAY_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+const OFFER_ACK_RELAY_TIMEOUT: Duration = Duration::from_secs(10);
 
 /// Three 60 ms frames absorb scheduling jitter without building a long capture delay.
 const MIC_CHANNEL_CAPACITY: usize = 3;
@@ -2151,7 +2151,7 @@ mod tests {
             if !sent.lock().unwrap().is_empty() {
                 break;
             }
-            tokio::time::sleep(std::time::Duration::from_millis(10)).await;
+            tokio::time::sleep(Duration::from_millis(10)).await;
         }
         assert!(
             !sent.lock().unwrap().is_empty(),
@@ -2165,7 +2165,7 @@ mod tests {
             ))
             .await
             .unwrap();
-        tokio::time::timeout(std::time::Duration::from_secs(2), handle.wait_ended())
+        tokio::time::timeout(Duration::from_secs(2), handle.wait_ended())
             .await
             .expect("wait_ended must resolve after the relay disconnects");
         assert_eq!(
@@ -2220,7 +2220,7 @@ mod tests {
             ))
             .await
             .unwrap();
-        tokio::time::timeout(std::time::Duration::from_secs(2), handle.wait_ended())
+        tokio::time::timeout(Duration::from_secs(2), handle.wait_ended())
             .await
             .expect("wait_ended must resolve after the relay disconnects");
 
@@ -2369,7 +2369,7 @@ mod tests {
         .expect("replacement spawn_call");
 
         // The replacement aborted the stale task; its wait_ended must still resolve.
-        tokio::time::timeout(std::time::Duration::from_secs(2), stale.wait_ended())
+        tokio::time::timeout(Duration::from_secs(2), stale.wait_ended())
             .await
             .expect("stale handle wait_ended must resolve, not hang");
     }
@@ -2409,9 +2409,9 @@ mod tests {
         };
         // Let the waiter register its listener and pass the still-present phase check, so it is truly
         // parked on `listener.await` (the path the guard must cover), not the early return.
-        tokio::time::sleep(std::time::Duration::from_millis(20)).await;
+        tokio::time::sleep(Duration::from_millis(20)).await;
         handle.hangup().await;
-        tokio::time::timeout(std::time::Duration::from_secs(2), waiter)
+        tokio::time::timeout(Duration::from_secs(2), waiter)
             .await
             .expect("wait_ended must resolve after hangup aborts the task")
             .expect("waiter task");
@@ -2606,7 +2606,7 @@ mod tests {
             "the relay-attach material must be parked pending the relay"
         );
 
-        let node = tokio::time::timeout(std::time::Duration::from_secs(2), waiter)
+        let node = tokio::time::timeout(Duration::from_secs(2), waiter)
             .await
             .expect("offer must be sent")
             .expect("waiter");
@@ -2740,7 +2740,7 @@ mod tests {
         .await
         .expect("place encoded call");
 
-        let node = tokio::time::timeout(std::time::Duration::from_secs(2), waiter)
+        let node = tokio::time::timeout(Duration::from_secs(2), waiter)
             .await
             .expect("offer must be sent")
             .expect("waiter");
@@ -2812,7 +2812,7 @@ mod tests {
         .await
         .expect("place_call");
 
-        let node = tokio::time::timeout(std::time::Duration::from_secs(2), waiter)
+        let node = tokio::time::timeout(Duration::from_secs(2), waiter)
             .await
             .expect("offer must be sent")
             .expect("waiter");
@@ -2871,7 +2871,7 @@ mod tests {
             "the offer for the surviving devices must be sent"
         );
 
-        let node = tokio::time::timeout(std::time::Duration::from_secs(2), waiter)
+        let node = tokio::time::timeout(Duration::from_secs(2), waiter)
             .await
             .expect("offer must be sent")
             .expect("waiter");
@@ -3015,7 +3015,7 @@ mod tests {
             0,
             "hangup must deregister the dormant call"
         );
-        tokio::time::timeout(std::time::Duration::from_secs(2), handle.wait_ended())
+        tokio::time::timeout(Duration::from_secs(2), handle.wait_ended())
             .await
             .expect("dormant hangup must resolve wait_ended (no engine task to notify it)");
     }
@@ -3033,7 +3033,7 @@ mod tests {
             client.pending_outgoing_calls.lock().unwrap().is_empty(),
             "disconnect must drain dormant outgoing calls"
         );
-        tokio::time::timeout(std::time::Duration::from_secs(2), handle.wait_ended())
+        tokio::time::timeout(Duration::from_secs(2), handle.wait_ended())
             .await
             .expect("disconnect must resolve a dormant call's wait_ended");
     }
@@ -3106,7 +3106,7 @@ mod tests {
             if client.call_registry().active_count() == 1 {
                 break;
             }
-            tokio::time::sleep(std::time::Duration::from_millis(5)).await;
+            tokio::time::sleep(Duration::from_millis(5)).await;
         }
         assert_eq!(
             client.call_registry().active_count(),
@@ -3124,7 +3124,7 @@ mod tests {
             0,
             "the spawned task must not resurrect a stale entry after cleanup"
         );
-        tokio::time::timeout(std::time::Duration::from_secs(2), handle.wait_ended())
+        tokio::time::timeout(Duration::from_secs(2), handle.wait_ended())
             .await
             .expect("an aborted-before-poll task must still notify ended via the drop-guard");
     }
@@ -3233,17 +3233,17 @@ mod tests {
             }
         });
         // Let attach_engine reach the gated connect before hanging up.
-        tokio::time::sleep(std::time::Duration::from_millis(30)).await;
+        tokio::time::sleep(Duration::from_millis(30)).await;
 
         handle.hangup().await;
 
-        tokio::time::timeout(std::time::Duration::from_secs(2), handle.wait_ended())
+        tokio::time::timeout(Duration::from_secs(2), handle.wait_ended())
             .await
             .expect(
                 "hangup in the connect window must wake wait_ended without the dial completing",
             );
 
-        let res = tokio::time::timeout(std::time::Duration::from_secs(2), attach)
+        let res = tokio::time::timeout(Duration::from_secs(2), attach)
             .await
             .expect("attach_engine must return once hangup aborts the dial")
             .expect("attach task");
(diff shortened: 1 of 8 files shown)


jlucaso1:
Pushed three things on top of the original commit.

**Review feedback (both CodeRabbit comments applied).** The contacts fold was genuinely non-deterministic: with two device-keyed rows for the same peer carrying different names and no bare row to win the insert, which one landed was left to the scan order. `INSERT OR IGNORE` resolves in SELECT order, so the source rows are now ordered and the newest wins deterministically. The two `WHERE` clauses flagged by SQLFluff moved onto their own line.

**Merged `main`** (`7df3a3c5`) into the branch.

**`Build & Lint (all features)` was already red on `main`**, independent of this PR — see [run 30124720041](https://github.com/oxidezap/whatsapp-rust/actions/runs/30124720041/job/89585382475) on `7df3a3c5`. Since it blocks this PR too, the fix is included here rather than reported and left: the `unused_qualifications` lint from #1093 was swept with `cargo fix` under the default feature set, so the voip-gated code — which only compiles with `--all-features` — kept 34 qualifications that the clippy step rejects under `-D warnings`. Every rewrite is the compiler's own machine-applicable suggestion (`std::time::Duration` where `Duration` is already imported, 33 in `src/voip/facade.rs` with all but one inside `mod tests`, plus one `crate::types::call::VideoState` in `wacore/src/voip/registry.rs`). No behavior change, and `cargo fmt` reflows nothing. Happy to split it into its own PR if you would rather keep this one to the chat-store.

`Semver Checks (informational)` is red here and on `main` alike — it is `continue-on-error` and the findings are pre-existing proto drift.

---

jlucaso1:
On the `message_receipts.chat_jid` follow-up (outside-diff comment on `up.sql` 45-70) — skipping that one deliberately, because the row it heals cannot exist.

`message_receipts` has exactly one writer, the group branch of `apply_receipt` (`store.rs:1361`), gated on `receipt.source.chat.is_group()` — which is `server() == Server::Group`, i.e. `@g.us`. `chat_jid` is therefore always a group JID, and a group JID never carries a device: the multi-device suffix identifies a *user's* linked device, so `<group>:48@g.us` is not a shape the server emits. The only other toucher of the table is `merge_split_chat`, which moves rows between a peer's PN and LID keys, and 1:1 receipts never write a row here at all.

That is also why the bug this PR fixes never corrupted `chat_jid` in the first place — the split it caused was on `user_jid`, which does carry the device (`sender`), and that is what the migration folds.

Normalizing `chat_jid` too would mean repeating the CASE expression across the dedup comparison, the rename and the cleanup, for a state no writer can produce. In a data migration that trade seems the wrong way round, so I would rather keep the SQL narrow. Happy to add it if you would prefer the belt-and-braces version.

---

PR #1103, opened 2026-07-25
perf(stats): stop dating every wire frame for a field nothing reads

jlucaso1:
## Summary

A consumer reported the core asking for the time 16 times per direct message in a ping-pong benchmark, which on a platform without a cheap clock is a boundary crossing per read. This batch started as an audit of that attribution and ends with one real cut plus a decision to leave the rest exact. The cut: `SessionStats::record_frame_sent` read the wall clock on every frame written, and only one of the two fields it fed actually needed an instant.

## Audit

I could not run the reporter's embedder, so I instrumented the core instead. `wacore::time` now counts every read that leaves the module, behind the existing `test-util` feature. Counting at the boundary rather than inside a provider sidesteps the `OnceLock`: the process's first `now_millis()` fills `TIME_PROVIDER` with the default, so a suite sharing one process cannot install an instrumented provider per test. The counters are per thread, so a measurement stays exact while the rest of the suite runs in parallel.

Measured on a warm client (registry, LID mapping and Signal sessions seeded) sending one direct message, counting the core's own thread:

| | before | after |
|---|---|---|
| wall reads | 5 | 4 |
| monotonic reads | 2 | 2 |

Per-frame and per-event bookkeeping, measured directly on `SessionStats`:

| | before | after |
|---|---|---|
| frame sent, anchor already armed | 1 | 0 |
| frame sent, arming the anchor | 1 | 1 |
| received transport event | 1 | 1 |

How much this is worth depends on the traffic shape, and it is worth saying plainly. The anchor is cancelled by every receive, so in a strictly alternating request/response pattern the first send after each receive re-arms and still pays: only the extra frames of the same burst come out free, and the saving is small by construction. It scales with how many frames a send burst writes before the next receive, so offline sync, group fanout and bulk sends are where it actually pays.

Two corrections to the original attribution:

- **`last_data_received_ms` has two internal readers, not one.** Besides `is_dead_socket`, the keepalive loop reads it at `src/keepalive.rs:165` to decide whether to skip the ping. Both measure elapsed time, so the field cannot be sampled: the only refresh trigger the core has is the keepalive tick itself, which is 15-30 s apart and coarser than the 15 s gate it would feed. Deriving the watchdog from counters observed at tick boundaries was the other option, and it pushes worst-case detection from ~50 s to ~80 s. Rejected, and the reasoning now lives next to the field.
- **`last_data_sent_ms` had zero internal readers.** Nothing in the core ever called the getter. That is the read this PR removes.

The cache finding held up, with a caveat: `PortableCache::get` did sample the monotonic clock before the lookup, so a miss paid for a timestamp nothing compared. But in the steady state that dominates a ping-pong the registry lookups are hits, so moving the read after the lookup measures the same 2 reads per send here. It is still the correct shape and it is now pinned by a test.

## Changes

- `SessionStats` reads the clock only when the dead-socket anchor arms, not per frame. Sends that continue an already-armed burst cost nothing.
- **Breaking**: `StatsSnapshot::last_data_sent_ms` and `SessionStats::last_data_sent_ms()` removed. Migration: `frames_sent` answers "is it still sending?", `first_send_since_recv_ms()` dates the send the watchdog anchors on. An embedder that needs "when did I last write?" should stamp it at its own send call site rather than make the wire path pay for it.
- `PortableCache::get`/`remove` read the monotonic clock after the lookup, so a miss is free.
- `is_dead_socket_at` / `ms_since_at` take the instant explicitly. The keepalive tick's dead-socket branch now evaluates both against one read, and the watchdog boundary becomes testable without depending on how long a test took to run.
- `wacore::time::clock_reads` (test-util only) counts reads per thread.
- The stats "Cost model" section now states the per-frame and per-event read counts, so the next PR cannot multiply them without noticing.
- `create_iq_test_client` wires the noise socket to the client's stats like the real one does, so per-frame bookkeeping is part of what tests observe.

## Guarantees

Written as tests, not prose:

- `wire_bookkeeping_reads_the_clock_only_where_a_value_is_used` pins 0 reads for sends under an armed anchor, 1 for the arming send, 1 per received event, 2 for a multi-frame batch.
- `dm_send_stays_within_its_clock_budget` asserts the per-send budget and fails if it grows.
- `received_stanza_handling_reads_no_clock` pins that handling a stanza asks for no time.
- `dead_socket_boundary_is_exact` and `continued_sends_do_not_hide_a_silent_socket` prove detection against a supplied instant, including the exact deadline. The existing `dead_socket_anchor_holds_across_continued_sends` and `stale_pre_receive_anchor_self_heals_on_next_send` pass unchanged.
- `expiry_boundary_is_exact_under_a_controlled_clock` covers the entry that expires exactly at the TTL and TTI deadlines; `a_miss_does_not_read_the_clock` pins the lazy read.
- `wire_timestamp_keeps_real_time` proves a timestamp that reaches the server still carries the real second.

No embedder has to register a new provider, change a provider signature, or keep time state of its own.

## Validation

```
cargo fmt --all --check
cargo test -p wacore --lib
cargo test -p whatsapp-rust --lib
```

Full matrix left to CI.

### DIFF
--- a/agent_docs/observability.md
+++ b/agent_docs/observability.md
@@ -35,7 +35,8 @@ chokepoints:
 It also owns the activity timestamps the keepalive dead-socket watchdog reads:
 `last_data_received_ms` (one clock read per received transport event, plus one
 more when that event carries several frames, so a slow drain is not read as
-silence) and `first_send_since_recv_ms` (read only on the send that arms). There
+silence) and `first_send_since_recv_ms`, which every frame loads but only the
+send that arms or re-arms the anchor spends a clock read on. There
 is deliberately no "last send" timestamp: nothing in the core reads one, and it
 cost a clock read on every frame written, which is the client's hottest path
 and a call out of the module on wasm32/embedded. `frames_sent` answers "is it
--- a/agent_docs/observability.md
+++ b/agent_docs/observability.md
@@ -32,9 +32,13 @@ chokepoints:
   transport write — post-noise wire bytes (frame header + AEAD tag included).
 - **Received**: the read loop (`node_io.rs`) per `DataReceived` batch.
 
-It also owns the `last_data_sent_ms`/`last_data_received_ms` activity
-timestamps the keepalive dead-socket watchdog reads (they were loose fields on
-`Client` before). Message-level counters piggyback on the existing
+It also owns the activity timestamps the keepalive dead-socket watchdog reads:
+`last_data_received_ms` (one clock read per received transport event) and
+`first_send_since_recv_ms` (read only on the send that arms the anchor). There
+is deliberately no "last send" timestamp: nothing in the core reads one, and it
+cost a clock read on every frame written, which is the client's hottest path
+and a call out of the module on wasm32/embedded. `frames_sent` answers "is it
+still sending?" for free. Message-level counters piggyback on the existing
 `telemetry::send`/`recv` chokepoints; reconnect attempts are counted in the
 run loop. VoIP relay sockets pass `None` and are not counted — this is the
 main WA session socket only.
--- a/agent_docs/observability.md
+++ b/agent_docs/observability.md
@@ -33,8 +33,9 @@ chokepoints:
 - **Received**: the read loop (`node_io.rs`) per `DataReceived` batch.
 
 It also owns the activity timestamps the keepalive dead-socket watchdog reads:
-`last_data_received_ms` (one clock read per received transport event) and
-`first_send_since_recv_ms` (read only on the send that arms the anchor). There
+`last_data_received_ms` (one clock read per received transport event, plus one
+more when that event carries several frames, so a slow drain is not read as
+silence) and `first_send_since_recv_ms` (read only on the send that arms). There
 is deliberately no "last send" timestamp: nothing in the core reads one, and it
 cost a clock read on every frame written, which is the client's hottest path
 and a call out of the module on wasm32/embedded. `frames_sent` answers "is it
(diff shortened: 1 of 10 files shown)


jlucaso1:
Took three of the four:

- Anchor arming is now a CAS that re-checks the same condition, so a race between two first sends keeps the earlier deadline. The plain load still gates the clock read, so the free path is unchanged.
- Cost model and the observability doc now say the multi-frame batch spends a second read on completion.
- Reworded the two `_at` helper docs.

Skipped the phone number: `15551234001` is copied from the existing test at `src/send/mod.rs:2745`, and the repo convention is plain `555XXXXXXX` (`5551234567` alone shows up 135 times in `src/`). Changing just this one would make it the odd one out.

On the CAS, worth noting for whoever reads this later: `record_frame_sent` is only reached from the noise sender task, which processes send jobs serially, so the race was not reachable on the main session socket. Took it anyway since it costs nothing.

jlucaso1:
Both fair, and the CAS one is a real correction on my part.

Greptile is right that the comment overstated it: the CAS guarantees that once an anchor is set a later send cannot overwrite it, but two senders that both observe it unarmed resolve by whichever CAS lands first, which may well be the one holding the later instant. My commit message from the previous push said "keep the earlier deadline", which is wrong for the same reason. Comment now states the actual guarantee and notes the serial sender task makes the case unreachable in the first place.

Also reworded the observability doc: the anchor is loaded on every frame, only the arming send spends the clock read.

PR #1106, opened 2026-07-25
perf(send): stamp one message from one clock read

jlucaso1:
## Summary

Follow-up to #1103, applied to the send path. Sending one direct message read the wall clock four times on the client's own thread for what is a single logical instant: the message id, the biz node, the privacy-token decision and the outbound message secret. Sample it once as `SendInstant` and carry it down. Besides the reads, this fixes an incoherence: four reads microseconds apart can straddle a second boundary and leave one message described by different seconds in its id, its stanza and its store row.

## Audit

Measured with the `clock_reads` counter #1103 added, attributing every read by stack over one steady-state send (registry, LID mapping and Signal sessions warm, one frame written):

| call site | before | after |
|---|---|---|
| `RequestUtils::generate_message_id` | 1 | 0 |
| biz node in `send_message_with_options_inner` | 1 | 0 |
| `should_send_new_tc_token_with` | 1 | 0 |
| `persist_outbound_msg_secret` | 1 | 0 |
| `SendInstant::now` | 0 | 1 |
| **wall total** | **4** | **1** |

Not visible in that harness but on the same instant: `is_tc_token_expired_with` runs whenever a stored token exists, so a real send against a known contact read five times, not four. Both privacy-token decisions now take the shared instant.

The code already knew these were one instant. `persist_outbound_msg_secret` carried a comment saying it wants "the parent event time" and re-read the clock only because it had no way to receive one.

## Changes

- `SendInstant`, sampled at the top of `send_message_with_options_inner` and carried through `SendPipelineOptions` into the DM branch, the privacy-token decision and the outbound secret. `send_message_impl` samples its own when a caller did not, so the other entry points keep exactly one read too.
- `generate_message_id_at`, `should_send_new_tc_token_with_at`, `is_tc_token_expired_with_at`: `_at` variants that take the instant, matching `should_send_new_tc_token_at` and the `is_dead_socket_at` shape from #1103. The clock-reading versions stay, so nothing becomes a required parameter of the public API.
- `src/send/mod.rs` gains the cost-model line the module lacked, which is how it accumulated four reads without anyone noticing.
- The stats cost model now records that one direct message arrives as roughly four transport events, so "one read per event" is four per message. That corrects the assumption #1103 reasoned from, without changing its conclusion.

## Left alone, and why

Each of these is recorded next to the code so the next reader does not redo the analysis.

- **Device-registry lookups (2 monotonic reads/send).** They could share one instant: the TTL is an hour, so microseconds cannot change the outcome. Sharing it needs a clock-taking method on `PortableCache` and `TypedCache`, both public, and the parameter would be meaningless for store-backed caches whose TTL lives in the store. One read per send did not justify that.
- **Chat-lane latency guard (2 monotonic reads/message).** Sampling it, or only measuring when the lane has a backlog, both stop reporting the single pathological message the guard exists to catch. A controlled-clock test of it is not reachable either: the monotonic provider is the same process-wide `OnceLock`, which is what pushed #1103 to count at the boundary in the first place.
- **`SessionStats` per-frame and per-event bookkeeping.** #1103 rejected sampling `last_data_received_ms` because two decisions measure elapsed time from it, and rejected deriving the watchdog from tick-observed counters because it pushes worst-case detection from ~50s to ~80s. The new data (four events per message, not one) changes the size of the group, not either argument.

## Guarantees

- `a_send_reads_the_clock_once` asserts exactly one wall read across a send, which is the proof that every stamp shares an instant, and checks the outbound secret really landed inside the measured window.
- `dm_send_stays_within_its_clock_budget` drops its budget from 4 wall reads to 1 and fails if it grows.
- `wire_timestamp_keeps_real_time` still proves a timestamp reaching the server carries the real second.
- `supplied_instant_decides_the_same_bucket_boundary` pins both privacy-token decisions at the exact bucket boundary against a supplied instant.
- The `PortableCache` TTL/TTI boundary test and the dead-socket tests from #1103 pass unchanged.

No embedder has to register a provider, change a provider signature, or keep time state of its own.

## Validation

```
cargo fmt --all --check
cargo test -p wacore --lib
cargo test -p whatsapp-rust --lib
```

Full matrix left to CI.

### DIFF
--- a/src/features/comments.rs
+++ b/src/features/comments.rs
@@ -120,6 +120,7 @@ impl<'a> Comments<'a> {
                 &result.message_id,
                 &comment_secret,
                 wacore::msg_secret::RetentionClass::Text,
+                crate::send::SendInstant::now(),
             )
             .await;
         Ok(result)
--- a/src/handlers/message.rs
+++ b/src/handlers/message.rs
@@ -98,6 +98,9 @@ fn create_chat_lane(client: &Arc<Client>) -> ChatLane {
                     log::debug!(target: "MessageQueue", "Stale worker exiting; remaining messages will be redelivered by server");
                     break;
                 }
+                // Two clock reads per message, kept: sampling or gating on lane
+                // backlog would stop reporting the single pathological message
+                // this guard exists to catch.
                 let start = wacore::time::Instant::now();
                 let client = client_for_worker.clone();
                 // Awaited inline (not boxed): the future lives in this
--- a/src/message/tests.rs
+++ b/src/message/tests.rs
@@ -11354,6 +11354,7 @@ async fn msmsg_outbound_put_and_inbound_get_match_for_lid_bot() {
             outbound_id,
             &secret,
             wacore::msg_secret::RetentionClass::Bot,
+            crate::send::SendInstant::now(),
         )
         .await;
 
--- a/src/request.rs
+++ b/src/request.rs
@@ -183,9 +183,15 @@ impl Client {
     ///
     /// A string containing the generated message ID in the format expected by WhatsApp.
     pub fn generate_message_id(&self) -> String {
+        self.generate_message_id_at(wacore::time::now_secs_u64())
+    }
+
+    /// Same as [`Self::generate_message_id`], but against a caller-supplied
+    /// second (see [`RequestUtils::generate_message_id_at`]).
+    pub(crate) fn generate_message_id_at(&self, unix_secs: u64) -> String {
         let device_snapshot = self.persistence_manager.get_device_snapshot();
         self.get_request_utils()
-            .generate_message_id(device_snapshot.pn.as_ref())
+            .generate_message_id_at(device_snapshot.pn.as_ref(), unix_secs)
     }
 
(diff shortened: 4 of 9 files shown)


jlucaso1:
Took the first of the three, skipped the other two.

`send_message_impl` really was resampling for its fallback id, so the entry points that do not go through `send_message_with_options_inner` kept two reads instead of one. Fixed.

Not taking the tc-token issuance one: `issue_tc_token_after_send` is fire-and-forget on its own task, after the send returned. Its `t=` goes on the wire and the server echoes it, and the stored sender timestamp drives the bucket rate limit. Stamping that IQ with the send instant would make a wire timestamp older than the event it describes, which is reading it worse rather than reading it once. The shared instant is meant to stop where the operation does.

Not taking the comments one either: the comment secret is persisted after `send_message` returns, and `SendResult` does not carry the instant, so reusing it means widening a public non_exhaustive type or duplicating a send entry point for one read off the hot path. The skew it would remove lands in `message_ts`, which feeds retention horizons measured in days.

Also fixed the rustdoc failure in the same push: the module doc linked `SendInstant`, which is crate-private.

PR #1113, opened 2026-07-25
fix(voip/mlow): decode frames coded inactive instead of silencing them

jlucaso1:
Stacked on #1112 (which is stacked on #1111) — it adds the harness that generates the vector this is validated against. Review those first; this diff is the decoder change plus its test.

## Summary

The top field of the smpl TOC is a **two-bit SID/VoA selector**, not a pair of independent flags. The reference documents `00` as *"Normal frame(s) sent while there is no voice activity i.e. when DTX is off"* — a normal frame carrying background noise. Only `10` is a SID.

The decoder treated anything that was not active-voice as silence and returned a zero slot, so a peer running with DTX off had every pause dropped on the floor. It does not sound broken, it sounds like a slightly dead line, which is why it survived this long.

## Evidence

Found while measuring why the 60 ms reference vector only correlates 0.969 — it is not fixture drift, it is this. Splitting that correlation by TOC:

| TOC | samples | corr | rms reference | rms ours |
|---|---|---|---|---|
| `0x10` (coded inactive) | 12480 (12%) | **NaN** | **0.047** | **0.000** |
| `0x12` (VoA=0, hang-over set) | 1920 | 0.9995 | 0.111 | 0.111 |
| `0x50` (active) | 91200 | 0.9856 | 0.092 | 0.092 |

`NaN` because one side is constant: we emit exact zeros where the reference emits noise. With the fix, overall correlation against the reference on a DTX-off stream goes from **0.969 to 0.999914**.

Note `0x12` shares `VoA=0` but sets the hang-over bit, which the reference folds into `coded_as_active_voice` — it was already decoding correctly, and still does.

## Changes

- A SID (`0x80`) still routes to silence without opening the range coder. A frame that is merely coded inactive now takes the normal decode path.
- Two symbols are gated on `coded_as_active_voice` in the reference and are **absent from the bitstream** when it is clear: the voicing symbol (such a frame is unvoiced by definition) and the LSF interpolation index. Reading either would consume symbols that were never written and desync every field after them, so `decode_smpl_lsf` takes the flag and skips them.
- The pulse decode selects its table with the real `coded_as_active_voice` rather than a hardcoded `1`, matching `smpl_max_pulses_per_frame[lowRate][coded_as_active_voice + Voiced]`.

## Contract changes, called out

Two existing tests changed, both because they pinned the old behavior rather than a property:

- **`sid_frame_is_silenced_not_dropped`** (was `inactive_off_point_frame_is_silenced_not_dropped`) used TOC `0x00`, which is coded-inactive rather than SID. It now uses `0x80`. What it asserts is unchanged.
- **`decoder_silence_frames_produce_zero`** asserted over every near-silent frame in `e2e_vectors.json`. Per `PROVENANCE.md`, that fixture had its inactive-TOC frames *"zeroed to match DTX routing"* when it was generated — it was adjusted to the behavior this PR removes, so those entries cannot serve as an oracle for silence. The assertion now covers real SID frames, and the coded-inactive path is covered against an unmodified vector instead.

The second one is worth pausing on: a fixture was edited to agree with the decoder, and that is what kept the bug invisible. The new vector is generated end to end by the reference with nothing zeroed.

## Validation

- `dtx_off_frames_decode_to_audio` — the regression, over a 60 ms DTX-off stream generated by the reference. Fails before this change with `0.00000 rms against the reference's 0.04658`.
- `multi_frame_decode_matches_the_reference` and `e2e_decode_matches_usesmpl` both still pass, so neither the 120 ms path nor the existing capture regressed.

```
cargo fmt --all --check
cargo test -p wacore --features voip-mlow --lib   # 1561 passed
```

Full matrix left to CI.

## Scope

Inbound only. The encoder always codes active voice, so nothing about what we transmit changes.


jlucaso1:
Folded into #1111 so the bots review the whole MLow change set together — the stacked bases were making CodeRabbit skip this one, and the three findings are easier to judge as one story. Same commits, same review points (the short-read failure and the pinned oracle revision are in).

PR #1114, opened 2026-07-25
perf: cut two per-message allocations on the pingpong hot path

jlucaso1:
Two allocation cuts on the pingpong hot path, found by profiling the harness scenario (120k messages at 12k/s) with the dhat, cpu, rss and tokio passes.

## What changes

**1. `sent_messages` eviction stops cloning every key.** When the in-memory store hit its 4096 entry cap it cloned each retained key (two Strings) into a Vec and sorted the whole map, holding the state lock throughout, once per 1024 inserts. It now collects only the timestamps, picks the cutoff with `select_nth_unstable` (O(n), no full ordering) and trims with a counted `retain`.

The counter is load bearing: under a flood every entry shares the same second, so the cutoff timestamp is the same value the whole map carries. A plain `timestamp <= cutoff` retain would empty the store instead of trimming it, dropping the retry/receipt payloads of messages that were just sent. The new test pins that case.

**2. The ratchet updates the chain key in place.** `set_sender_chain_key` and `set_receiver_chain_key` built a fresh `ChainKey` and boxed it into the message field, allocating once per sent and once per received message for a struct holding an index and 32 bytes. They now overwrite the existing box; the `None` arm keeps the previous behaviour for a chain with no key yet.

## Measurements

Harness `pingpong`, 120k messages at 12k/s, `MODE=rss` for allocation counts and `MODE=normal` for CPU/latency. Two interleaved A/B rounds (main, branch, main, branch) against `ee9cb5f8`, both sides built from the same worktree.

| metric | main | branch | delta |
|---|---|---|---|
| allocations per message | 199.7 | 184.8 | -7.5% |
| RSS peak | 77.5 MB | 69.8 MB | -10% |
| allocator churn | 30.4 KB/msg | 29.8 KB/msg | -2 to -6% |
| CPU total | 10.10 s | 10.20 s | no detectable change |
| pong average | 0.885 ms | 0.845 ms | within noise |
| throughput / lost / pong_failed | 120000 / 0 / 0 | 120000 / 0 / 0 | unchanged |

Call site proof from the dhat pass (20k messages): `store_sent_message` at 131072 blocks and `Box<ChainKey>::new` at 60056 blocks both disappear, while every other call site stays put (transport 6.39 -> 6.40 per message, spawn 5.97 -> 5.96, `read_attributes` 5.07 -> 5.07).

Being straight about the CPU column: malloc and free are only around 7% of this workload's profile now, so a 7.5% cut in allocation count predicts well under 1% of CPU, which is below the run to run noise. The gain here is memory and lock hold time, not throughput. The RSS drop also comes with visibly lower variance across runs (69-70 MB versus 72-83 MB).

## Verification

`cargo fmt --all`, `cargo clippy --all-targets` clean, `cargo test --workspace --exclude e2e-tests` green (50 suites), including the 181 libsignal unit tests plus `counter_lease` and `session_divergence`.

### DIFF
--- a/wacore/libsignal/src/protocol/state/session.rs
+++ b/wacore/libsignal/src/protocol/state/session.rs
@@ -444,19 +444,27 @@ impl SessionState {
         next_chain_key: &ChainKey,
     ) -> Result<(), InvalidSessionError> {
         use bytes::Bytes;
-        let chain_key = session_structure::chain::ChainKey {
-            index: Some(next_chain_key.index()),
-            key: Some(Bytes::copy_from_slice(next_chain_key.key())),
-        };
-
-        let mut new_chain = self
+        let chain = self
             .session
             .sender_chain
-            .take()
+            .as_option_mut()
             .ok_or(InvalidSessionError("missing sender chain"))?;
-        new_chain.chain_key = MessageField::some(chain_key);
 
-        self.session.sender_chain = MessageField::some(new_chain);
+        // Overwrite the boxed ChainKey in place. Ratchet advance runs once per
+        // sent message, so allocating a fresh Box for a two-field struct was
+        // per-message churn for no state change beyond these two fields.
+        match chain.chain_key.as_option_mut() {
+            Some(existing) => {
+                existing.index = Some(next_chain_key.index());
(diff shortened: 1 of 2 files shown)


jlucaso1:
@codex review

PR #1144, opened 2026-07-27
fix: six chat-store and send-path issues

### LINKED ISSUE #1132: send: every named `<biz>` classification is rejected by the server, only `mixed` still delivers
## Summary

`classify_button` gives fourteen native-flow names a bespoke `<biz>` shape. Against live servers today every one of those shapes is rejected, and only the `mixed` fallback still delivers. The table now breaks precisely the names it knows best, while any name it does not recognise works.

## Evidence

Sent 2026-07-27 from a WhatsApp Business account (linked companion device) to a consumer Android handset. Fifteen probes, confirmed on both ends: the sender's ack/status and the receiving account's own history agree on every one.

| Classification | Probes | Result |
|---|---|---|
| `mixed` | 10 | 10 delivered |
| `NestedNamed(..)` | 4 | 4 rejected |
| `PaymentSimple(..)` | 1 | 1 rejected |

Rejected under their own branch: `cta_url` (473), `call_permission_request` (473), `payment_info` (473), `open_webview` (405), `galaxy_message` (405).

## The controlled comparison

`infer_biz_node` derives the stanza from the **first** button's name, so putting a `quick_reply` first re-classifies an otherwise byte-identical message to `mixed`:

- `[cta_url]` -> 473 nack, never delivered
- `[quick_reply, cta_url]` -> delivered, read, and the URL button renders and opens

Same button, same `buttonParamsJson`, same recipient, same session, same second. Only the `<biz>` shape differs.

The same holds for `call_permission_request`, `open_webview` and `galaxy_message`: each is rejected alone and delivers behind a leading `quick_reply`, and each then renders its real UI on the handset (a native call-permission card, an "Open page" button, and the full-screen Flow container respectively). So the flows themselves are alive and only the classification is refused.

## Consequence for callers

A caller must know to lead with a throwaway `quick_reply`, which is invisible in the API. For `call_permission_request` that also costs a button: the permission card takes over the bubble and suppresses the sibling quick reply.

## Suggested fix

For the eight `NestedNamed` names the evidence supports collapsing to `BizCategory::Mixed` outright.

Two caveats I cannot resolve from here:

- I tested only one of the six `PaymentSimple` names (`payment_info`). It behaved identically (473), but a merchant-provisioned account on real payment rails may legitimately differ, so those may deserve a separate look rather than the same collapse.
- An alternative to deleting the named form is that it has simply gone stale in a smaller way. `build_nested_biz` pins `v="9"` on `<native_flow>`; it is possible the named shape still works at a newer `v` and only the version is wrong. I have no way to enumerate valid versions, but it would be cheap to try if you can.

## Environment

Present at rev `54272f37` and unchanged at `main` (`1118084d`). Account type: WhatsApp Business. Recipient: consumer WhatsApp on Android.

### LINKED ISSUE #1133: send: `infer_biz_node` ignores `carouselMessage`, so a carousel is delivered but never rendered
## Summary

`infer_biz_node` matches only `InteractiveMessage::NativeFlowMessage`:

```rust
let wa::message::interactive_message::InteractiveMessage::NativeFlowMessage(nf) =
    interactive.interactive_message.as_ref()?
else {
    return None;
};
```

A `carouselMessage` therefore gets no `<biz>` node and no `<bot biz_bot="1"/>`, and the recipient's client renders nothing at all.

## Why this is worse than a rejection

The message is **accepted**. No nack. Status goes `delivered`, then `read`. The receiving account's own linked devices hold it and classify it as an interactive message. The handset shows nothing: no bubble, no "unsupported message" placeholder, nothing after a full chat reload.

Every signal available to the sender says success.

Contrast `listMessage`, the other proto that reaches the server without a `<biz>` node: that one is refused `405`, which at least surfaces as an error.

## Evidence

2026-07-27, Business account -> consumer Android handset. Carousel of three cards, each with an image header, body, footer and one `quick_reply`.

- Built as the engine builds it today: delivered, read, invisible.
- Byte-identical carousel with the engine's own `mixed` `<biz>` shape supplied through `SendOptions::extra_stanza_nodes`: **renders completely** - horizontally scrollable cards, each showing its image, title, body, footer and button.

The missing stanza is the sole cause.

## Suggested fix

Let `infer_biz_node` handle the carousel variant. A carousel's buttons live on its cards rather than at the top level, so the natural analogue of the existing rule is the first button of the first card - though given #1132, `mixed` is the only shape the server currently accepts anyway.

Worth noting the `<bot biz_bot="1"/>` node matters here too and is emitted from the same branch, so a carousel currently also misses the AI/business marking that every other interactive DM carries.

## Environment

Present at rev `54272f37` and unchanged at `main` (`1118084d`).

### LINKED ISSUE #1134: chat-store: same-second messages sort by msg_id, placing a reply above the message it answers
## What happens

In a live conversation, a reply renders *above* the message it answers. It is not intermittent jitter: the same pair inverts every time, and in a fast back-and-forth every one of the local device's messages slides one position earlier than it should.

Reproduced on `54272f37` against a real linked-device session, cross-checked against what the peer's phone displays for the same two messages.

## Why

`t` from the server is whole seconds, so every `timestamp_ms` in the store ends in `000`. Two messages exchanged inside the same second are not merely close, they are byte-identical, and the sort falls through to its tiebreak:

```rust
.order((dsl::timestamp_ms.desc(), dsl::msg_id.desc()))
```

`msg_id` encodes nothing about time. Worse, it is **biased in a fixed direction**, because this library stamps a constant prefix on everything it sends (`wacore/src/request.rs`, `message_id_at`):

```rust
// WA Web message IDs are "3EB0" + 18 hex chars (9-byte truncated hash)
id.push_str("3EB0");
```

Outgoing ids therefore always begin with `'3'`. Inbound ids come from the peer's client and are effectively uniform hex. Sorting descending, an inbound id wins whenever its first character exceeds `'3'` — 12 of 16 values, so roughly 75% of same-second ties, always in the same direction.

That is why it presents as a consistent one-position shift rather than random noise.

## Evidence

Two messages, one from the peer's phone, then a reply from the linked device:

```
timestamp_ms  | from_me | text | msg_id
1785101675000 |       0 | 1st  | ACF59A7CB022679C9C44060A10C25026
1785101675000 |       1 | 2nd  | 3EB025E0465016A858A333
```

Identical timestamps. `'A' > '3'`, so `1st` sorts as the newest and the thread renders `2nd` above `1st`. The peer's phone shows `1st` then `2nd`.

The correct order was in the table the whole time, in insertion order:

```
rowid 123  1st   <- inbound, inserted first
rowid 124  2nd   <- the reply, inserted second
```

Reordering the same rows by `rowid` reproduces the phone's order exactly.

## Suggested fix

Break the tie on **insertion order** rather than on the id. That is the order the socket delivered the messages, which is the question the tiebreak is actually trying to answer, and SQLite already records it.

Four sites sort this way:

- `storages/chat-store/src/lid.rs:158`
- `storages/chat-store/src/queries.rs:208`
- `storages/chat-store/src/store.rs:1402`
- `storages/chat-store/src/store.rs:1432`

The last two are `newest_chat_head` / the chat-preview update, so the same tie also picks the wrong "newest" message for a chat-list preview.

**This is not only an `ORDER BY` change.** Keyset pagination in `queries.rs::messages` uses a filter that must match the sort exactly:

```rust
dsl::timestamp_ms.lt(cursor.timestamp_ms)
    .or(dsl::timestamp_ms.eq(cursor.timestamp_ms)
        .and(dsl::msg_id.lt(&cursor.msg_id)))
```

If the sort changes and the filter does not, pages will skip or repeat rows at a tie boundary. `MessageCursor` (`types.rs:175`) would need to carry the same discriminator, which is a breaking change to a public type.

## Notes on the approach

- `messages` is a normal rowid table (composite PK, not `WITHOUT ROWID`), so the counter already exists at no storage cost.
- Rowid reuse after a delete does not break within-chat ordering: SQLite assigns `max(rowid)+1` over surviving rows, so a new message still outranks older ones in its chat. Worth stating explicitly given `store.rs:1084`, `lid.rs:289` and `delete_chat_rows` do delete.
- If depending on `rowid` is unattractive, an explicit monotonic column is equivalent for ordering purposes and immune to the reuse question, at the cost of a schema change.
- History sync inserts older messages later and therefore out of rowid order, but those carry distinct older timestamps, so the tiebreak never fires for them.
- For reference, WhatsApp Web has the same whole-second `t` and orders on its own local arrival sequence rather than on message ids.

Happy to send a PR if the rowid direction (and the `MessageCursor` change it implies) looks right to you.

### LINKED ISSUE #1140: chat-store: live inbound batches are materialized twice when a durability hook already committed them
## Summary

On a client with an inbound durability hook, every LIVE inbound message is materialized into the chat-store twice: once by the hook before the transport ack, and again by the `ChatStoreHandler` when `Event::Messages` is dispatched. The second pass is not a cheap idempotent no-op - because the inbound path passes `overwrite: true`, the conflicting insert falls through to a full `UPDATE` of the proto blob, which fires the FTS `messages_fts_au` trigger (an FTS5 delete plus insert), and `bump_chat` runs again. It also emits a second `Chats` + `Messages{chat}` `StoreChange` pair, so every subscriber re-queries every surface twice per message.

Measured statically at rev `d2a3a16a`, the duplicate pass is roughly 40% of the per-message write cost on the session DB's single permit, and exactly 2x the store-change fan-out.

## Details

- Live traffic commits as a batch of one through the durability hook (`src/message/commit_batch.rs:9-10`); the host's hook enqueues the batch and flushes, so by the time `event_bus.dispatch` runs at `src/message/commit_batch.rs:852`, the rows are already committed.
- The chat-store's own handler then re-applies the same batch: `apply_inbound` -> `insert_message` (`storages/chat-store/src/store.rs:1155-1180`, `:2108-2154`) with `overwrite: true` (`:1175`), so the conflict arm rewrites `kind`/`text_content`/`proto` (`:2136-2149`), the FTS trigger runs (`storages/chat-store/src/fts.rs:52-56`), the row reports `Refreshed`, and `bump_chat` (`:2170-2183`) plus both change flags fire again.

## Why the handler cannot just skip `Event::Messages`

Two producers dispatch `Event::Messages` while deliberately bypassing the commit pipeline, both tagged `BatchOrigin::Live`:

- newsletters (`src/message/dispatch.rs:53-60`)
- PDO-recovered messages (`src/pdo.rs:419-423`)

For those, the handler's materialization is the ONLY write, so "already committed" is not representable in `BatchOrigin` today.

## Proposal

Add a marker on `MessageBatch` distinguishing hook-committed from hook-bypassed batches - a third `BatchOrigin` variant (e.g. `LiveCommitted`) or a `hook_committed: bool` - set by the commit pipeline when a durability hook has already run `commit()` for the batch. The `ChatStoreHandler` skips materialization for hook-committed batches and applies everything else unchanged. Hosts without a hook see no behavior change; hosts with one save the duplicate UPDATE + FTS churn and the doubled `StoreChange` fan-out.

Happy to PR this if the shape sounds right.

### LINKED ISSUE #1141: chat-store: chats() ORDER BY cannot use idx_chats_order, and there is no cursor or point lookup for the chat list
## Summary

`queries::chats()` orders by `(pinned_at IS NULL, pinned_at DESC, last_message_ts DESC)` (`storages/chat-store/src/queries.rs:162-166`), but the only ordering index is `idx_chats_order (device_id, archived, last_message_ts DESC)` (`migrations/2026-07-09-000000_chat_store_initial/up.sql:27`). The sort cannot be satisfied from the index, so every `chats()` call is a full scan of the chats table plus a temporary B-tree sort.

There is also no cursor/offset form: `chats(include_archived, limit)` always sorts the whole table even when a caller wants one page - and, worse, a caller that needs to FIND one chat (map an addressed jid back to a store key, fold one chat's unread) has no point-lookup alternative, so embedders end up calling `chats(_, i64::MAX)` on hot read paths.

## Asks

1. Align the index with the real sort so the common non-pinned case can stream from the index - e.g. `(device_id, archived, pinned_at DESC, last_message_ts DESC)`, or split the query into a pinned pass and a non-pinned pass, whichever profiles better. (SQLite cannot serve the `pinned_at IS NULL` leading expression from a plain column index, so the two-pass shape may be the honest fix.)
2. A keyset cursor for the chat list (`before (pinned_at, last_message_ts, jid)` style, mirroring what `messages()` already does with `(timestamp, msg_id)`), so paging the chat list stops being O(all chats) per page.
3. A point lookup `chat(device_id, jid)` on the public store API (the PK already supports it) so "find one chat by key" stops requiring the full list.

Filed from the waahan embedding, where the chat list, unread folding, and jid->store-key resolution all currently ride full-table `chats()` calls; (3) alone would remove most of them.

### LINKED ISSUE #1142: chat-store: a message-class server ack for a not-yet-recorded outgoing row is silently dropped
## Summary

A message-class server `<ack>` that arrives before its outgoing row exists in the chat-store is silently dropped. `apply_server_ack` only updates an EXISTING `from_me` row: it tries the device-wide unique-id fallback, then one PN/LID counterpart-key retry, and then drops the ack with no log, no stub row, and no deferred re-apply (`storages/chat-store/src/store.rs:1771-1844`, `resolve_server_ack_message` at `:1727-1769`, the silent drop around `:1723` at rev `d2a3a16a`).

Consequence for that message: its status stays `pending` (clock icon) until a delivery receipt happens to lift it - indefinitely for an offline recipient - and its timestamp is never corrected to the server's authoritative `t` (#1127 reconciles timestamps only for rows that already exist).

## How the race arises

`Event::ServerAck` is dispatched synchronously on the socket-read path before any send waiter resolves, and `send_message` returns at the stanza write without waiting for the ack. A host that records the outgoing message into the chat-store after `send_message` resolves (record-after-send is the safe order - recording first leaves a forever-pending ghost row when the send fails, and the store has no row delete) therefore has a window where the ack can beat the insert. In practice the window needs the host's enqueue to lose a race against a full network round trip, so it is rare - but the failure is silent and permanent when it happens, and the window widens for any host with a slower record path.

## Proposal

Hold unmatched message-class acks briefly (keyed by message id, small TTL/cap) and re-apply on the next `Outgoing` insert of that id - the same materialize-later shape the store already uses for out-of-order edits and revokes, which both create placeholders when they arrive before their target. Alternatively, an explicit warn on the drop would at least make the loss observable.

jlucaso1:
Closes #1142, closes #1141, closes #1140, closes #1134, closes #1133, closes #1132.

Two independent clusters: four in `chat-store`, two on the send path. Every claim below was checked against the WA Web bundle (561 JS files at `2.3000.1043873090`, fetched via whatspec's `bundles.lock.json`) or against SQLite's `EXPLAIN QUERY PLAN` before the code was written.

### #1134 — same-second messages sort by `msg_id`, placing a reply above the message it answers

The server's `t` is whole seconds, so every `timestamp_ms` ends in `000` and a live back-and-forth lands several messages on an identical sort key. The tiebreak then fell through to `msg_id`, which encodes nothing about time and is biased in a fixed direction: we stamp a constant `3EB0` prefix on the ids we generate while a peer's are effectively uniform hex, so descending id order put the peer's message on top for ~75% of ties.

The tiebreak is `rowid` now — the order the socket delivered the rows, which is the question the tiebreak was always asking. All four sites move together (`queries.rs`, `lid.rs`, and both chat-head sites in `store.rs`), and `MessageCursor` carries the same discriminator, because a keyset filter that does not mirror its sort exactly skips or repeats rows at a page boundary inside a same-second run.

`idx_messages_chat_time` is rebuilt ASC. An index key carries the rowid as its trailing column, so a reverse scan of the ASC form yields `timestamp_ms DESC, rowid DESC` directly:

```
ASC index:  SEARCH messages USING INDEX idx_messages_chat_time (device_id=? AND chat_jid=?)
DESC index: SEARCH messages USING COVERING INDEX idx_messages_chat_time (...)
            USE TEMP B-TREE FOR RIGHT PART OF ORDER BY
```

**Breaking:** `MessageCursor.msg_id` becomes `seq`, and `StoredMessage` gains `seq`. Build cursors with `MessageCursor::from(&stored_message)` and neither matters.

### #1141 — `chats()` ORDER BY cannot use `idx_chats_order`

`(pinned_at IS NULL, pinned_at DESC, last_message_ts DESC)` cannot be served from any column index, so every call was a full scan plus a temp B-tree — including the calls that wanted one page, or one chat.

The list is two ordered runs concatenated now (pinned by pin time, then the rest by activity), each a plain range scan that stops at `LIMIT`. `archived` leaves the index key and becomes a filter: as a leading column it split the activity order into two runs that the archived-inclusive list then had to merge back, and as a filter both list modes stream from one index. All five access paths are sort-free:

| query | plan |
| --- | --- |
| activity run, hide archived | `SEARCH chats USING INDEX idx_chats_order (device_id=?)` |
| activity run, incl. archived | `SEARCH chats USING INDEX idx_chats_order (device_id=?)` |
| activity run + cursor | `SEARCH chats USING INDEX idx_chats_order (device_id=? AND last_message_ts<?)` |
| pinned run | `SEARCH chats USING INDEX idx_chats_pinned (device_id=? AND pinned_at>?)` |
| pinned run + cursor | `SEARCH chats USING INDEX idx_chats_pinned (device_id=? AND pinned_at>? AND pinned_at<?)` |

Adds `chats_page(include_archived, after, limit)` with a `ChatCursor`, and `chat(&Jid)` for the point lookup the primary key always supported (resolving either of a peer's identities). `chats()` keeps its signature and delegates.

### #1142 — a message ack for a not-yet-recorded outgoing row is silently dropped

Unmatched message-class acks now wait in a bounded, TTL'd queue and re-apply on the `Outgoing` insert that answers them — the same materialize-later shape the store already uses for out-of-order edits and revokes, minus the placeholder row (an ack carries no content to place). Applying it at insert time also corrects the optimistic timestamp to the server's before anything renders the row. Every discard warns, so the loss is never silent again.

The queue is handed to the writer's blocking closure and taken back on both paths, so a rolled-back batch — which applied none of them — leaves them waiting rather than dropping them.

### #1140 — live inbound batches materialized twice when a durability hook already committed them

`MessageBatch` gains `hook_committed`, set by the commit pipeline once the hook's `commit()` has returned. The handler drops those batches before they reach the writer, so they do not take a batch slot either.

It is a separate field rather than a third `BatchOrigin` variant for two reasons: a hook commits live and drain batches alike, so it is orthogonal to the delivery shape `BatchOrigin` documents; and `BatchOrigin` is not `#[non_exhaustive]`, so a new variant would break exhaustive matches. The producers that bypass the commit pipeline — newsletters (`dispatch.rs`) and PDO recovery (`pdo.rs`) — leave it unset, so the handler stays their only materializer, and hosts without a hook see no change at all.

### #1132 — every named `<biz>` classification is rejected, only `mixed` delivers

The eight `NestedNamed` names collapse to `mixed`, per the live probes. The payment family is deliberately left alone: only one of its six names was probed, and a merchant-provisioned account on real payment rails may legitimately answer differently — its flat four-attr shape is also the one form WA Web itself emits attrs-only.

On the open question of whether the named form had merely gone stale at a newer `v`, the bundle answers directly: **it is not a newer `v`, it is no `v` at all.** All three of WA Web's `<native_flow>` builders pass `name` and nothing else. Two further mismatches, same source:

- `createFanoutMsgStanza` builds `<biz>` in one of three *mutually exclusive* shapes. The privacy attrs (`actual_actors` / `host_storage` / `privacy_mode_ts`) belong to an attrs-only form that never has children; the nested form is `wap("biz", null, ...)` with no privacy attrs. Our nested form carries both.
- `<quality_control>` appears exactly once in the entire bundle — in the **incoming** parser. WA Web never sends it.

So `mixed` is plausibly just the one name the server does not validate strictly enough to notice the malformed envelope. That is a hypothesis, not a finding, and correcting the envelope would risk the one path that demonstrably delivers today — so this PR does not touch it, and records the evidence in the code for whoever probes it next.

### #1133 — `infer_biz_node` ignores `carouselMessage`

`infer_biz_node` follows WA Web's rule now (`getNativeFlowNameFromMsg`): when there are no native-flow buttons but the message carries a renderable envelope — body text, header title, footer text, or a header image — and is not a storefront, it announces itself as `mixed`. A carousel's buttons live on its cards, so the button rule never fires and it lands here naturally, which matches what the reporter observed when injecting a `mixed` node by hand.

## Tests

105 chat-store and 1235 lib tests pass locally, with regressions covering each issue: the exact inversion from #1134 plus an exact-pagination sweep across a same-second run; point lookup via both identities and single-row paging across the pinned/activity boundary; an ack (and a nack) beating its own row, and a held ack refusing to be consumed by a different id; and marked/unmarked batch materialization.

One existing test, `same_millisecond_insert_order_does_not_hijack_preview`, asserted the old id tiebreak — it is rewritten to lock the new rule, and joined by a case proving arrival only *breaks ties* and never lets a genuinely older message take the preview.

Heavy workspace clippy is left to CI. `alsa-sys` fails to build in this container for a missing system library, unrelated to these changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PyCszWBEYZoqnTRfHpSQ2X)_

### DIFF
--- a/src/message/commit_batch.rs
+++ b/src/message/commit_batch.rs
@@ -721,6 +721,9 @@ impl Client {
             return false;
         }
 
+        // Rides on the dispatched event so a consumer that the hook already
+        // fed can skip re-materializing the batch.
+        let mut hook_committed = false;
         if let Some(hook) = self.inbound_durability_hook() {
             // Key strings live for the whole commit; rows borrow them and the
             // encode arena, so the batch write allocates nothing per row
@@ -805,6 +808,7 @@ impl Client {
                 );
                 return true;
             }
+            hook_committed = true;
 
             let delete_keys: Vec<PendingInboundKey<'_>> = items
                 .iter()
@@ -853,6 +857,7 @@ impl Client {
             MessageBatch::builder()
                 .messages(items)
                 .origin(origin)
+                .hook_committed(hook_committed)
                 .build(),
         ));
         true
--- a/src/send/mod.rs
+++ b/src/send/mod.rs
@@ -567,16 +567,39 @@ const BIZ_PRIVACY_MODE_TS_OFFSET: u64 = 77_980_457;
 
 enum BizCategory<'a> {
     /// `<biz actual_actors host_storage privacy_mode_ts native_flow_name=X/>` — no children.
+    /// The one shape WA Web also emits attrs-only: `createFanoutMsgStanza`
+    /// builds exactly these four attrs (and never a child) when the peer
+    /// contact carries a `privacyMode`.
     PaymentSimple(&'a str),
-    /// Nested form preserving the button's flow name.
-    NestedNamed(&'a str),
-    /// Nested form with `name="mixed"`. Fallback for buttons the server
-    /// silently drops when sent under their literal name.
+    /// Nested form with `name="mixed"`.
     Mixed,
 }
 
+/// Pick the `<biz>` shape for a native-flow message from its first button.
+///
+/// Only the payment family keeps a literal flow name. Every other name routes
+/// through `mixed`, because live probes (issue #1132: fifteen sends from a
+/// Business account to a consumer handset) found that every one of the named
+/// nested shapes is refused — `cta_url`, `call_permission_request` and
+/// `payment_info` with a 473, `open_webview` and `galaxy_message` with a 405 —
+/// while `mixed` delivered on all ten of its attempts. Leading an otherwise
+/// byte-identical message with a `quick_reply` re-classified it to `mixed` and
+/// it went through, so the name is the only variable.
+///
+/// The name is very likely not the whole story, and the rest is worth a live
+/// probe before anyone reinstates the named form. The WA Web bundle
+/// (`WAWebSendMsgFanout.createFanoutMsgStanza`) builds a `<biz>` in exactly one
+/// of three mutually exclusive shapes, and our nested form matches none of
+/// them: it merges the privacy attrs with the nested child, stamps a `v="9"` on
+/// `<native_flow>` that WA Web never emits (all three of its builders pass
+/// `name` alone), and adds a `<quality_control>` child that appears in the
+/// bundle only in the INCOMING parser. `mixed` may simply be the one name the
+/// server does not validate strictly enough to notice.
 fn classify_button(button_name: &str) -> BizCategory<'_> {
     match button_name {
+        // Untouched by the collapse: #1132 probed only `payment_info` of the
+        // six, and a merchant-provisioned account on real payment rails may
+        // legitimately answer differently than the test account did.
         "payment_info" => BizCategory::PaymentSimple("payment_info"),
         "review_and_pay" => BizCategory::PaymentSimple("order_details"),
         "review_order" | "order_status" => BizCategory::PaymentSimple("order_status"),
@@ -584,44 +607,69 @@ fn classify_button(button_name: &str) -> BizCategory<'_> {
         "payment_method" => BizCategory::PaymentSimple("payment_method"),
         "payment_reminder" => BizCategory::PaymentSimple("payment_reminder"),
 
-        "cta_url" => BizCategory::NestedNamed("cta_url"),
-        "cta_catalog" => BizCategory::NestedNamed("cta_catalog"),
-        "catalog_message" => BizCategory::NestedNamed("catalog_message"),
-        "galaxy_message" => BizCategory::NestedNamed("galaxy_message"),
-        "booking_confirmation" => BizCategory::NestedNamed("booking_confirmation"),
-        "call_permission_request" => BizCategory::NestedNamed("call_permission_request"),
-        "open_webview" => BizCategory::NestedNamed("message_with_link"),
-        "message_with_link_status" => BizCategory::NestedNamed("message_with_link_status"),
-
-        // quick_reply / cta_copy / cta_call / single_select / send_location
-        // and every other unknown name go through the mixed fallback. The
-        // server silently drops messages sent under the literal name for
-        // these buttons.
         _ => BizCategory::Mixed,
     }
 }
 
-/// Derive the `<biz>` stanza child for native-flow interactive messages.
+/// Does this interactive message carry something a client can render on its
+/// own, independent of native-flow buttons?
 ///
-/// Returns `None` when the message has no native-flow interactive payload.
-/// Otherwise returns the assembled `<biz>` node. The caller is responsible
-/// for prepending `<bot biz_bot="1"/>` for DM-bound sends (see
-/// `build_extra_stanza_nodes`).
+/// WA Web's rule verbatim (the `f` term of `getNativeFlowNameFromMsg`): a body,
+/// a header title, a footer, or a header image. `hasMediaAttachment` and the
+/// non-image header media are deliberately not part of it.
+fn has_renderable_envelope(im: &wa::message::InteractiveMessage) -> bool {
+    let non_empty = |text: Option<&str>| text.is_some_and(|t| !t.is_empty());
+
+    if non_empty(im.body.as_option().and_then(|b| b.text.as_deref()))
+        || non_empty(im.footer.as_option().and_then(|f| f.text.as_deref()))
+    {
+        return true;
+    }
+    let Some(header) = im.header.as_option() else {
+        return false;
+    };
+    non_empty(header.title.as_deref())
+        || matches!(
+            header.media,
+            Some(wa::message::interactive_message::header::Media::ImageMessage(_))
+        )
+}
+
+/// Classify an interactive payload into the `<biz>` shape it should carry,
+/// mirroring WA Web's `getNativeFlowNameFromMsg`: the first native-flow
+/// button's name decides when there is one, otherwise a payload that renders
+/// on its own is announced as `mixed`.
+///
+/// That second arm is what makes a carousel work (issue #1133): its buttons
+/// live on the cards, not at the top level, so the button rule never fires and
+/// the message used to leave without a `<biz>` at all — accepted, acked, and
+/// then invisible on the handset. A `shopStorefrontMessage` is excluded, as it
+/// is in WA Web.
+fn classify_interactive(im: &wa::message::InteractiveMessage) -> Option<BizCategory<'_>> {
+    use wa::message::interactive_message::InteractiveMessage as Payload;
+    match im.interactive_message.as_ref()? {
+        Payload::NativeFlowMessage(nf) if !nf.buttons.is_empty() => {
+            nf.buttons.first()?.name.as_deref().map(classify_button)
+        }
+        Payload::ShopStorefrontMessage(_) => None,
+        _ => has_renderable_envelope(im).then_some(BizCategory::Mixed),
+    }
+}
+
+/// Derive the `<biz>` stanza child for interactive messages.
+///
+/// Returns `None` when the message has no interactive payload, or one that
+/// announces nothing (a storefront, or a payload with neither buttons nor any
+/// renderable envelope). Otherwise returns the assembled `<biz>` node. The
+/// caller is responsible for prepending `<bot biz_bot="1"/>` for DM-bound
+/// sends (see `build_extra_stanza_nodes`).
 ///
 /// `now_unix_secs` is the current wall-clock time in unix seconds. Taking it
 /// as a parameter keeps the function pure and lets tests pin the resulting
 /// `privacy_mode_ts` deterministically without touching the global time
 /// provider.
 fn infer_biz_node(msg: &wa::Message, now_unix_secs: u64) -> Option<Node> {
-    let interactive = extract_interactive_message(msg)?;
-    let wa::message::interactive_message::InteractiveMessage::NativeFlowMessage(nf) =
-        interactive.interactive_message.as_ref()?
-    else {
-        return None;
-    };
-
-    let first_button_name = nf.buttons.first()?.name.as_deref()?;
-    let category = classify_button(first_button_name);
+    let category = classify_interactive(extract_interactive_message(msg)?)?;
     let privacy_mode_ts = now_unix_secs
         .saturating_sub(BIZ_PRIVACY_MODE_TS_OFFSET)
         .to_string();
@@ -633,7 +681,6 @@ fn infer_biz_node(msg: &wa::Message, now_unix_secs: u64) -> Option<Node> {
             .attr("privacy_mode_ts", &privacy_mode_ts)
             .attr("native_flow_name", flow_name)
             .build(),
-        BizCategory::NestedNamed(flow_name) => build_nested_biz(&privacy_mode_ts, flow_name),
         BizCategory::Mixed => build_nested_biz(&privacy_mode_ts, "mixed"),
     })
 }
@@ -4142,24 +4189,26 @@ mod tests {
             }
         }
 
-        /// Named-nested buttons keep their flow name and gain the new
-        /// privacy attrs plus `<quality_control>`.
+        /// Every non-payment button name announces itself as `mixed`. The
+        /// eight names that used to keep their own flow name are the ones
+        /// #1132 measured as universally refused (473/405), so they must not
+        /// regain a bespoke shape without fresh live evidence.
         #[test]
-        fn nested_named_form() {
-            let cases: &[(&str, &str)] = &[
-                ("cta_url", "cta_url"),
-                ("cta_catalog", "cta_catalog"),
-                ("catalog_message", "catalog_message"),
-                ("galaxy_message", "galaxy_message"),
-                ("booking_confirmation", "booking_confirmation"),
-                ("call_permission_request", "call_permission_request"),
-                ("open_webview", "message_with_link"),
-                ("message_with_link_status", "message_with_link_status"),
+        fn formerly_named_buttons_now_route_through_mixed() {
+            let cases: &[&str] = &[
+                "cta_url",
+                "cta_catalog",
+                "catalog_message",
+                "galaxy_message",
+                "booking_confirmation",
+                "call_permission_request",
+                "open_webview",
+                "message_with_link_status",
             ];
-            for (button, expected_flow) in cases {
+            for button in cases {
                 let biz = infer_biz_node(&msg_with_native_flow_button(button), FIXED_NOW)
                     .unwrap_or_else(|| panic!("{button}: should produce biz"));
-                assert_nested_biz(&biz, expected_flow, button);
+                assert_nested_biz(&biz, "mixed", button);
             }
         }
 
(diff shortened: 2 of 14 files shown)


jlucaso1:
Pushed `9c546b2` addressing the review. Five fixes, two skips, all with reasoning.

## Fixed

**The hook-committed skip drops messages for unrelated hooks** (Greptile P1, CodeRabbit, Codex P1)

Correct, and the most serious finding here — this was a data-loss regression, not a nit. `hook_committed` says *a* hook committed the batch, not that it committed it into this store. My handler treated the two as the same thing.

The skip is opt-in now via `ChatStore::skip_hook_committed_batches(true)`, and the default materializes everything exactly as before. A host whose hook writes elsewhere sees no change; a host whose hook feeds this store opts in and gets #1140's savings. Only the host can answer that question, so it asks the host.

**A rolled-back batch permanently ate the acks it consumed** (Greptile P1, CodeRabbit critical, Codex P2)

Also correct, and it reintroduced #1142's failure one layer down: `take_matching` mutates a `VecDeque`, which does not roll back with SQLite. The queue is snapshotted before the transaction and restored on error. Deferred acks are rare, so the common snapshot is of an empty queue.

**An ambiguous ack got deferred instead of dropped** (CodeRabbit major)

Good catch. `resolve_server_ack_message` returned `None` both for "no row yet" and for "this id matches several outgoing rows, refusing to guess" — collapsing them turned a deliberate refusal into a 60-second deferral that the next row claiming that id would inherit. Split into `Resolved` / `NotYet` / `Ambiguous`; only `NotYet` defers. Regression test: `ambiguous_ack_is_dropped_rather_than_deferred`.

**Deferred acks keyed by id alone** (Codex P2)

Right. Ids are sender-chosen and unique only within a chat, so chat A's ack could be consumed by chat B's insert and then applied to B through the device-wide fallback. A deferral now carries the storage key its ack named and only matches an insert into that chat. A chatless ack still matches on the id alone — that is the same basis its own resolution falls back to.

**Pinned run lost the activity tiebreak** (Codex P2)

Right, and I introduced it. The old combined sort ranked equally-pinned chats by activity; mine settled them on `jid`. Since history-sync pin times are second-precision and collide, a stale pinned chat could outrank a more recently active one. Ordering is `(pinned_at, last_message_ts, jid)` again, the cursor carries all three, and the partial index matches — still no temp B-tree:

```
SEARCH chats USING INDEX idx_chats_pinned (device_id=? AND pinned_at>? AND pinned_at<?)
```

Plus `chat()` gained the list's `jid` tiebreak (CodeRabbit minor), so the point lookup and the list cannot disagree about which row of a split pair is the thread.

## Verified, then documented rather than changed

**"`rowid` is unsafe because VACUUM renumbers"** (CodeRabbit major) — half right, and the half that matters is fine. VACUUM rewrites the table *in rowid order*, so the relative order this sort depends on survives. I checked rather than assumed:

```
before: [(1,'m0'), (3,'m2'), (5,'m4'), (6,'m5')]     # after deleting m1, m3
after : [(1,'m0'), (3,'m2'), (5,'m4'), (6,'m5')]
relative order preserved: True
```

What does not survive is a `seq` value persisted across a VACUUM — documented on the field. The FTS index has the identical dependency and `fts.rs` already documents it, so this is consistent with how the crate already treats rowid. An explicit column would need a counter on every insert to buy back only the case of a cursor persisted across a maintenance operation it is not meant to survive.

**"`eq_any` still pays a temp B-tree for aliased chats"** (CodeRabbit trivial) — entirely correct, and my migration comment overclaimed. Confirmed:

```
one key : SEARCH messages USING INDEX idx_messages_chat_time (device_id=? AND chat_jid=?)
two keys: SEARCH messages USING INDEX idx_messages_chat_time (device_id=? AND chat_jid=?)
          USE TEMP B-TREE FOR ORDER BY
```

Comment qualified. The index shape is unchanged — it is still right for the single-key case, which is every chat once its PN/LID pair reconciles.

**"`seq` is insertion order, not send order"** (Codex P2) — a real limitation, worth stating plainly. Inbound rows are inserted as the socket delivers them, but an outgoing row is inserted when the host calls `record_outgoing`, after its send resolves. A peer reply decrypted and materialized inside that gap, landing on the same whole second, would take the lower `seq`.

That needs a full network round trip to complete within one second while a local enqueue is still pending — the mirror of the race #1142 describes as needing "the host's enqueue to lose against a full network round trip" — and it is bounded to same-second pairs. Against that: the ordering it replaces is wrong for ~75% of same-second pairs, always in the same direction, which is what #1134 reported as a reproducible defect. Fixing the residue properly means a discriminator captured at the send boundary, which is a new column threaded through the public `record_outgoing` signature — a larger change than these issues asked for. Documented on `StoredMessage::seq`; happy to open a follow-up if you want it closed.

## Tests

21 unit + 92 integration, all passing. New coverage for the review findings: TTL expiry, `DEFERRED_ACK_CAP` eviction, snapshot/restore, chat-scoped matching, chatless-ack matching, the ambiguous drop, and hook-committed batches both opted-in and default.

---

jlucaso1:
Note on the red **Semver Checks (informational)** — it is not caused by this branch, and I checked rather than assumed.

The job only ever gets to one crate:

```
Summary semver requires new major version: 10 major and 0 minor checks failed
Finished [  63.412s] waproto
```

All 62 flagged items are generated protobuf types in `waproto` (`AiRichResponse*`, `AdvDeviceIdentity`, `LidMigrationMapping`, `message::UrlMetadata`, …), compared against the published `waproto-0.6.0`. Nothing from `wacore`, `whatsapp-rust`, or the chat-store appears in the failure list, and this branch does not touch `waproto` at all — it reproduces on `main`.

Worth being explicit about why *my* breaking changes are absent from it, since this PR does have some:

- `MessageBatch::hook_committed` is an added field on a `#[non_exhaustive]` struct, so it is not a break.
- `MessageCursor.msg_id` → `seq` and the new `StoredMessage.seq` are genuine breaks, but they live in `whatsapp-rust-chat-store` (0.1.0, unpublished), so there is no baseline to compare against and the tool never sees them. They are called out under #1134 in the PR description instead.

The check is informational by design and says so itself: *"Breaking changes detected. This does not block the PR — bump the minor version if the break is intended."* The `waproto` drift is worth a separate look at release time, but it is not this PR's to fix.

Everything else on `bcf96a0` is green so far, including Clippy — which is the one gate I could not run in full locally (`alsa-sys` has no system library in my container).

---

jlucaso1:
Both Codex P2s on `bcf96a0` describe real behaviour. I'm skipping both, and pushed `99b2c1c` stating each limit at the point where it is decided — they were true before, just only implied.

**Chatless acks are order-dependent under host id reuse** (`store.rs`)

Correct as described: a deferral with no chat matches on the id alone, so if a host reuses one id across two chats in a single batch, the first insert can take an ack the second would have made ambiguous.

Two reasons not to restructure for it. It is not introduced here — the undeferred chatless path has always resolved against the rows that exist when it runs, so a duplicate insert arriving later never reopened the question either. And the proposed fix, holding acks until every insert in the batch is visible, trades the writer's in-order application of events for a case that requires the host to break id uniqueness first. The library's own generator makes collisions negligible; this needs a caller-supplied id deliberately reused across threads.

Worth noting the named case, which is the one that can happen without a host bug, *is* fixed — `bcf96a0` scopes it to the peer's own PN/LID pair.

**`chat()` does not merge a split pair's metadata** (`queries.rs`)

Also correct: while a PN/LID pair is split, pin/mute/archive/name can sit on the row `chat()` does not return.

But `chats()` has the same property — it lists a split pair as *two* entries and synthesizes nothing. Unioning in the point lookup alone would make it disagree with the list in a new way, and it would put `merge_chat_metadata`'s write-path precedence rules inside a query. The store's answer to a split pair is reconciliation, which merges the two rows for every surface at once; a read path quietly presenting a merge the table does not contain would paper over the state that reconciliation exists to clear.

This is also in tension with the CodeRabbit finding I did take: that one asked `chat()` to settle its tiebreak the same way the list does, precisely so the two surfaces name the same row. Both are now consistent — one row, same choice, on both surfaces.

Happy to revisit either with a concrete failing case.

---

PR #1149, opened 2026-07-27
fix(signal): stop a DH ratchet stranding the counter lease

### LINKED ISSUE #1146: Recurring own-device session corruption ("implausibly far ahead") and retry receipts failing with "backend store error" on fresh messages
**Setup**: fleet of 17 independent accounts running headless group-reply bots on this crate. Git pin via `[patch.crates-io]` (all workspace crates on the same rev), SqliteStore (one `whatsapp.db` per account), tokio transport, Debian slim containers, large groups (several hundred devices each). Observed across revs `d37b1756` (post #1043), `ffed9e8d` (post #1059, includes #1048) and `99d9aa89` (post #1137).

## Symptom 1: session row toward the account's own primary device rots, bot goes deaf to its owner

Every few days one of the accounts starts failing like this:

- inbound messages from the owner's primary phone stop decrypting, with the counter error saying the message is "implausibly far ahead";
- every group send logs an encrypt failure toward one single address, always the account's own lid with device 0 (`<owner-lid>@lid`, device `.0`), while the other several hundred devices in the group encrypt fine.

From the owner's point of view the bot simply ignores their commands. We had five separate accounts hit within one week in late July. The most recent case (Jul 24) happened on `ffed9e8d`, which already includes #1048, so whatever that closed, this vector is still alive.

Workaround that cured every case with no re-pairing and no user action:

1. stop the container;
2. back up the db, then delete exactly one row: `DELETE FROM sessions WHERE address = '<owner-lid>.0';`
3. start again. The session renegotiates by itself within about 3 minutes and stays clean afterwards.

One detail that may help: the rotten session blob is consistently ~1096 bytes, while the freshly renegotiated one is ~414 bytes. We kept backups of the rotten rows.

## Symptom 2: retry receipt for a message sent half a second earlier fails with "backend store error"

On the busiest accounts (notably the same ones that suffered symptom 1) many group replies produce this sequence:

```
15:30:44.559 app: reply sent, id=3EB01B6254FFBC8627472B
15:30:45.032 WARN whatsapp_rust::receipt: Failed to handle retry receipt for 3EB01B6254FFBC8627472B: backend store error in backend
15:30:47.142 app: same id, server ack 267ms, delivery receipts arriving normally
```

Some device in the group asks for a retransmission of a message we sent 0.5s earlier and serving that retry dies in the store layer. The message itself is fine for the rest of the group (server ack, delivery receipts and group confirmations all normal), so the device behind the failed retry presumably stays on the permanent "waiting for this message" placeholder.

Rates are identical on `ffed9e8d` and `99d9aa89` (we ran both on the same day on the same accounts), so this is not a regression from the recent perf work; it looks older.

We also see bursts of the same warning for pre-restart message ids in the first minutes after a container restart (80 to 110 within 20 minutes on busy accounts). If retransmission is served from an in-memory cache, that part is probably expected after a restart; the fresh-id case above is the one that looks wrong.

## Hypothesis

The overlap between the two symptoms makes us suspect a single cause: one peer device with a corrupted session row. That device cannot decrypt anything we send, requests a retransmission every time, and the retry handler fails on the same rotten row with the store error. Symptom 1 would be the special case where the rotten row is the one toward our own account's primary device, which is also the loudest one because it makes the bot deaf to its owner.

## Questions

- Is this corruption vector known? Was #1048 expected to close it? We got a fresh case on a rev that includes it.
- For the failed retry receipts, would a debug log of the requesting participant help pin the device down? We can build with extra logging if you point us at the right spot.
- We can share redacted rotten-row blobs, full timelines and db dumps privately if useful.

jlucaso1:
Fixes #1146, and explains both symptoms reported there as one cause.

## The bug

The sender-chain counter lease is a **record-level** ceiling. The chain it bounds is **per-ratchet-epoch**: a DH ratchet derives a replacement sender chain from a fresh random ephemeral, overwrites the old one in place at counter zero, and drops it without archiving. The ceiling was left describing the retired chain.

For ping-pong traffic that gap is one batch and nobody notices. For a peer we only ever monologue at, it is not: **the account's own primary device gets a copy of every message we send and rarely replies**, so the chain climbs to a few thousand before one reply arrives, and that reply strands the ceiling thousands of counters above a chain that just restarted at zero.

Nothing fails yet — a live reload skips the fast-forward entirely. It fails on **recovery**: a restart, or any lossy cache reset, which is what changes the store incarnation. The reload then has to fast-forward across a gap no send ever created, refuses it at `MAX_RESERVATION_FAST_FORWARD`, and fails the record load with `reserved sender chain index implausibly far ahead` — the exact string in the report.

From there the address is stranded, because **every path that could repair the session has to load it first**: inbound decrypt, the group-send fan-out, and the retry receipt handler all fail with `backend store error in backend`. Meanwhile `has_session` checks row existence without decoding, so it keeps reporting the session present and no rebuild is ever triggered. Deleting the row by hand was the only way out, which is exactly the workaround in the issue.

That accounts for both reported symptoms:

- **Symptom 1** — the rotten row is toward `<owner-lid>@lid` device 0, the one peer that fits the monologue shape, and every group send logs an encrypt failure against that single address while the other several hundred devices are fine.
- **Symptom 2** — a retry receipt for a message sent moments earlier dies in the store layer, because serving it means loading the same unloadable record.

The reporter's own hypothesis in the issue — one peer device with a corrupted session row, with symptom 1 as the loud special case — is correct. This is the mechanism behind it.

## The fix

**1. Rebase the lease when the chain is retired.** `rebase_lease_after_sender_chain_reset` lowers the ceiling back to one batch as part of the same mutation that swaps the chain, so no snapshot can pair the retired chain with the rebased ceiling. It only ever lowers (`min`), so a counter is never published under a ceiling that is not yet durable. It keeps one batch rather than dropping to zero, so the fresh chain's first sends stay lease-covered and the write-behind send path is unchanged.

An archived-state promotion that ratcheted now takes `promote_fresh_state`, which burns the outgoing chain before resetting, instead of having a stale ceiling burned into a chain that never spent anything.

**2. Let an already-damaged database heal.** The cache reports an undecodable session row as absent rather than surfacing a load error, so the ordinary no-session recovery fetches a pre-key bundle and overwrites it. This is what repairs the fleets that are already broken, since fix 1 only prevents new cases. A record that cannot be decoded derives no key material, so it cannot repeat a counter either. `wa_session_record_quarantined_total` counts them; steady state is zero, and a non-zero rate is worth investigating rather than ignoring.

## Why the tests never caught it

The durability chaos state machine only ever created fresh sessions and advanced chains. It had no DH-ratchet action, so the interaction between an in-place chain swap and a record-level lease was never exercised. That action is added here.

## Verification

Four new tests, all failing without fix 1:

| Test | Without the fix |
|---|---|
| `a_dh_ratchet_rebases_the_lease_onto_the_fresh_chain` | `the retired chain's ceiling (2048) must not survive onto the fresh chain` |
| `a_ratcheted_record_still_loads_after_a_restart` | `InvalidSessionStructure("reserved sender chain index implausibly far ahead")` |
| `the_rebased_lease_still_covers_the_fresh_chain_without_a_flush` | fails |
| `a_rebased_lease_never_republishes_a_counter_across_a_crash` | fails |

The second reproduces the operator-visible failure verbatim, including that it only appears after a restart: the same record round-trips fine under a live incarnation, which is why this survived so long in normal operation.

`cargo fmt --all`, `cargo clippy --workspace --all-targets` with zero warnings, and green suites: `whatsapp-rust` 1238, `wacore` 1269, `wacore-libsignal` 199 plus the 11 `counter_lease` integration tests. Both commits build and pass on their own.

## For the reporter

The workaround in the issue (`DELETE FROM sessions WHERE address = '<owner-lid>.0'`) stops being necessary with fix 2 — an already-rotten row is discarded and renegotiated on its own. Fix 1 stops new rows from rotting. The blob size difference noted in the issue (~1096 bytes rotten vs ~414 fresh) is consistent with this: the rotten record carries archived previous sessions the fresh one has not accumulated yet.

### DIFF
--- a/agent_docs/signal_durability.md
+++ b/agent_docs/signal_durability.md
@@ -19,6 +19,25 @@ previously persisted lease and can use write-behind. A send at the bound raises
 it by `SENDER_CHAIN_RESERVATION_BATCH` and must wait for a successful durable
 flush before its ciphertext is published.
 
+A reservation describes one sender chain, not the address. A DH ratchet
+replaces the chain in place with fresh random key material and drops the old
+one without archiving it, so the inherited ceiling stops describing anything
+reachable: `rebase_lease_after_sender_chain_reset` lowers it back to one batch
+as part of the same mutation. Lowering is sound only there — no snapshot can
+pair the retired chain with the rebased ceiling — and only downward, so a
+counter is never published under a ceiling that is not yet durable. Leave it
+stranded and a long monologue followed by one peer reply puts the gap past
+`MAX_RESERVATION_FAST_FORWARD`, where recovery can neither burn nor accept the
+record. A chain that is *archived* rather than discarded keeps its claim:
+`promote_fresh_state` burns the outgoing state to the ceiling before resetting
+it.
+
+An undecodable session row is reported absent rather than surfaced as a load
+error. Every path that could replace it — the peer's next pre-key message, the
+retry repair — must load it first, so a propagated error strands the address
+permanently. `wa_session_record_quarantined_total` counts these; steady state
+is zero.
+
 The cache takes ownership of transient record gates. A failed write, a checked
 out record skipped by a flush, or a tombstone whose delete failed must remain
 gated. Only the backend operation that persisted that address may release it.
--- a/wacore/libsignal/src/protocol/session_cipher.rs
+++ b/wacore/libsignal/src/protocol/session_cipher.rs
@@ -851,7 +851,10 @@ fn create_decryption_failure_log(
 }
 
 enum RecordDecryptState {
-    Current(DecryptSnapshot),
+    Current {
+        snapshot: DecryptSnapshot,
+        sender_chain_reset: bool,
+    },
     Previous {
         state: Box<SessionState>,
         effect: StateDecryptEffect,
@@ -920,7 +923,7 @@ impl RecordDecryptTransaction<'_> {
             .as_ref()
             .expect("a live decrypt transaction owns its rollback")
         {
-            RecordDecryptState::Current(_) => self
+            RecordDecryptState::Current { .. } => self
                 .record
                 .session_state()
                 .expect("a current decrypt transaction keeps the current state installed"),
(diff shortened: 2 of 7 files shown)


jlucaso1:
@codspeedbot why this regret?

PR #1151, opened 2026-07-27
perf(storage): let WAL readers run, and stop search from monopolizing them

### LINKED ISSUE #1147: chat-store: FTS search ranks the whole index for short prefixes, has no chat filter, and hydrates hits one query at a time
## Summary

`ChatStore::search_messages` is unusable as a type-ahead search on a real store, for two independent reasons, and its cost is not bounded by the caller's `limit`.

Measured on a live session (warm page cache, SQLite on a Docker bind mount, `pool_size: 1`):

| query | limit | wall clock |
|---|---|---|
| `hello` | 10 | 3.18 s |
| `hello` | 40 | 3.36 s |
| `thanks` | 40 | 5.82 s |
| `h` | 40 | > 30 s (request timeout) |

Note `limit=10` and `limit=40` differ by only 0.18 s: the ~3 s is a fixed floor in the MATCH itself, and hydration is ~6 ms per hit warm (~110 ms cold).

## 1. A short prefix term ranks the whole index

`build_match_query` (`storages/chat-store/src/fts.rs:70-90`) makes every token a prefix term (`"tok"*`), and the query is `... WHERE messages_fts MATCH ? AND m.device_id = ? ORDER BY rank LIMIT ?` (`fts.rs:114-125`). `ORDER BY rank` has to score the entire match set before `LIMIT` applies, so a one- or two-character token scores essentially every message in the store.

**This is not merely slow - it is a denial of service against the session**, because the work runs inside `spawn_blocking`, which cannot be cancelled. When the caller's request times out the query keeps running and keeps holding the store's single `db_semaphore` permit, so every other read on that session (chat list, thread, contacts, groups) queues behind a request nobody is waiting for. We observed a single stray keystroke making one account's whole inbox return 408 for over two minutes. We now refuse queries shorter than two characters embedder-side, but the underlying shape means a two-character query on a large store has the same character.

## 2. No chat predicate, so a scoped search over-fetches and post-filters

There is no way to ask for "hits in this chat", though the query already joins `messages` and could take `AND m.chat_jid = ?`. An embedder scoping a search has to over-fetch globally and post-filter, which multiplies the hydration cost by whatever safety factor it picks, and still silently loses hits for a chat that ranks sparsely.

## 3. Hydration is sequential, and each hit costs two queries

`search_messages` hydrates each hit with `self.message(&chat, &hit.msg_id).await?` in a loop (`fts.rs:128-140`), and `message()` itself first runs `chat_key_candidates` (`queries.rs:226` -> `lid.rs:92-105`). So N hits cost ~2N queries, serialized, each acquiring the single permit. At 320 hits on a cold cache that is ~35 s before any ranking cost.

## Asks

1. **A chat filter parameter** on `search_messages` (`AND m.chat_jid IN (..)` inside the existing join), so scoped search stops over-fetching.
2. **Hydrate hits concurrently**, or better, return the rows from one query rather than N point queries - the FTS join already reaches `messages`, so the full row could be selected in the same statement.
3. **A minimum term length, or a non-ranked fast path** for very short prefixes, so a short query cannot monopolize the store. Even returning unranked `LIMIT`-ed matches would be preferable to ranking the whole index.

Happy to PR any of these.

### LINKED ISSUE #1148: sqlite-storage: pool_size gates reads and writes together, so WAL's concurrent readers are unusable
## Summary

`SqliteStoreConfig::pool_size` sets the r2d2 pool size **and** the `db_semaphore` permit count in lockstep (`storages/sqlite-storage/src/sqlite_store.rs:153-155`, `:386`, `:439`), so with the default `pool_size: 1` every operation on a session - reads included - serializes behind one permit. WAL is enabled unconditionally (`:424`), and WAL exists precisely so that many readers can run concurrently with one writer, so reads are being serialized for no storage-level reason.

## Why we cannot just raise pool_size

We tried. At `pool_size: 4` the chat-store's batching writer began reporting `write batch failed: database operation error` from `commit_batch` within minutes. A failed batch is dropped with a warning and reported to the next `flush()`, which suppresses the inbound transport ack, so WhatsApp redelivers - the single permit is what serializes that writer, and it is load-bearing. We pin `pool_size: 1` deliberately and have a comment recording the incident.

The result is that the knob which would fix read concurrency is the same knob that breaks write correctness.

## Impact for an embedder

One session's reads queue linearly. Measured on a live account: a chat-list page costs ~0.5 s of permit time and a thread page ~0.28 s, so a third concurrent viewer of the same account waits ~1.5 s, and any long-running read (see the FTS issue) stalls every other read on that session for its whole duration.

## Ask

Separate the two: a read connection pool (N connections, no permit or its own permit count) and a single write permit, rather than one `pool_size` governing both. Concretely, either

- split `pool_size` into `read_pool_size` / `write_pool_size` where the semaphore only gates writes, or
- keep one pool but gate only the write path on the 1-permit semaphore, letting reads acquire a connection directly.

That preserves the write-batch serialization exactly as it is today while letting concurrent readers use WAL as intended. Happy to PR it if the shape sounds right.

### LINKED ISSUE #1150: chat-store: unavailable view-once fanouts materialize as plain `undecryptable`, dropping the type the UI needs
## What happens

When a peer sends a view-once photo/video/voice note, companions never receive the content - the server fans out an `<unavailable type="view_once"/>` child instead. `src/message/receive.rs` handles this correctly: it detects the subtype the way WA Web's parser does, skips the futile PDO placeholder-resend, acks directly, and dispatches

```rust
UndecryptableMessage { is_unavailable: true, unavailable_type: UnavailableType::ViewOnce, decrypt_fail_mode: Show, .. }
```

(log line: `Message has unrecoverable <unavailable> child (type: ViewOnce); skipping futile PDO placeholder-resend and acking directly`.)

But the chat-store's event apply flattens every `UndecryptableMessage` into the same placeholder row:

```rust
// storages/chat-store/src/store.rs (Event::UndecryptableMessage arm)
kind: KIND_UNDECRYPTABLE,
text: None,
```

`unavailable_type` is dropped, so a frontend reading the store cannot tell a transient decrypt failure ("Waiting for this message", may still resolve via PDO/resend) from a view-once fanout (will NEVER resolve on a companion - the phone won't share it).

## Why it matters

The two states need opposite UI. WA Web renders exactly this stanza as its one-time chip (the `(1)` "Photo - open on your phone" bubble), and it can only do that because the wire told it `type="view_once"`. A store reader today has no choice but to render the generic "Waiting for this message" placeholder for a message that is permanently unavailable by design - which reads as a sync bug to the operator.

## Ask

Materialize unrecoverable unavailable fanouts under their own kind - e.g. `kind: "view_once"` (`MessageKind::Other` already round-trips unknown labels for older readers), or keep `kind: "undecryptable"` and store the unavailable type alongside it. The `Hosted` and `Bot` flavors from `UnavailableType::from_fanout_flags` deserve the same treatment - all three are `is_unrecoverable_fanout()` and none of them will ever be filled in by a retry.

Observed at rev 99d9aa89 with a live account (view-once photo from a phone to a paired companion).

jlucaso1:
Closes #1148, closes #1147, closes #1150.

#1147 and #1148 arrived as two issues but they are one story. The search query was slow; the reason a slow query took the whole *session* down was that reads and writes shared a queue. Fixing only the query would have left the next slow read to do the same thing, so both halves are here. #1150 is independent and rides along in the same crate.

Every number below was measured before the code was written — the SQLite behaviour is reproducible with `python3` and a synthetic index.

## #1148 — `pool_size` gates reads and writes together

`pool_size` drove r2d2's `max_size` **and** the serialization semaphore in lockstep. So the only knob that could admit concurrent readers admitted concurrent writers with them — and concurrent writers is precisely what SQLite will not have: two deferred transactions that read and then write deadlock on the upgrade, and `busy_timeout` cannot break that. Hence `pool_size: 4` producing `database operation error` from the batching writer within minutes, and the deliberate pin at 1. The knob that would fix read concurrency was the knob that broke write correctness.

**`read_pool_size` is the knob that was missing.** Its connections sit *on top of* the write pool rather than sharing it, so readers saturating their permits can never leave the writer waiting on r2d2 for a connection. Reads take a permit of their own, so the number of blocking threads parked on `pool.get()` stays bounded by the pool rather than by the caller.

- `SharedSqlite::read` is the path that uses them; `run` remains the write path and keeps the single permit that serializes the chat-store's batching writer — untouched, because it is load-bearing exactly as the issue says.
- Left at its default of `0`, `read` queues on the write permit exactly like `run`: no extra connections, no behaviour change, no memory change for a process holding many per-session stores. Opting in is now safe in a way raising `pool_size` never was.

The chat-store's nine query paths move onto `read`. `put_media_ref` writes, so it stays on `run`.

**`read` also runs its closure in a deferred transaction.** The permit was quietly doing double duty: while a read held it, the writer could not commit *between that read's statements*. Dropping the permit dropped that too, and a reader that resolves a chat's identity keys and then queries by them — or collects search hits and then hydrates them — would straddle two committed states and come back short. Doing it inside `read` rather than at each call site makes "read" mean "consistent read" by construction. A deferred transaction over read-only statements never asks for the write lock, so it pins the snapshot without contending with the writer. *(Thanks @coderabbitai — this was a real hole opened by the split.)*

Three regression tests, and they discriminate: with `read_pool_size: 0` the four readers would share one permit and deadlock on the barrier until the timeout fires.

## #1147 — FTS search

Three independent costs, of which only the last is really about search.

**Scope.** `search_messages_in_chat` filters inside the FTS join. Over-fetching globally and post-filtering cost more *and* silently lost hits in a chat that ranks sparsely. Resolves the peer's PN/LID alias like every other read.

**Hydration.** Each hit was a point query, and each of those first re-resolved the chat's identity keys — so N hits cost ~2N serialized round trips, every one taking the store's only permit. The rowids now come back from the match and the rows from a single `eq_any`, reordered to the rank the search produced.

**Ranking.** `ORDER BY rank` must score every row a term matches before `LIMIT` can discard any, so its cost tracks the match set rather than the page asked for — a one-character prefix scored most of the store. Terms under three characters order by arrival instead, which FTS5 streams and `LIMIT` cuts short:

| index size | `ORDER BY rank` | `ORDER BY rowid DESC` |
|---|---:|---:|
| 20k rows | 20.8 ms | 1.1 ms |
| 80k rows | 83.5 ms | 4.2 ms |
| 320k rows | 339.5 ms | 16.8 ms |

*(`"h"*` at `LIMIT 40`. The query plan changes with it — `INDEX 192` instead of `INDEX 32` — which is FTS5 switching to a rowid-ordered walk.)* An ordinary term gains too: `"hello"*` on 60k rows goes 13.3 ms → 0.8 ms.

**Worth stating plainly:** both orderings still scale with the match set, because a short prefix's doclist grows with the store. This is a ~20× constant, not a bound. What actually stops a slow search from taking the session down is the read/write split above; the ordering change just makes it much cheaper to begin with.

The rule is `all(token >= 3 chars)`, so one short token demotes the whole query to recency — pinned by a test, since `any` would be an easy "optimization" to make by accident.

## #1150 — unavailable fanouts lose the type the UI needs

Every `UndecryptableMessage` flattened into the same placeholder row, so a frontend could not tell a decrypt failure a retry may still resolve from a view-once fanout that never will — the phone does not share that content with a companion. The two want opposite UI, and `receive.rs` had already classified it correctly; the store threw the distinction away.

The three unrecoverable subtypes now materialize under their own kinds — `view_once`, `hosted`, `bot` — while a plain fanout stays `undecryptable`, because PDO can still fill that one in.

They are typed `MessageKind` variants rather than `Other` labels. The enum's stated purpose is that no frontend hard-codes label strings; leaving these as `Other("view_once")` would have forced exactly that. The chat preview carries the kind too, so a chat list can render the chip without opening the thread.

The labels are restated in the chat-store rather than forwarding `UnavailableType::as_str`: that string is the wire value and belongs to the protocol, while these are on-disk values that must not move if a wire attribute is renamed. A test pins them.

## Breaking

`SqliteStoreConfig` gains `read_pool_size`, which breaks struct-literal construction — the repo's own test needed updating, which is the proof. `with_read_pool_size` joins the existing `with_mmap_size` builder, whose doc already states the intent that "new optional knobs don't force struct-literal churn"; making that true in general would mean `#[non_exhaustive]` on the struct, which is a larger API call than this PR should make on its own.

Note the informational Semver check does **not** catch this: it only processes `waproto` (`Finished [61.6s] waproto`), and its red is the same pre-existing drift against the published `waproto-0.6.0` that predates this branch.

`MessageKind` gains three variants, which is fine — it is `#[non_exhaustive]`.

## Tests

189 pass across the two crates (103 chat-store integration + 26 chat-store unit + 60 sqlite-storage). Clippy clean in both feature modes, since `fts.rs` is behind `search`.

## Not covered

The device store's own reads still go through `with_semaphore`, so they remain serialized with its writes. The paths this PR moves are the chat-store's — the ones the issue measured (chat list ~0.5 s, thread ~0.28 s of permit time). Splitting the device store's trait methods means classifying eight call sites as read or write, which deserves its own change rather than being done by inspection here.

### DIFF
--- a/storages/chat-store/src/fts.rs
+++ b/storages/chat-store/src/fts.rs
@@ -195,8 +195,11 @@ fn fts_hits(
 ) -> wacore::store::error::Result<Vec<FtsHit>> {
     use diesel::sql_types::{BigInt, Integer, Text};
 
-    // Constant fragments, never caller input.
-    let order = if ranked { "rank" } else { "f.rowid DESC" };
+    // Constant fragments, never caller input. `rank` is qualified because it is
+    // only unambiguous today by accident — `messages` has no such column, and
+    // an unqualified reference would silently start resolving to it if one were
+    // ever added.
+    let order = if ranked { "f.rank" } else { "f.rowid DESC" };
     let Some(first_key) = keys.first() else {
         return diesel::sql_query(format!(
             "SELECT f.rowid AS rowid
--- a/storages/chat-store/src/fts.rs
+++ b/storages/chat-store/src/fts.rs
@@ -100,6 +100,10 @@ fn build_match_query(input: &str) -> Option<String> {
 /// things that start with this" is a defensible answer for a term that short.
 const MIN_RANKED_TERM_LEN: usize = 3;
 
+/// Max rowids per hydration statement, under SQLite's default 999
+/// host-parameter limit.
+const ID_PARAM_CHUNK: usize = 900;
+
 #[derive(QueryableByName)]
 struct FtsHit {
     #[diesel(sql_type = diesel::sql_types::BigInt)]
@@ -111,7 +115,8 @@ impl ChatStore {
     /// query is plain words (prefix-matched); FTS5 operators are neutralized.
     ///
     /// A term shorter than three characters is ordered newest-first instead of
-    /// by relevance — see [`MIN_RANKED_TERM_LEN`].
+    /// by relevance: ranking has to score every row such a prefix matches
+    /// before `limit` can discard any, which on a real store means most of it.
     pub async fn search_messages(&self, query: &str, limit: i64) -> Result<Vec<StoredMessage>> {
         self.search(query, None, limit).await
     }
@@ -163,16 +168,25 @@ impl ChatStore {
                     if hits.is_empty() {
                         return Ok(Vec::new());
                     }
-                    // One statement for every hit, instead of a point query per hit
-                    // (each of which used to re-resolve the chat's identity keys
-                    // first, so N hits cost ~2N serialized round trips).
+                    // One statement per chunk of hits, instead of a point query
+                    // per hit (each of which used to re-resolve the chat's
+                    // identity keys first, so N hits cost ~2N serialized round
+                    // trips). `limit` is the caller's, so the id list is chunked
+                    // rather than trusted to stay under SQLite's host-parameter
+                    // ceiling.
                     let ids: Vec<i64> = hits.iter().map(|hit| hit.rowid).collect();
-                    let mut rows: Vec<MessageRow> = schema::messages::dsl::messages
-                        .filter(schema::messages::dsl::rowid.eq_any(&ids))
-                        .load(conn)
-                        .map_err(db_err)?;
-                    // `eq_any` returns table order; restore the order the ranking
-                    // (or the recency scan) put them in.
+                    let mut rows: Vec<MessageRow> = Vec::with_capacity(ids.len());
+                    for chunk in ids.chunks(ID_PARAM_CHUNK) {
+                        rows.extend(
+                            schema::messages::dsl::messages
+                                .filter(schema::messages::dsl::rowid.eq_any(chunk))
+                                .load::<MessageRow>(conn)
+                                .map_err(db_err)?,
+                        );
+                    }
+                    // `eq_any` returns table order, and chunking splits it
+                    // further; restore the order the ranking (or the recency
+                    // scan) put them in.
                     let rank_of: std::collections::HashMap<i64, usize> =
                         ids.iter().enumerate().map(|(at, id)| (*id, at)).collect();
                     rows.sort_by_key(|row| rank_of.get(&row.rowid).copied().unwrap_or(usize::MAX));
--- a/storages/chat-store/src/fts.rs
+++ b/storages/chat-store/src/fts.rs
@@ -9,9 +9,11 @@
 //! `INSERT INTO messages_fts(messages_fts) VALUES('rebuild')`.
 
 use diesel::prelude::*;
-use log::warn;
+use wacore_binary::Jid;
 
 use crate::error::{ChatStoreError, Result, db_err};
+use crate::queries::MessageRow;
+use crate::schema;
 use crate::store::ChatStore;
 use crate::types::StoredMessage;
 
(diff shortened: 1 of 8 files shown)


jlucaso1:
Pushed `49ee658`, which closes out the review round.

**The P1 on the read path (Codex).** The finding was right, and the fix went further than raising the semaphore. Readers now get their **own r2d2 pool**; the write pool's `max_size` and permit count are untouched. Widening the single semaphore was the bug — it also let concurrent *writers* into the pool, and SQLite serializes those itself, so the win was readers and the cost was contention nobody asked for.

Two properties came along with the split:

- The reader pool opens with `PRAGMA query_only = 1`. A write that wanders onto `read()` now fails loudly instead of being a hazard we document and hope callers read.
- It is only built when the database actually came up in WAL — read back from `PRAGMA journal_mode`, not assumed from the setting we requested. Without WAL a reader blocks on the writer anyway, so the store logs a warning and keeps everything on the write permit. Default stays `read_pool_size = 0`; embedders opt in.

Covered by five tests in `shared.rs`: readers overlap when configured, a write proceeds while every reader permit is held, a write through the read path is refused, reader connections are declined without WAL, and the default falls back to the write permit (the writer holds the permit and the concurrent read is asserted *not* to have completed).

**Host-parameter ceiling (CodeRabbit).** Real. `limit` is the caller's number, and past 999 ids SQLite refuses the statement outright. Hydration now chunks at 900 per statement, and the rank-restoring sort was already keyed off the hit order so it survives the split.

**Repeated rationale.** Dropped the duplicated comment at the `unavailable_kind` call site — AGENTS.md puts the *why* at the single point where the decision is made.

Four findings I did not apply, with reasons:

- **Inline the label literals in `materialize.rs`.** The stated rationale — that `MessageKind::ViewOnce.as_str()` has a lifetime problem — isn't accurate; unit variants are const-promoted and it yields `&'static str`. Going through `MessageKind` keeps it the single source of truth for on-disk labels, which `unavailable_kind_labels_are_stable` pins.
- **Add an inverse-label test.** Already covered: `unrecoverable_fanouts_keep_their_type` reads the rows back through `from_db`.
- **`pool_size + read_pool_size` overflow.** Obsolete after the redesign — separate pools, no addition.
- **"#1150 is out of scope."** Requested for this PR.

`TxnError` stayed a manual `From` impl rather than `thiserror` — this crate doesn't depend on it, and it's a private two-variant adapter that is never `Display`ed.

Verified locally on both touched crates, tests and clippy, with `search` on and off. Workspace clippy is CI's.

---

jlucaso1:
Two red checks on `49ee658`, neither from this diff. Details below; E2E is re-running.

**E2E — `chatstate_ttl::test_fresh_chatstate_delivered_on_reconnect`**

```
panicked at tests/e2e/tests/chatstate_ttl.rs:69:10:
B should receive fresh chatstate within TTL: Timed out waiting for event
```

This PR can't reach it: `tests/e2e` depends on neither `whatsapp-rust-sqlite-storage` nor `whatsapp-rust-chat-store`, and every `TestClient` runs on `InMemoryBackend`. The same job was green on `5705d2d`, the previous commit on this branch.

The mock-server log shows a real race in the test, and it's a narrow one:

```
17:30:47.960  test: B disconnected (reconnect_immediately)
17:30:47.963  test: A sent typing indicator to offline B
17:30:47.972  server: Unregistered connection e31739b6 (B's old socket)
17:30:47.980  server: Registered connection 48130e41 (B back online)
17:30:48.264  server: Draining 0 offline events (0 ordered + 0 presence) …
```

A's `send_composing` landed 9 ms *before* the server unregistered B's previous connection, so the server routed the presence to a socket that was already closing instead of queuing it. B's new connection came up 8 ms later to an empty queue — "0 presence" — and the 15 s wait timed out.

`wait_for_disconnected` polls the *client's* view of the connection, which flips as soon as the client drops its socket; the server's view lags by however long the close takes to process. `reconnect_immediately()` makes that gap maximally likely, because teardown and re-dial race each other. There's no server-side offline signal for the test to wait on, so this is a pre-existing flake in the test rather than something the client can fix — leaving it as-is, since it's unrelated to this PR.

**Semver Checks (informational)**

30 failures in `wacore`, 2 in `wacore-binary`, 10 in `waproto`. This PR touches none of those three crates — the only mentions of `storages/chat-store` and `storages/sqlite-storage` in the log are Cargo.toml paths in the workspace listing. Same pre-existing baseline drift as on #1144 (`v0.6.0 -> v0.6.0`, "no change; assume minor", against accumulated API changes), which is why the job is marked informational.

---

jlucaso1:
Pushed `7da383b`. All three Codex P2s were real and are fixed.

**Shared cache (P2, line 516).** Confirmed with a direct experiment rather than taken on trust — and the failure is exactly as described:

```
journal_mode: wal
shared-cache write while a read txn is open:  OperationalError: database table is locked: t
private-cache write while a read txn is open: OK
```

`busy_timeout` was set to 3 s on both connections and the write failed instantly, because `SQLITE_LOCKED_SHAREDCACHE` isn't retried by the busy handler. The WAL gate alone let these through: shared cache reaches WAL, then replaces WAL's snapshot isolation with table locks held for the length of a transaction. A `cache=shared` URI now declines reader connections the same way a non-WAL journal does, and the two reasons share one warning path. Test asserts both directions — same file declines with the parameter, gets readers without it — so it's the cache mode being tested, not the path.

**Configured thread pool (P2, line 530).** Correct: the reader pool called `shared_r2d2_thread_pool()` unconditionally, so a caller supplying `SqliteStoreConfig::thread_pool` got their scheduler honoured for writes and silently ignored for readers — the one thing that field exists to control. Both pools share the configured `Arc` now.

**`resource_report` (P2, line 528).** Also correct, and it's the bound's whole purpose that it was undermining: reader connections carry a page cache each, and the estimate counted only `self.pool`. It now adds the reader pool's open connections. Test builds one store with readers and one without and asserts the estimate widens.

---

**On the CodSpeed report** — `bench_unpack_compressed`, 8 KB → 54.5 KB. Not a regression from this PR, which touches neither `wacore-binary` nor anything it depends on. The numbers say what happened:

| run | allocs | bytes |
|---|---|---|
| main `99d9aa8` | 3 | 8224 |
| main `f3ef3d2` (base) | 3 | 8224 |
| this branch `5705d2d` | 3 | 8224 |
| this branch `49ee658` (flagged) | **4** | **55776** |

One extra allocation, of 47552 B. That matches the per-thread `zlib_rs::Inflate` state — "~48 KB", per the comment on `decompress_zlib_pooled` — which `zlib_pool.rs` initializes lazily in a `thread_local!`. The flagged run measured the one-time init inside its measured iteration; every other run had already paid it before measurement began. Nothing in the decompression path changed, and no dependency moved (the size report shows 471 → 471 crates). The run on `7da383b` should read 8 KB again.

---

PR #1155, opened 2026-07-27
fix(chat-store): keep receipt times for 1:1 chats, per state

### LINKED ISSUE #1154: chat-store: message_receipts are group-only, so a DM message info has no delivered/read times
## What happens

`apply_receipt` materializes `message_receipts` rows (user, type, `ts_ms`) only when the receipt's chat is a group:

```rust
// storages/chat-store/src/store.rs, apply_receipt
if receipt.source.chat.is_group() {
    insert_into(message_receipts)...
}
```

For a DM the receipt only advances `messages.status` - the state a message reached survives, but the instant it was delivered/read/played is dropped on the floor, even though the `Receipt` event carries `timestamp` for every flavor.

## Why it matters

A frontend cannot render WhatsApp's own Message info for a 1:1: WA Web shows "Delivered hh:mm / Read hh:mm" in DMs and keeps per-message info rows for both chat kinds, not just groups. A store reader today can only show the bare state ("Delivered") with no time, which reads as a missing feature next to the real client.

## Ask

Materialize the DM receipt rows too - the natural shape is the same table with `user_jid` = the peer user (self receipts already take the read-state path above and would stay out), keeping the existing advance-only conflict handling. Alternatively, timestamp columns on `messages` (`delivered_at_ms`/`read_at_ms`) would carry the same information with less cardinality; the row form has the edge that `Played` on a voice note and multi-device idempotency already work exactly as they do for groups.

Observed at rev 99d9aa89 (DM receipts verified arriving and advancing `messages.status` while the receipts read stays empty).

jlucaso1:
Closes #1154.

`apply_receipt` materialized `message_receipts` rows only when the receipt's chat was a group. For a DM the receipt advanced `messages.status` and nothing else — the state a message *reached* survived, the instant it got there was dropped, even though every `Receipt` carries a timestamp. A store reader could render "Delivered" but never "Delivered 14:32".

## Checked against WA Web first

The issue says WA Web shows "Delivered hh:mm / Read hh:mm" in DMs. It does, and the bundle also settles the shape the rows should take — which turned out to matter more than the gate itself.

WA Web ships **two** message-info drawers: `WAWebContactMsgInfoDrawer.react` (`placement:"contact"`) and `WAWebGroupMsgInfoDrawer.react` (`placement:"group"`). The contact one — the 1:1 case — builds this:

```js
var oe = U.played.head(),   ae = oe?.t;   Z.push(ChatContactMsgInfoCell({title:"Played"/"Opened", t:ae, …},"played"))
var se = U.read.head(),     ue = se?.t;   Z.push(ChatContactMsgInfoCell({title:"Read",            t:ue, …},"read"))
var de = U.delivery.head(), me = de?.t;   Z.push(ChatContactMsgInfoCell({title:"Delivered",       t:me, …},"delivery"))
```

`U` is `MsgInfoCollection.find(msgId)`, the same model the group drawer uses; a DM just calls `.head()` because there is one peer. `ChatContactMsgInfoCell` renders `Clock.relativeDateAndTimeStr(t)` as each row's secondary line — the timestamp. So a message's receipts are **three separate collections**, each entry carrying its own `t`, and the drawer reads all three at once.

## Why the key had to change

The issue proposes the same table with `user_jid` = the peer, "keeping the existing advance-only conflict handling". That alone does not get there, and this is the one place I departed from the ask.

The old primary key ended at `user_jid`:

```sql
PRIMARY KEY (device_id, chat_jid, msg_id, user_jid)
```

so a user's row advanced *in place* — the `read` receipt overwrote the `delivered` one and took its `ts_ms` with it. That is enough to render a group's "read by" list, which only ever asks each member's furthest state, and it is why the narrower key held up until now. But it cannot answer "when was this delivered" once the message has been read, which is exactly what the 1:1 drawer stacks above the read line. Following the letter of the ask would have shipped a feature that still could not render the screenshot the issue is about.

The state joins the key, so each transition keeps its own row and both instants survive. That is also WA Web's own model rather than an invention.

The write becomes insert-only as a result: the stored instant is the **first** report of a state. A replayed receipt is a duplicate, not a later event, and a peer's second device acking afterwards does not mean the user got it later.

`delivered_at_ms`/`read_at_ms` columns on `messages` — the issue's alternative — were the other option. Rows won for the reason the issue itself gives: `Played` on a voice note is a third state, and multi-device idempotency already works. Cardinality stays bounded either way, at most three rows per outgoing message per participant.

## Two consequences elsewhere

**A DM receipt resolved by alias files under the message's key.** The PN/LID retry updates a message under the counterpart key; its receipt belongs there too, because the satellite prune is per chat and deletes receipt rows whose `msg_id` is absent from *that* chat. A row left under the wire key would not survive the next trim.

**The LID merge drops its "keep the furthest state" pass.** That pre-delete removed the destination's row when the source held a higher state, so the surviving row was the furthest one. With the state in the key, a side holding `read` and a side holding `delivered` are two facts about one message rather than two candidates for one row. `UPDATE OR IGNORE` still drops a source row colliding on the whole key, where the destination already records that state at the earlier instant by the same rule that wrote it.

## Migration

`2026-07-27-000002_message_receipts_per_state` rebuilds the table with the wider key. Existing rows are already unique on the narrower key, so they stay unique under the wider one and carry over untouched — each user keeping the single furthest state recorded so far, with earlier states simply absent rather than wrong. The table carried no indexes beyond its primary key, so the rebuild drops nothing. `down.sql` collapses back to the furthest state per user.

## Groups

Unaffected in what they can answer. Two devices of one member still collapse to one participant, because `user_jid` remains the bare identity — the existing regression test for that now asserts the property it was really guarding (one member, not one row) and pins the two states that member accumulates. Group readers gain the delivered instant they were previously losing to the read receipt.

## Tests

107 pass in the chat-store (26 unit + 81 integration, of which four are new):

- `dm_receipts_record_when_each_state_was_reached` — both instants survive, and `messages.status` is unchanged by any of it
- `dm_played_receipt_joins_read_rather_than_replacing_it` — the voice-note case
- `a_replayed_dm_receipt_does_not_move_the_recorded_instant` — insert-only
- `a_dm_receipt_resolved_by_alias_files_under_the_message_key` — reachable under either identity, and stored under the message's key

Clippy clean in both feature modes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PyCszWBEYZoqnTRfHpSQ2X)_

### DIFF
--- a/storages/chat-store/migrations/2026-07-27-000002_message_receipts_per_state/down.sql
+++ b/storages/chat-store/migrations/2026-07-27-000002_message_receipts_per_state/down.sql
@@ -1,0 +1,26 @@
+-- Back to one row per user. Several states per user collapse to the furthest
+-- one, which is what the narrower key could represent; the earlier instants
+-- are dropped because there is nowhere to put them.
+CREATE TABLE message_receipts_old (
+    device_id INTEGER NOT NULL,
+    chat_jid TEXT NOT NULL,
+    msg_id TEXT NOT NULL,
+    user_jid TEXT NOT NULL,
+    receipt_type INTEGER NOT NULL,
+    ts_ms BIGINT NOT NULL,
+    PRIMARY KEY (device_id, chat_jid, msg_id, user_jid)
+);
+
+INSERT INTO message_receipts_old
+    (device_id, chat_jid, msg_id, user_jid, receipt_type, ts_ms)
+SELECT r.device_id, r.chat_jid, r.msg_id, r.user_jid, r.receipt_type, r.ts_ms
+FROM message_receipts r
+WHERE r.receipt_type = (
+    SELECT MAX(s.receipt_type) FROM message_receipts s
+    WHERE s.device_id = r.device_id AND s.chat_jid = r.chat_jid
+      AND s.msg_id = r.msg_id AND s.user_jid = r.user_jid
+);
+
+DROP TABLE message_receipts;
+
+ALTER TABLE message_receipts_old RENAME TO message_receipts;
--- a/storages/chat-store/migrations/2026-07-27-000002_message_receipts_per_state/up.sql
+++ b/storages/chat-store/migrations/2026-07-27-000002_message_receipts_per_state/up.sql
@@ -1,0 +1,37 @@
+-- Keep one receipt row per state, not one per user that overwrites itself.
+--
+-- The old key ended at `user_jid`, so a user's row advanced in place: the
+-- `read` receipt overwrote the `delivered` one and took its `ts_ms` with it.
+-- That is enough to render a group's "read by" list, which only ever asks for
+-- each member's furthest state, and it is why the narrower key held up until
+-- 1:1 message info needed something else.
+--
+-- WA Web models a message's receipts as three separate collections
+-- (`.delivery`, `.read`, `.played`), each entry carrying its own `t`, and its
+-- 1:1 info drawer reads all three at once to show "Delivered hh:mm" above
+-- "Read hh:mm". Both of those instants have to survive for that to render, so
+-- the state joins the key and each transition keeps its own row.
+--
+-- Existing rows are already unique on the narrower key, so they stay unique
+-- under the wider one and carry over untouched — each user keeping the single
+-- furthest state recorded so far, with earlier states simply absent rather
+-- than wrong.
+CREATE TABLE message_receipts_new (
+    device_id INTEGER NOT NULL,
+    chat_jid TEXT NOT NULL,
+    msg_id TEXT NOT NULL,
+    user_jid TEXT NOT NULL,
+    -- Same scale as messages.status: 3 delivered, 4 read, 5 played.
+    receipt_type INTEGER NOT NULL,
+    ts_ms BIGINT NOT NULL,
+    PRIMARY KEY (device_id, chat_jid, msg_id, user_jid, receipt_type)
+);
+
+INSERT INTO message_receipts_new
+    (device_id, chat_jid, msg_id, user_jid, receipt_type, ts_ms)
+SELECT device_id, chat_jid, msg_id, user_jid, receipt_type, ts_ms
+FROM message_receipts;
+
+DROP TABLE message_receipts;
+
+ALTER TABLE message_receipts_new RENAME TO message_receipts;
--- a/storages/chat-store/src/lid.rs
+++ b/storages/chat-store/src/lid.rs
@@ -357,14 +357,18 @@ pub(crate) fn merge_split_chat(
     .bind::<Text, _>(src)
     .bind::<Text, _>(dest)
     .execute(conn)?;
-    // A row already under `dest` whose renamed form collides with an existing
-    // one is skipped by `OR IGNORE` above, and the `chat_jid = src` sweep at
-    // the end cannot reach it. Left alone it would outlive the merge still
-    // naming the retired identity — one peer read back as two users, the exact
-    // failure this reconciliation exists to prevent. Its instant is already
-    // folded in, so it is a pure duplicate by now.
+    // Past that rename, naming `src` is proof of a collision: the only rows
+    // still doing so are the ones `OR IGNORE` skipped because their renamed
+    // form already existed. Their instants are folded in, so every one of them
+    // is a pure duplicate — and both chat keys need sweeping, not just `dest`.
+    // A survivor under `src` would otherwise be carried to `dest` intact by the
+    // chat rename below, and one under `dest` is beyond that rename's reach
+    // entirely. Either way it outlives the merge still naming the retired
+    // identity: one peer read back as two users, the exact failure this
+    // reconciliation exists to prevent.
     diesel::sql_query(
-        "DELETE FROM message_receipts WHERE device_id = ?1 AND chat_jid = ?3 AND user_jid = ?2",
+        "DELETE FROM message_receipts \
+         WHERE device_id = ?1 AND chat_jid IN (?2, ?3) AND user_jid = ?2",
     )
     .bind::<Integer, _>(device_id)
     .bind::<Text, _>(src)
--- a/storages/chat-store/src/lid.rs
+++ b/storages/chat-store/src/lid.rs
@@ -316,62 +316,69 @@ pub(crate) fn merge_split_chat(
         .bind::<Text, _>(src)
         .execute(conn)?;
 
-    // A 1:1's receipts name the peer, and the peer is what this merge is
-    // reconciling: rows recorded while the thread answered to its other
-    // identity carry that identity in `user_jid`. Nothing else would ever put
-    // the two halves back together — relocation moves only `chat_jid`, and the
-    // collision passes below match `user_jid` exactly — so one person would
-    // stay split across two users forever. Rewrite it first, so a row that is
-    // really a duplicate is recognisable as one by the time collisions are
-    // resolved. Self receipts never reach here, so the peer is the only user a
-    // 1:1 row can name.
-    // Both sides need it, not just the source: the identity a row names is the
-    // one the peer sent from, which is independent of the key the row was
-    // filed under. A receipt addressed to the surviving thread can still carry
-    // the retiring identity.
-    diesel::sql_query(
-        "UPDATE OR IGNORE message_receipts SET user_jid = ?1 \
-         WHERE device_id = ?2 AND chat_jid IN (?1, ?3) AND user_jid = ?3",
-    )
-    .bind::<Text, _>(dest)
-    .bind::<Integer, _>(device_id)
-    .bind::<Text, _>(src)
-    .execute(conn)?;
-
-    // No "keep the furthest state" pass here, unlike reactions: receipts are
-    // keyed per state, so a side holding `read` and a side holding `delivered`
-    // are two facts about one message rather than two candidates for one row.
+    // Receipts, unlike reactions, get no "keep the furthest state" pass: they
+    // are keyed per state, so a side holding `read` and a side holding
+    // `delivered` are two facts about one message rather than two candidates
+    // for one row. What the merge has to settle instead is that both the chat
+    // key and the *peer's identity* are being unified at once — a 1:1's receipt
+    // names whoever the peer sent from, which is independent of the key the row
+    // was filed under, so one person can be spread across four combinations of
+    // (chat, user). Self receipts never reach here, so the peer is the only
+    // user a 1:1 row can name.
     //
-    // Whole-key collisions — same message, user AND state on both sides — are
-    // the one case the move cannot decide for itself. The two rows were
-    // recorded independently under two identities, so neither side is
-    // automatically the earlier one, and `OR IGNORE` below keeps the
-    // destination's. Pull the earlier instant over first, so the survivor is
-    // the first time that state was reported rather than whichever identity
-    // happened to win the merge direction.
+    // Every statement below binds `?1` device_id, `?2` src, `?3` dest.
+    //
+    // Fold the instants first, over all four combinations at once. Doing it
+    // before anything is moved or renamed means the passes that follow are
+    // discarding exact duplicates rather than deciding between them: whichever
+    // row survives already carries the earliest time that state was reported.
+    // Neither identity is automatically the earlier one — the merge direction
+    // is chosen by chat activity, which says nothing about who saw it first.
     diesel::sql_query(
         "UPDATE message_receipts SET ts_ms = (SELECT MIN(s.ts_ms) FROM message_receipts s \
-          WHERE s.device_id = message_receipts.device_id AND s.chat_jid = ?2 \
-            AND s.msg_id = message_receipts.msg_id AND s.user_jid = message_receipts.user_jid \
+          WHERE s.device_id = message_receipts.device_id \
+            AND s.chat_jid IN (?2, ?3) AND s.user_jid IN (?2, ?3) \
+            AND s.msg_id = message_receipts.msg_id \
             AND s.receipt_type = message_receipts.receipt_type) \
-         WHERE device_id = ?1 AND chat_jid = ?3 AND EXISTS \
-         (SELECT 1 FROM message_receipts s WHERE s.device_id = ?1 AND s.chat_jid = ?2 \
-           AND s.msg_id = message_receipts.msg_id AND s.user_jid = message_receipts.user_jid \
-           AND s.receipt_type = message_receipts.receipt_type \
-           AND s.ts_ms < message_receipts.ts_ms)",
+         WHERE device_id = ?1 AND chat_jid IN (?2, ?3) AND user_jid IN (?2, ?3)",
     )
     .bind::<Integer, _>(device_id)
     .bind::<Text, _>(src)
     .bind::<Text, _>(dest)
     .execute(conn)?;
+
+    // Now the identity, on both sides: a receipt addressed to the surviving
+    // thread can still name the retiring one.
     diesel::sql_query(
-        "UPDATE OR IGNORE message_receipts SET chat_jid = ? WHERE device_id = ? AND chat_jid = ?",
+        "UPDATE OR IGNORE message_receipts SET user_jid = ?3 \
+         WHERE device_id = ?1 AND chat_jid IN (?2, ?3) AND user_jid = ?2",
     )
-    .bind::<Text, _>(dest)
     .bind::<Integer, _>(device_id)
     .bind::<Text, _>(src)
+    .bind::<Text, _>(dest)
     .execute(conn)?;
-    diesel::sql_query("DELETE FROM message_receipts WHERE device_id = ? AND chat_jid = ?")
+    // A row already under `dest` whose renamed form collides with an existing
+    // one is skipped by `OR IGNORE` above, and the `chat_jid = src` sweep at
+    // the end cannot reach it. Left alone it would outlive the merge still
+    // naming the retired identity — one peer read back as two users, the exact
+    // failure this reconciliation exists to prevent. Its instant is already
+    // folded in, so it is a pure duplicate by now.
+    diesel::sql_query(
+        "DELETE FROM message_receipts WHERE device_id = ?1 AND chat_jid = ?3 AND user_jid = ?2",
+    )
+    .bind::<Integer, _>(device_id)
+    .bind::<Text, _>(src)
+    .bind::<Text, _>(dest)
+    .execute(conn)?;
+
+    diesel::sql_query(
+        "UPDATE OR IGNORE message_receipts SET chat_jid = ?3 WHERE device_id = ?1 AND chat_jid = ?2",
+    )
+    .bind::<Integer, _>(device_id)
+    .bind::<Text, _>(src)
+    .bind::<Text, _>(dest)
+    .execute(conn)?;
+    diesel::sql_query("DELETE FROM message_receipts WHERE device_id = ?1 AND chat_jid = ?2")
         .bind::<Integer, _>(device_id)
         .bind::<Text, _>(src)
         .execute(conn)?;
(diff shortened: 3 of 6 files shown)


jlucaso1:
Pushed `54e5b3b`. Both Codex P2s were real; Greptile's P1 is mostly the same defect, and the part that remains is deliberate.

**Alias resolution vs. status advancement (Codex, store.rs:1793).** Correct, and it was the root defect. I read ownership off the status update's row count, so a receipt that advanced nothing — a replay, or one arriving *behind* the state already recorded — left `relocated` empty and filed under the wire key even when the message plainly sat under the counterpart. Which chat owns the receipt is a question about where the message is, so it now asks that directly, on the miss path only, where a lookup was already due. Pinned by `a_dm_receipt_behind_the_current_status_still_files_by_alias`: a `Read` followed by a late `Delivered`, where the second advances nothing and previously landed on the wrong key.

**Earliest instant on merge (Codex, lid.rs:324).** Also correct, and my comment there was wrong — I claimed the destination's timestamp was already the earlier one "by the same rule that wrote it". That rule holds *within* one chat key; a split pair has two independent writers, and the merge direction is decided by chat activity, which says nothing about which identity saw the receipt first. The earlier instant is pulled over before the move now.

Worth noting for anyone reading the diff: the collision needs one sender addressing the thread by each of its identities in turn, since `user_jid` otherwise tracks `chat_jid` and the rows never collide. My first attempt at a test used matching chat/sender and proved nothing. The real shape is a PN/LID transition, and `merge_keeps_the_earlier_instant_for_a_state_both_sides_recorded` builds it — verified to fail without the fix (keeps `1700000800`, should keep `1700000200`), not just to pass with it.

**Orphan rows for unmatched ids (Greptile P1).** Half of this is the bug above and is fixed: an id whose message exists anywhere now files against that key. What remains is the case where the id matches under *neither* key, where the receipt is still recorded against the wire key. That is deliberate, and dropping it would be a regression:

- It is the pre-existing behaviour for groups, which wrote unconditionally. Nothing here made it worse.
- A receipt can legitimately arrive **before** its outgoing row is materialized — the same race #1142 fixed for server acks, where discarding the early arrival lost the server's timestamp permanently. Dropping the receipt would lose the instant with no way to recover it, whereas keeping it means the row is already correct once the message lands under that key, which is the common case.
- The cost is bounded: rows for a message that never materializes are collected by the same per-chat satellite prune that already handles reactions, and `receipts()` for an id with no message is a query for something the caller does not have.

So the trade is "a row that may be pruned later" against "a timestamp that can never be recovered", and I would rather keep the timestamp. Happy to revisit if you'd prefer the stricter rule for both chat kinds — that would be a deliberate change to group behaviour too, so I did not make it unilaterally.

109 tests pass in both feature modes; clippy clean.

---

jlucaso1:
Pushed `f9f04e9` — a test rather than an argument, since Greptile is holding at 3/5 on the unmatched-id fallback and my last reply only asserted that it was deliberate.

`a_receipt_that_arrives_before_its_message_is_not_lost` demonstrates why it exists:

1. A `Delivered` receipt for `OUT-DM-EARLY` arrives while no chat holds that id.
2. `message()` confirms there is nothing to attach it to yet.
3. The host calls `record_outgoing` afterwards.
4. `receipts()` returns `Delivered @ 1_700_000_200` — the instant survived.

That ordering is not hypothetical: `Event::Receipt` is dispatched on the socket-read path while a host records the message it sent whenever it gets around to it, which is exactly the race #1142 had to build a deferred-ack queue for. Under the stricter rule the row would be discarded at step 1, and nothing re-sends the receipt once the message lands at step 3 — so the delivered instant would be gone for good, in a PR whose entire purpose is to stop losing that instant.

So the trade is a row that the per-chat satellite prune may collect later, against a timestamp that cannot be recovered at all. The fallback was the one part of the write path with nothing asserting it, which is fair grounds for reading it as an unchecked insert; it is specified and pinned now.

Still happy to switch to the stricter rule if you'd prefer it — it would change group behaviour too, which is why I'm not making that call unilaterally.

Unrelated, for the record: **Semver Checks (informational)** is red here as it is on `main` — 30 findings in `wacore`, 2 in `wacore-binary`, 10 in `waproto`, and zero naming `chat-store`, the only crate this PR touches. Same `v0.6.0 -> v0.6.0` baseline drift as on #1151.

110 tests pass in both feature modes; clippy clean.

---

jlucaso1:
Pushed `30da2ad`. All three of this round's Codex P2s were real.

**Earliest instant, not first processed (line 1851).** Correct, and inconsistent with what I'd just done one function away: `on_conflict_do_nothing` kept whichever report arrived first, while the alias merge resolved the equivalent collision with `MIN(ts_ms)`. Receipts don't arrive in time order — an offline queue drains after the live socket — so a delayed report of a state can land behind a later one and arrival order decides what message info shows. The write lowers an existing row to the earlier instant now.

**Early aliased receipts (line 1820).** Also correct, and it exposed a real gap in the fallback I'd defended two comments ago. `chat` here is `receipt.source.chat.to_non_ad_string()` — unrouted, unlike the self-read branch just above it, which goes through `route_chat_key`. So a receipt arriving by LID for a peer whose thread is PN-keyed filed against a chat that never comes into existence, and the message routed elsewhere when it landed. It resolves to the existing thread now: `route_chat_key`'s rule minus its merge, since a receipt should file itself against a chat rather than reshape one. Worth noting `route_chat_key`'s own doc names this hazard class — "rows split across both keys (the state receipts dropped under the wrong identity leave behind)".

**Canonicalizing the participant (line 1847).** Real, but not fixable where you suggested, and my first attempt proved it. I tried normalizing `user` at write time to the thread's key; the existing merge test caught it immediately, because when the thread itself is split *both* keys exist and the two sides still disagree. The deeper problem is that the mapping may not exist yet when the first receipt lands, so no write-time rule can know the peer's other identity.

The merge can, and it is the only place that ever reconciles the two. It rewrites `user_jid` on both sides — not just the source, since the identity a row names is independent of the key it was filed under — before resolving collisions, so a row that is really a duplicate is recognisable as one by the time they're compared. Self receipts never reach that path, so the peer is the only user a 1:1 row can name.

Three tests, each verified to fail against `f9f04e9` and pass on `30da2ad` rather than merely to pass:

```
an_out_of_order_receipt_lowers_the_recorded_instant ... FAILED
an_early_aliased_receipt_files_against_the_existing_thread ... FAILED
merge_folds_a_split_peer_identity_into_one_user ... FAILED
```

113 tests pass in both feature modes; clippy clean in both.

---

jlucaso1:
Pushed `cb8a0af`, answering the narrowed objection — that "the available evidence does not establish that these ownerless rows survive the existing per-chat pruning lifecycle until their messages arrive." That's a fair thing to want established rather than asserted, so here it is.

Satellite pruning has exactly two entry points, and both are the user removing the conversation:

```
storages/chat-store/src/store.rs:1124   Event::DeleteChatUpdate  -> delete_chat_rows(...)
storages/chat-store/src/store.rs:1148   Event::ClearChatUpdate   -> delete_chat_rows(...)
storages/chat-store/src/store.rs:2744   fn delete_chat_rows      (the only definition)
```

There is no periodic sweep, no size or age retention, and nothing time-based anywhere in the crate. So a receipt waiting for its message is collected only if the user deletes or clears that chat in the interim — and in that case dropping it is correct, because the conversation it belonged to is gone.

Both halves of that lifecycle are now tests rather than an inference from the absence of other callers:

- `a_receipt_that_arrives_before_its_message_is_not_lost` — the row waits, then `record_outgoing` lands and `receipts()` returns the instant.
- `clearing_a_chat_collects_a_receipt_still_waiting_for_its_message` — a `ClearChatUpdate` in the interim takes it, as it should.

114 tests pass in both feature modes; clippy clean.

For completeness on the one remaining red check: **Semver (informational)** has now failed identically on all four heads. I diffed the failure sets between runs — byte-identical, 30 in `wacore`, 2 in `wacore-binary`, 10 in `waproto`, zero naming `chat-store`, which is the only crate this PR touches. It's the same `v0.6.0 -> v0.6.0` baseline drift that is red on `main`.

---

jlucaso1:
Pushed `7bd4829`. Six findings this round across CodeRabbit and Codex; five were real and are fixed, one I'm skipping with a reason.

**The retained receipt was never applied to its row (Codex, store.rs:1848).** The best catch of the round, and it exposed a hole in my own test. `insert_message` starts an outgoing row at `Pending` and the status only ever advances on a receipt arriving *after* the row — so a retained early receipt left `receipts()` saying Delivered and `message()` saying Pending, permanently, since nothing re-sends it. `a_receipt_that_arrives_before_its_message_is_not_lost` passed anyway because it only asserted `receipts()`. It asserts the row's status now, and a freshly inserted outgoing row adopts the state its receipts already record.

**The written thread wasn't invalidated (CodeRabbit major, Codex P2 — both, independently).** Correct. `cs.message_chats` carried the wire key, plus the counterpart only when something relocated; an early aliased receipt writes under neither. A subscriber watching the real thread saw nothing until an unrelated event touched it. Now announced, and the test asserts it through `subscribe()` rather than reading the table — which is the gap CodeRabbit pointed out in the test itself.

**`OR IGNORE` stranded a twin under the destination key (CodeRabbit major).** Real, and my `merge_folds_a_split_peer_identity_into_one_user` couldn't reach it: it puts Delivered on one identity and Read on the other, so there is no whole-key collision to skip. A row already under `dest` naming `src`, whose renamed form would duplicate one that's there, is skipped by the rename — and the `chat_jid = src` sweep never reaches it, because it was never under `src`. It outlived the merge naming the retired identity: one peer read back as two users, the exact failure the rewrite exists to prevent.

Rather than patch that one collision, I reordered the block: instants are folded across **both** keys and **both** identities up front, which makes every later pass a discard of exact duplicates instead of a choice between them, then the skipped twin is dropped. That also subsumes the narrower `MIN` pass I added last round. `merge_drops_the_twin_left_by_a_same_state_collision` is the case you asked for — same state, both identities, one already under the surviving key, retiring identity holding the earlier instant.

**Ownership under the addressed key (CodeRabbit, outside-diff).** Right, and the reasoning is right too: a delete can retire the `chats` row while its messages await cleanup. The miss path asked only the counterpart, so a row present under the addressed key but already at or past this state fell through to "unowned" and could be filed against a different thread. It asks the addressed key first now, separately from whether that chat still exists.

**Lazy counterpart resolution (CodeRabbit, nitpick).** Fair, and the comment claiming the consistent path "stays free" had become false — three extra indexed lookups per DM receipt. Counterpart resolution now sits behind the miss and thread resolution behind an actually-unowned id.

**Bind ordering (CodeRabbit, outside-diff).** Adopted while rewriting the block: every receipt statement in `merge_split_chat` is now `?1` device_id, `?2` src, `?3` dest, stated once above them.

Three tests verified to fail against `cb8a0af` and pass on `7bd4829`:

```
a_receipt_that_arrives_before_its_message_is_not_lost ... FAILED
an_early_aliased_receipt_files_against_the_existing_thread ... FAILED
merge_drops_the_twin_left_by_a_same_state_collision ... FAILED
```

115 tests pass in both feature modes; clippy clean in both.

---

jlucaso1:
Pushed `9de7683`. Both findings were real.

**The symmetric collision (CodeRabbit, lid.rs:381).** Correct, and the walk is exactly right — I fixed one side and left the other. `(chat=src, user=src)` whose renamed form already exists under `src` is skipped by `OR IGNORE`, missed by a sweep looking only at `dest`, and then carried to `dest` intact by the chat rename, arriving as a second user on the merged thread.

Your generalization is the part that matters and I've taken it: past the rename, naming `src` *is* proof of a collision, because that is the only way a row keeps the retiring identity. So every survivor is a duplicate wherever it sits, and both keys get swept. `merge_drops_a_same_state_collision_left_on_the_retiring_side` pins it, alongside the destination-side case from last round.

**Receipts that outrun the mapping (Codex, store.rs:1856).** Also real, and a door I hadn't checked. A receipt can beat the PN/LID mapping as well as its message; with no mapping to consult it can only file under the identity it was addressed by, and if the peer's thread is keyed by the other one the message lands there instead. Nothing reunited them — the chat merge needs two chats and a receipt creates none — so the row stayed behind, invisible to the message that owned it and out of reach of a clear or delete of the real thread.

Materialization is where the true key is finally known, so `adopt_recorded_receipts` now pulls waiting rows over from the counterpart key, folding instants first for the same reason the merge does.

Worth recording: my first version of that test asserted the wrong thing and failed for the right reason. With no chat rows at all, `route_chat_key` picks the LID side, so message and receipt both landed there — consistent, not stranded. The scenario needs the counterpart thread to already exist, which the test now sets up.

Both new tests verified to fail against `7bd4829`:

```
a_receipt_predating_the_alias_mapping_is_claimed_when_the_message_lands ... FAILED
merge_drops_a_same_state_collision_left_on_the_retiring_side ... FAILED
```

117 tests pass in both feature modes; clippy clean.

---

jlucaso1:
Pushed `2e5f123`. One of this round's two findings is fixed; the other I'm deferring, with reasoning.

**Adoption reconciled the thread but not the peer (CodeRabbit, store.rs:2625).** Real, and a defect in the code I added last round. `adopt_recorded_receipts` folded and moved by `chat_jid` while leaving `user_jid` untouched, so a waiting row naming one of the peer's identities and a settled row naming the other both landed on the message as separate users — the same failure the chat merge exists to close, reached through the adoption path instead. Your prescription was right: it now follows `merge_split_chat`'s sequence exactly, scoped to one message — fold instants across both keys *and* both identities, rename the identity to the key the message is stored under, drop the collision survivors, then move the chat key. Groups never reach it, since a group JID has no counterpart. `adoption_folds_the_peer_identity_as_well_as_the_thread` verified failing against `9de7683`.

**Receipts that arrive after the message but before the mapping (Codex, store.rs:1858).** The analysis is correct — I traced it and it holds. An outgoing row exists under PN, a LID-addressed receipt arrives before the mapping is known, so neither ownership lookup can cross the alias and the row files under LID. `adopt_recorded_receipts` runs only at insert, which already happened, and `merge_split_chat` needs a LID chat that a receipt never creates. The PN message stays Pending with a stranded LID row.

I'm not fixing it here, because every available fix is worse than the gap:

- **Device-wide lookup by `msg_id`** would contradict #1142 directly. That change established that a receipt naming a chat must stay inside it, precisely because a same-id row in an unrelated chat was resolving as the target. Re-adding a device-wide fallback for a *named* receipt reintroduces the bug that fixed.
- **Reconcile when the mapping is learned** is the right shape, and it's what your comment asks for. But the mapping lives in the device store, not the chat-store; this crate has no "mapping learned" hook. `reconcile_chat` is the natural entry point and nothing in the repo calls it outside tests, so wiring this properly means touching the client's usync path — a different change from #1154, and one that deserves its own review rather than riding along at the end of this one.

The gap is narrower than the ones already fixed (it needs the receipt to land after the message but before the mapping, and the peer to be mid-PN/LID transition), and it self-heals on any later receipt for that message under the settled key. Happy to open a follow-up issue for the reconcile-on-mapping hook if you'd like it tracked.

118 tests pass in both feature modes; clippy clean.

---

jlucaso1:
Pushed `97a6ded`, which **reverses** the retention I'd been defending for several rounds. @coderabbitai @greptile-apps @chatgpt-codex-connector — you were collectively right and I was wrong; the details matter, so here they are.

**Codex's finding is what settled it.** A delayed or replayed receipt arriving after `ClearChatUpdate` or `DeleteMessageForMeUpdate` re-created the row, so `receipts()` served metadata for a message the user had deliberately deleted. That is a different class of wrong from an untidy orphan.

**My justification does not survive scrutiny.** I kept citing #1142's server-ack race. Re-examined: a server ack fires the instant the server accepts our send, so it genuinely races `record_outgoing`. A **peer** receipt costs a round trip to that peer and back, so it lands well after the send it answers. The window I was protecting is much narrower than I claimed, and nothing at the receipt site can distinguish it from the far likelier reading of the identical state — the message was deleted and its receipts swept with it. I was defending the rare interpretation at the cost of the common one.

**What the code does now.** A receipt naming a message no chat holds is discarded. What that removed:

- waiting-row adoption at insert (`adopt_recorded_receipts`)
- the PN/LID identity fold inside it, added just last round
- guessed-thread resolution and its `chat_exists` lookups
- the invalidation those required

All of it existed to serve rows that no longer exist. **Net −339 lines.**

What stays, because it was never about retention: relocation to the counterpart key (a receipt whose message sits under the peer's other identity still files against it), per-state rows, earliest-instant retention, and the merge's identity reconciliation.

**A bug fell out of the change and is fixed here.** A message present under the addressed key whose status simply did not advance was being classified as absent whenever no counterpart existed, because the existence check sat inside the counterpart branch. Two existing tests caught it immediately — whether a message is here at all is a separate question from whether this receipt changed it.

Tests: three replacing the five that encoded retention, the two behavioural ones verified failing against `2e5f123`:

```
a_receipt_for_a_message_no_chat_holds_is_dropped ... FAILED
a_receipt_arriving_after_a_delete_does_not_resurrect_it ... FAILED
an_aliased_receipt_still_files_against_its_message ... ok   (passes both ways — relocation preserved)
```

116 tests pass in both feature modes; clippy clean.

This also closes Greptile's standing objection across nine passes and Codex's earlier mapping-timing case, since neither has anything left to strand.

---

jlucaso1:
Recording a contradiction rather than acting on it, since the two positions cannot both be satisfied.

On this same commit, @greptile-apps posted:

> The reversal is well-reasoned, and commit `97a6ded` looks correct throughout. […] The change is strictly better than what it replaced — and the ghost-metadata resurrection problem Codex surfaced was the right reason to reverse it.

and, in its scored summary:

> The PR is not yet safe to merge because its orphan-prevention path can permanently discard a valid receipt that precedes message persistence.

Retaining and dropping are the only two options for a receipt whose message no chat holds. Across ten passes this PR has now been marked not-safe-to-merge for **doing each of them**:

| | retain (through `2e5f123`) | drop (`97a6ded`) |
|---|---|---|
| Greptile | not safe — ownerless rows persist | not safe — valid receipt discarded |
| Codex | metadata resurrected after delete | — |
| CodeRabbit | — | verified correct, "deletion must be authoritative" |

So this is a trade-off, not a defect with a fix waiting:

- **Drop** loses a Delivered/Read instant when a receipt genuinely precedes its message — and Greptile is right that a later duplicate then supplies a *later* time rather than the true one. Bounded, self-healing in state if not in precision.
- **Retain** serves metadata for messages the user deleted, and needs adoption, alias folding and thread-guessing machinery to work at all — the ~340 lines this commit removed.

I chose drop because deletion being authoritative is a correctness property while the early-arrival gap is a precision one, and because a peer receipt costs a round trip to the peer and back, so it lands well after the send it answers. I still think that's right, and CodeRabbit's independent verification agrees. But it is your call, and reversing it is `git revert 97a6ded` — the retain implementation is intact in history, tests and all.

If you'd prefer neither compromise, the third option is a deletion tombstone: retain early receipts, record deleted ids so a late receipt for one is discarded. That fixes both objections at the cost of a new table and its own retention policy — more machinery than #1154 called for, and better as its own PR. Happy to open an issue for it.

I'm treating the review loop as closed here unless you'd like a different direction.

---

jlucaso1:
Pushed `85d0f75` with both of CodeRabbit's cleanups. Both were right.

**The redundant UPDATE.** `on_conflict_do_nothing` reports rows inserted, and a row this call just created already holds the incoming instant — so the lowering update only has anything to reconsider when the insert hit a conflict. A first report of a state is the common case on a path that runs for every receipt of every message, so it is now gated on `inserted == 0`.

**The duplicated status advance.** Also right, and the drift risk is not hypothetical: the `from_me = true AND status < status` filter had already been edited twice during this review. It is `advance_status` now, returning whether the row moved, which is the only thing either caller wanted — and it made both call sites shorter than the query they replaced.

Also noting for the record: @chatgpt-codex-connector independently ran the suite against `97a6ded` and reported no further findings, with 25 unit and 109 integration tests passing — matching my local run exactly.

116 tests pass in both feature modes; clippy clean.

---

PR #1158, opened 2026-07-27
fix(appstate): let a diverged collection converge, and stop losing 409'd patches

### LINKED ISSUE #1156: app-state: a collection whose snapshot MAC diverged can never recover - the full_sync escalation is unreachable
### Summary

Once a collection's local LTHash diverges from the chain the other devices compute, every subsequent sync of that collection fails with `patch snapshot MAC mismatch` - forever.
Nothing in the client ever escalates to a snapshot re-fetch, so the collection stays broken across reconnects, restarts, and process reboots.
On our gateway this presented as: marking a chat read succeeds on the wire (receipts go out, the patch send IQ is accepted), but the post-send re-sync fails, so unread counts never converge on the companion; 114 identical failures over one evening, all on `regular_low`, on a session that had been healthy for days.

```
ERROR wa.appstate.send_patch{name=regular_low count=1}:wa.appstate.fetch{name=RegularLow}:wa.appstate.sync{name=RegularLow full_sync=false}: wacore::appstate_sync: error=patch snapshot MAC mismatch
WARN  ...: whatsapp_rust::client::app_state: Failed to re-sync regular_low after patch send: patch snapshot MAC mismatch
```

A process restart does not help.
Right after boot:

```
DEBUG ...: whatsapp_rust::client::app_state: Initializing AppStateProcessor for the first time.
ERROR ...: wa.appstate.sync{name=RegularLow full_sync=false}: wacore::appstate_sync: error=patch snapshot MAC mismatch
```

### Why it can never recover (rev `8bf3466f`)

1. `process_app_state_sync_task` (`src/client/app_state.rs`) forces a snapshot only at genesis:

```rust
let mut state = backend.get_version(name.as_str()).await?;
if state.version == 0 {
    full_sync = true;
}
```

2. The mismatch is raised by patch validation (`wacore/appstate/src/processor.rs:397`, `PatchSnapshotMACMismatch`) and propagates out of `process_parsed_patch_list` with `?` - before the `backend.set_version(...)` at the end of the loop.
So the stored version never advances, never resets, and never becomes 0.
The next attempt re-requests incremental patches against the exact base that has already been proven invalid.
This is a self-perpetuating loop by construction: a MAC mismatch means the local snapshot is not a valid base for patching, and `full_sync=false` retries can only re-derive the same mismatch.

3. `MajorSyncTask::AppStateSync { full_sync: true }` is never constructed anywhere in the repository - the escalation path exists but is dead code.

4. The failure is also invisible to the embedder: `process_sync_task` logs `warn!` and returns, and the send path logs "Failed to re-sync" and returns `Ok(())`.
No event, no error, no signal an embedder could react to.

### Reference behaviour

whatsmeow has three layers of recovery for exactly this state, all absent here:

- `fetchAppState` deletes the stored version before any full sync, and `FetchAppState(name, fullSync=true, ...)` is public, so an embedder can force a clean snapshot re-fetch (`appstate.go`):

```go
cli.appStateSyncLock.Lock()
defer cli.appStateSyncLock.Unlock()
if fullSync {
    err := cli.Store.AppState.DeleteAppStateVersion(ctx, string(name))
    ...
}
...
if version == 0 {
    fullSync = true
}
```

- Decode failures dispatch `events.AppStateSyncError{Name, FullSync, Error}` so the embedder knows the collection is broken.
- It implements the native fatal-recovery protocol (`BuildAppStateRecoveryRequest`, `COMPANION_SYNCD_SNAPSHOT_FATAL_RECOVERY`), which asks the primary device for a fresh copy of the collection.

### Suggested fix

Minimal and self-contained: in `process_app_state_sync_task`, when a non-full sync fails with `PatchSnapshotMACMismatch` / `SnapshotMACMismatch` / `PatchMACMismatch` / `MismatchingLTHash`, reset the stored `HashState` for that collection and retry once with `full_sync = true`.
The snapshot path already rebuilds the hash from scratch (`HashState::default()` + `state = snapshot_state` in `process_patch_list`) and re-persists mutation MACs (`put_mutation_macs` in the snapshot branch), so the reset needs no new machinery.
The reset should also clear the collection's `app_state_mutation_macs` rows, so a stale previous-value MAC cannot poison the first SET fold after recovery.
Emitting an error/recovered event would be a bonus for embedders.

The stale-snapshot guard (`snapshot_is_stale`) does not block this: a diverged client's stored version lags the server (its syncs have been failing while other devices kept patching), and resetting the state to genesis makes the guard moot anyway.

Related: #1157 covers the send-path defect that we suspect created the divergence in the first place; this issue is about the missing recovery once divergence exists, whatever its origin.

### LINKED ISSUE #1157: app-state: send_app_state_patch discards the server response - a 409 conflict is treated as success and the mutation is silently lost
### Summary

`send_app_state_patch` never looks at the server's response.
The app-state send endpoint enforces optimistic concurrency on the `version` attribute: when the patch was built against a stale base, the server answers the IQ successfully but embeds `<collection type="error">` with `<error code="409">` and the conflicting patches.
Because the response is discarded, a 409 is indistinguishable from success: the mutation is silently dropped, `Ok(())` is returned to the caller, and (for `markChatAsRead`) the chat simply stays unread on every other device with no error anywhere.

At rev `8bf3466f`, `src/client/app_state.rs`:

```rust
self.send_iq(iq).await?;   // response never bound, <collection type="error"> never checked

// Re-sync to get the latest state from the server after our patch was accepted.
if let Ok(patch_name) = collection_name.parse::<WAPatchName>()
    && let Err(e) = self.fetch_app_state_with_retry(patch_name).await
{
    log::warn!("Failed to re-sync {collection_name} after patch send: {e}");
}
Ok(())
```

Two things make the stale-base case likely in practice rather than theoretical:

1. `build_patch` bases the patch on whatever `get_version` returns right now - there is no fetch-before-send, so any pending dirty patches (the primary device is usually active) mean the base is stale.
2. Nothing serializes concurrent `send_app_state_patch` calls for the same collection, and nothing serializes them against the background sync worker.
Two quick mark-reads build on the same base version; at most one can win.

### Reference behaviour

whatsmeow parses the response, detects the 409, applies the conflict patches that came back in the error response, and retries the send once (`appstate.go`, `sendAppState`):

```go
if ok && errorTag.AttrGetter().Int("code") == 409 && allowRetry {
    zerolog.Ctx(ctx).Warn().Err(mainErr).Msg("Failed to update app state, trying to apply conflicts and retry")
    patches, err := appstate.ParsePatchList(ctx, &respCollection, cli.downloadExternalAppStateBlob)
    ...
    state, err = cli.applyAppStatePatches(ctx, patch.Type, state, patches, false, &eventsToDispatch)
    ...
    return cli.sendAppState(ctx, patch, false)
}
```

It also returns an error (`ErrAppStateUpdate`) for non-409 collection errors instead of reporting success.

### Observed impact

On our gateway the user marks chats read from the companion while the phone is active.
We saw bursts of read-marks 5-8 seconds apart complete "successfully" while unread state failed to converge, and about 90 seconds after one such burst the session's first `patch snapshot MAC mismatch` appeared on an inbound `regular_low` patch - after which the collection never recovered (filed separately as #1156).
We cannot prove the exact interleaving that diverged the LTHash, but this is the only place on the write path where the server's view and the client's view can disagree without anyone noticing, and a dropped-or-misordered SET on an index both devices write (`markChatAsRead` of the same chat) is exactly the shape that produces a locally-self-consistent but globally diverged hash.

### Suggested fix

- Bind and inspect the response; treat `<collection type="error">` as an error.
- On `code=409`, apply the conflict patches from the response and retry once, per whatsmeow.
- Serialize patch sends per collection with each other (and ideally with the sync path), so concurrent verbs cannot build on the same base.

jlucaso1:
Closes #1156. Closes #1157.

Two defects on the same collection, from the same root cause — the client refusing to look at what the protocol actually told it.

## #1156: what a `snapshotMac` mismatch means

A patch carries two aggregate MACs, and the fix hinges on their being different claims.

`patchMac` is an HMAC over the patch's own bytes under the app-state key, which the server does not hold. It is the only proof of authorship. `snapshotMac` is an HMAC over the *sender's* post-patch ltHash — it agrees only while the receiver's aggregate state is byte-identical to the sender's. Once a collection's ltHash diverges, `snapshotMac` can never match again, for any patch, from any device, forever.

We treated the second as tampering and propagated `PatchSnapshotMACMismatch` out of `process_parsed_patch_list` with `?`. The version never advanced, so the next sync re-requested the same patch against the same base already proven unusable. That is the reported "114 identical failures over one evening, and a restart does not help".

WA Web does not do that. In `WAWebSyncdAntiTampering` (`docs/captured-js/WAWeb/Syncd/AntiTampering.js`), `computeLtHashAndValidatePatch` checks `patchMac` first (fn `K`, fatal on mismatch) and `snapshotMac` second (fn `z`). Where `z` is reached from a patch rather than a snapshot, a mismatch persists `isCollectionInMacMismatchFatal` for the collection, logs `skip fatal after snapshot mac mismatch`, and **returns instead of throwing**; every later patch short-circuits at `if (E && k) return null`, skipping the comparison entirely. Only a mismatch reached from a *snapshot* is fatal — that is the one that escalates to peer recovery.

This is sound, not lenient: `patchMac` is computed over `snapshotMac`, so a valid `patchMac` proves the `snapshotMac` is the one the legitimate sender wrote. A server cannot forge a divergence, only a real one can produce it, and rejecting the patch repairs nothing.

**What changes.** `validate_patch_macs` returns a `PatchMacVerdict` instead of `()`. `patchMac` failures stay errors; a `snapshotMac` mismatch is reported through the verdict, and `process_patch` latches `HashState::mac_mismatch_fatal`, warns once, and applies the patch. The latch suppresses the comparison from then on, exactly like WA Web's short-circuit.

**Where the latch lives.** In `HashState`, so it rides along with the version and hash it belongs to, and in `HashStateWire` as field 4 — rows written before it decode to `false` under proto3, which is the healthy state, so there is no migration. It has to be persistent: an in-memory flag would re-detect the divergence on every restart and re-log it forever.

**One deliberate difference from WA Web:** a snapshot clears the latch. WA Web only ever merges the flag forward, so a collection stays latched for the lifetime of the installation. A snapshot rebuilds the ltHash from scratch, so the new baseline deserves to be trusted again; keeping the latch would mean permanently not validating a collection that has since been repaired.

### The follow-up comment's case

The issue's follow-up reported that a manual reset-and-refetch recovers the base but then dies on a *trailing* patch in the server's post-snapshot window — one the diverged client itself pushed. The same fix covers it, and there is a test for exactly that shape.

Two corrections to that comment, from measuring the current code rather than reading it:

- **"the version row never persists" is not what happens.** `process_patch_list` persists the snapshot's version before it touches the patches (`wacore/src/appstate_sync.rs`), so the collection *does* sit at the snapshot version — measured: v10 in the reproduction. The loop is real; it just re-requests the trailing patch incrementally rather than re-downloading the snapshot.
- **`AppStateError` never reaches the caller.** `crate::runtime::blocking(...).map_err(|e| anyhow!("{}", e))` flattens it to a string, so the `downcast_ref::<PatchSnapshotMACMismatch>()` the issue's suggested fix would need does not work today. Moot now — nothing needs to match on it — but worth knowing before building on it.

**Not done here:** `MajorSyncTask::AppStateSync` is dead code (the whole variant, not just `full_sync: true` — it is matched but never constructed), and the native `COMPANION_SYNCD_SNAPSHOT_FATAL_RECOVERY` peer request is still unimplemented. Neither is needed for this: recovery is now inline, and the peer request is WA Web's answer for a bad *snapshot*, which is a different failure.

## #1157: the send never read the answer

`w:sync:app:state` is optimistic-concurrency. A patch names the base version it was built on; if another device already moved past it, the server answers the IQ *successfully* and reports the failure inside it, as `<collection type="error"><error code="409">`, carrying the patches that won.

`send_app_state_patch` discarded the response node. A 409 was indistinguishable from success: `Ok(())` to the caller, mutation gone, and for `markChatAsRead` the chat simply stayed unread everywhere else with no error anywhere.

WA Web reads exactly that (`ResponseParser.js`, fn `h` → `CollectionState.Conflict{,HasMore}`), then does something we could not: it has no separate send path at all. `RequestBuilder`/`RequestBuilderBuild` embed pending mutations in the *same* sync IQ, so a conflicted collection still goes through `applyAppStateSyncResponse` — applying the winners — and `serverSync` re-queues it while pending mutations remain, up to `y = 5` iterations. The mutation is re-sent on the winner's base rather than dropped.

**What changes.** The response is bound and parsed with `parse_patch_list_ref` (which already knew how to read the error node — only the send path ignored it):

- **409** → apply the patches that came back, rebuild the patch against the new base, resend. Bounded by five attempts, matching `serverSync`'s cap. If the response carried no patches, or applying them failed, a re-sync is the fallback that advances the base; `has_more_patches` forces one too, since the server held patches back.
- **400/404 and anything else** → an error to the caller, never `Ok`.
- **No `<sync><collection>` at all** → accepted, as before. There is no per-collection verdict to read, and a transport-level failure would already have surfaced as `<iq type="error">` from `send_iq`.

**Serialization.** Nothing stopped two `send_app_state_patch` calls for the same collection from reading the same base and racing. One `app_state_send_lock` held across the whole build-send-resolve cycle fixes that — the base version stops being valid only once the send lands. One lock for every collection rather than one per collection matches whatsmeow's single `appStateSyncLock` and WA Web funnelling all collections through one `CollectionsStateMachine`; sends are user-paced, so finer granularity buys nothing.

The send↔sync-worker race is not lock-serialized, deliberately: that is precisely the case the 409 loop now resolves, and it is how WA Web handles it too.

## Tests

Written first, against the current code, and each one failed for the defect's own reason before the fix.

| Test | What it pins |
|---|---|
| `diverged_collection_keeps_applying_authentic_patches` | two authentic patches over a diverged v5 base both apply and advance the version — the reported forever-loop |
| `poisoned_trailing_patch_does_not_strand_a_valid_snapshot` | the follow-up comment's case: snapshot v10 validates, trailing v11 is poisoned, both mutations dispatch |
| `process_patch_latches_divergence_and_keeps_applying` | the patch applies *and* `mac_mismatch_fatal` is set, so it is not re-detected every patch |
| `validate_patch_macs_reports_snapshot_divergence_instead_of_failing` | divergence is a verdict, not an error |
| `validate_patch_macs_skips_snapshot_comparison_once_latched` | WA Web's `if (E && k) return null` |
| `validate_patch_macs_rejects_patch_mismatch_even_when_latched` | latching never weakens `patchMac` |
| `validate_patch_macs_rejects_genesis_{tampered,missing}_patch_mac`, `..._missing_snapshot_mac`, `..._accepts_genesis_valid_macs` | the curated-baseline hardening from #986-#992 survives: omitting an aggregate MAC is still fatal, because omission is not divergence |
| `hash_state_roundtrips` | a latched collection stays latched across a restart |
| `hash_state_row_without_the_latch_field_decodes_as_healthy` | pre-existing rows decode to `false`, no migration |
| `unresolvable_conflict_is_not_reported_as_success` | a server that keeps rejecting ends as an error, after exhausting the five rebuilds |
| `conflict_is_resolved_by_rebuilding_and_resending` | the resolution path: one 409, one rebuild, the mutation survives |
| `fatal_collection_error_is_not_reported_as_success` | 400 is an error and is not retried |
| `response_without_a_collection_verdict_is_accepted` | a terse `<iq type="result"/>` does not read as a rejection |

The `send_patch_response_tests` run a full IQ round trip through `create_iq_test_client` — real stanza on the wire, real response injected — with no socket or mock server, and the whole module finishes in 0.05s.

## Verification

`cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings` clean. `cargo test -p whatsapp-rust --lib` 1244 passed, `-p wacore --lib` 1283 passed, `-p wacore-appstate` 75 passed, `-p whatsapp-rust-sqlite-storage` 65 passed.

`storages/sqlite-storage/proto/wire.desc` and its `.sha256` were regenerated with `scripts/regenerate-wire-desc.sh`, as that file's header requires.

## Risk

The behaviour change worth reviewing carefully is that a `snapshotMac` mismatch on a patch is no longer an error. Its safety rests entirely on `patchMac` being validated first and covering `snapshotMac` — `generate_patch_mac` feeds `patch.snapshot_mac` into the HMAC, so the two cannot be decoupled by anything that lacks the app-state key. If that reasoning is wrong, this is the wrong fix; it is the one claim I would most like a second reader on.


jlucaso1:
Adjudicating the three CodeRabbit findings.

**Test comments still described the bug as unfixed** — valid, they were written TDD-first and not updated. Rewritten: the `#1156` block now says what the two tests pin (the whole-pipeline half; the pure-validation half lives in `wacore-appstate`), and the `#1157` block describes the four response shapes the send path must distinguish instead of claiming they fail.

**Assert the persisted latch** — valid gap. `diverged_collection_keeps_applying_authentic_patches` checked the version round-tripped but not `mac_mismatch_fatal`, so the `process_patch_list` → `set_version` path was uncovered (the latch itself is pinned in the processor unit test, its encoding in the wire test, but not the seam between them). Now asserted after each patch.

**Drop the send lock before the trailing re-sync** — declined, and the reasoning is now a comment on the guard so it does not get re-litigated. Releasing early lets the next send start from a base the in-flight re-sync is about to move, so it builds stale and takes a 409 — trading a short wait for exactly the conflicts this PR exists to reduce. There is no deadlock risk either way (`fetch_app_state_with_retry` never takes this lock), so the only axis is latency-vs-conflicts, and conflicts are the expensive side.

Separately: the earlier **Build & Test** / **Test (voip)** failure was mine — `wacore/tests/appstate_external_mutations_test.rs` asserted the old `Err` contract for a `snapshotMac` mismatch, and I had only run the lib targets before opening. Updated to assert the verdict instead, in both `has_missing_remove` states, since that flag must not change the outcome in either direction. Full workspace suite is green now.

**Semver Checks** is red but was already red on `main`: every entry is a delta accumulated since `wacore-0.6.0` (`SessionError` removed, the `non_exhaustive` sweep, `waproto` derive changes). Nothing from this PR appears in it, and the job is informational.

jlucaso1:
Both Codex findings were real. Fixed in 402a0d34.

**P1 — serialize conflict absorption with collection syncs.** Confirmed, and it is worse than a rollback: `process_patch_list` reads `get_version` once at entry and then writes `set_version` plus `put/delete_mutation_macs` per patch, so an absorb interleaving with a sync can land a version from one and MAC rows from the other. That leaves the ltHash disagreeing with the MAC store — the exact divergence class this PR exists to fix, which would have been an unusually bad thing to introduce here.

The send now takes the collection's own reservation from `app_state_syncing` and holds it across the whole build-send-resolve cycle. `try_begin` is wrong for this — skipping is only free for a sync, where an in-flight one already does the work — so `SyncInFlight` gained `begin`, which waits on an `event_listener` notified from the guard's `Drop` (and from `clear`). Registration happens before the re-check, so a release between them cannot be missed, and cancelling the future reserves nothing.

That also fixes the second half of your finding: the fallback re-sync used to hit `try_begin` → `None` → `Ok(())` and return without moving anything, so the rebuild retried against the unchanged base and burned the attempt budget. The re-syncs inside the send now call the `_inner` variant, since this task already holds the reservation they would otherwise take. `fetch_app_state_with_retry`'s dedup wrapper had no other caller and is gone; the batched path still reserves its own collections.

No deadlock: `app_state_send_lock` is taken before the collection reservation and nothing takes them in the other order.

**P2 — reject malformed collection responses.** Also right, and the distinction matters most for exactly the case you named. Absence and malformation are now separate: a response with no `<sync><collection>` is accepted (no verdict to read, and a transport failure would already have surfaced as `<iq type="error">`), while a collection that IS present but does not parse propagates the error with context. The response is converted once now instead of twice, so this is also marginally cheaper than what it replaces.

New tests: `begin_waits_for_the_holder_instead_of_skipping` (parks on the registry, proven by the listener count rather than a sleep, and is shown to hold the reservation afterwards rather than just having seen it free) and `unreadable_collection_is_not_mistaken_for_an_absent_one`. Full workspace suite and `clippy --workspace --all-targets -- -D warnings` are green.

PR #1162, opened 2026-07-28
ci(miri): check the unsafe decode path under the interpreter

jlucaso1:
Nothing in CI currently observes the workspace's `unsafe`. Clippy does not model aliasing, and a native test only fails if the miscompile has already happened — which is the point at which the bug is hardest to attribute.

## What there is to check

The workspace's load-bearing `unsafe` is concentrated in `wacore-binary`, on the hot decode path:

| Site | The claim being made |
|---|---|
| `node.rs` — `unsafe impl Yokeable for AttrsRef<'static>`, `make`/`transform`/`transform_mut` | three lifetime transmutes over decode output that borrows the inflated buffer |
| `node.rs` — `unsafe impl StableDeref for BytesCart` | the `Bytes` deref target stays put while the wrapper moves, which is what lets `OwnedNodeRef` self-reference |
| `zlib_pool.rs` — `out.set_len(out.len() + produced)` | inflate wrote exactly `produced` bytes into `spare_capacity_mut()`, so that prefix is initialized |

The rest of the tree is clean by construction: `wacore-libsignal` is `#[deny(unsafe_code)]`, and the only other `unsafe` in the workspace is `cfg(target_arch = "wasm32")` `Send`/`Sync` impls plus a counting allocator in an example.

## Scope

Three legs, `fail-fast: false` so each reports its own verdict. Measured on the final commit:

| Leg | Test step | Why |
|---|---|---|
| `wacore-binary --lib` | 37s | the three sites above |
| `wacore-binary --no-default-features --lib` | 1m07s | the portable-SIMD scanners in the decoder/encoder and their scalar fallbacks are separate code; `--no-default-features` is the only way to reach the latter |
| `wacore-noise --lib` | 3m16s | no `unsafe` of its own, but it drives the zero-copy decode over real Noise frames and pulls `curve25519`/`sha2`/AES-GCM, whose unsafe backends this exercises |

They run in parallel, so the gate costs about four minutes of wall clock.

**`wacore-appstate` was tried and dropped.** Miri rejects `inout` 0.2.2's `PaddedInOutBuf::into_out`, which builds a shared slice from a raw pointer that invalidates the `&mut [u8]` still held protected as `cbc::encrypt_padded`'s argument — reached from `wacore-libsignal`'s `aes_256_cbc_encrypt` through the documented API used the documented way, with two RustCrypto layers in between. That is the AES-CBC path every appstate record takes, so the leg cannot go green until the dependency does, and `#[cfg_attr(miri, ignore)]`-ing the ~20 tests that reach it would be deleting the leg while pretending to keep it. The blocker is named in the workflow so the criterion for re-adding it is written down. Full trace in [this comment](https://github.com/oxidezap/whatsapp-rust/pull/1162#issuecomment-5106073475).

Also out: everything Tokio-backed (Miri has no epoll), `e2e-tests`, and `--test` targets. The last is not only a cost decision — `wacore/binary/tests/*_alloc.rs` assert exact allocation counts under a counting global allocator, which is a statement about optimized codegen, not about the interpreter; and `roundtrip_proptest.rs` is fixed at 2048 cases (`ProptestConfig::with_cases` wins over `PROPTEST_CASES`) and writes regression files, which isolation forbids. Running that proptest under Miri at a reduced case count would be a genuinely good UB fuzzer for the decoder, but it needs the case count to come from the environment first — worth a follow-up, not worth weakening this gate to get.

## Decisions in the workflow that are not obvious

- **`components: miri, rust-src`.** `cargo miri setup` compiles the interpreted sysroot from `rust-src`; installing `miri` does not pull it in. Both are published for the pinned `nightly-2026-06-16` on `x86_64-unknown-linux-gnu` — checked against `static.rust-lang.org`'s channel manifest rather than assumed, since Miri is dropped from a nightly whenever it fails to build.
- **No sccache.** `cargo-miri` works by installing itself as `RUSTC_WRAPPER`; sccache wants the same slot. `~/.cache/miri` goes through `Swatinem/rust-cache`'s `cache-directories` instead — the sysroot restores in ~18s rather than being rebuilt — and each leg gets its own cache key because they interpret different feature sets into `target/miri`.
- **`RUSTFLAGS: ""`.** `.cargo/config.toml` sets lld, `--icf=all` and `-Zshare-generics` for this target. Miri never links, so those are dead weight; a set-but-empty `RUSTFLAGS` overrides config rustflags, the same lever `main.yml`'s stable job already pulls.
- **Default `MIRIFLAGS`** — Stacked Borrows, isolation on. Notably *not* `-Zmiri-strict-provenance`: `bytes` reconstructs its tagged `Shared` pointer out of an integer, and `slice_bytes` does its own pointer→integer arithmetic, so that flag would fail on sight, on a class of finding this gate is not looking for. Isolation stays on: nothing in these crates reads a clock or the filesystem, and Miri supplies `getrandom` itself.
- **`push`/`pull_request` rather than a nightly schedule.** These crates change on ordinary protocol PRs, which is exactly when the aliasing claims get invalidated; a nightly result arrives after the PR merged. At four minutes it is cheaper than most jobs already in the matrix.

## What the gate found before it went green

Three aliasing violations, all in dependencies, none in this workspace:

1. **`inout` 0.2.2**, above — production AES-CBC path. Cost the `wacore-appstate` leg.
2. **`zlib-rs` 0.6.6 deflate** — `deflate::end` frees its buffers while a `&mut` into them is still protected. Reached only by the *test helper* that built a fixture with `flate2`; compression is not a path this crate has, inflate is. Fixed by hand-building the fixture as a stored deflate block plus its adler32, which drops the compressor from the interpreted set entirely rather than dropping the test.
3. Neither is a miscompile anyone has observed — they are violations of an experimental aliasing model. Worth reporting upstream; not worth blocking this on.

## Tests

Two of the three unsafe sites were not reachable from any Miri leg as first written, and the second gap was subtler than the first. Both caught in review by @chatgpt-codex-connector.

`OwnedNodeRef` was constructed in exactly one test, inside `#[cfg(feature = "serde")]`, which neither leg enables. And even fixing that would not have reached `AttrsRef`'s hand-written impl: yoke's derive transmutes the whole struct in one go and never calls a field type's methods, so all three of its methods are reachable only through a yoke of `AttrsRef` itself, which nothing built. A gate that reports green over the regression it exists to catch is worse than no gate.

| Test | What it pins |
|---|---|
| `yoked_attrs_ref_survives_make_transform_and_mutation` (new) | a yoke of `AttrsRef` borrowing its key and value from the cart, so `make`, `transform` and `transform_mut` each carry a real lifetime across — our transmutes rather than yoke's generated ones |
| `borrowed_payloads_survive_moving_the_cart` (new) | an `OwnedNodeRef` moved through a `Box` and into a `Vec` still reads its tag, attrs and content back — the whole `StableDeref` claim, which nothing previously exercised |
| `slice_bytes_views_the_backing_buffer_without_copying` (new) | `slice_bytes` returns a view at the same address, not a copy — its pointer arithmetic had no coverage at all |
| `pooled_roundtrip_small_input` (new) | a 1 KB twin keeping `inflate_into_spare`'s `set_len` under the interpreter, on a compressor-free fixture |

The nine `zlib_pool` fixtures are 100 KB–4 MB, and not gratuitously — window refill, the ratio-projected growth path and shrink-on-return only happen at that scale — so they are `#[cfg_attr(miri, ignore)]` behind one comment explaining the pattern. `lookup_matches_reference_under_byte_mutation` joins them: ~3M table probes, one second natively, and it is pure safe code with no raw pointers. It was what left the `wacore-binary` leg still running after twelve minutes; that leg now finishes in 37 seconds.

Each ignore was paid for by a measurement. A proposal to trim the Noise crate's large fixtures too was declined for want of one — the whole leg is 3m16s, and `vec[0u8; 16 MiB]` lowers to `__rust_alloc_zeroed`, which Miri services as a single operation rather than 16 M writes.

## Verification

All three legs green on `89757cd`, at the times in the table above. `cargo fmt --all` and `cargo clippy -p wacore-binary --all-targets -- -D warnings` clean; `cargo test -p wacore-binary --lib` and `--no-default-features --lib` both pass 117 tests, so the `ignore` attributes did not quietly take anything out of the native run.

## Follow-ups, deliberately not smuggled in here

- Report the `inout` and `zlib-rs` aliasing violations upstream to RustCrypto and trifectatechfoundation.
- Decide whether `provider.rs` should use the buffer-to-buffer cipher API instead. Changing production crypto to satisfy an interpreter is a trade-off worth its own discussion, not a detail of adding CI.
- Pin third-party actions to commit SHAs **repo-wide** (raised by @greptile-apps). Fair in the abstract, but `main.yml`, `wasm.yml`, `supply-chain.yml`, `e2e.yml` and `docker.yml` all use `@v6`/`@master`; pinning one new file leaves two conventions and no rule, and the supply-chain surface is unchanged while the other five resolve the same mutable refs. Its own PR, ideally with Dependabot's `github-actions` ecosystem pointed at the pins so they do not rot.

### DIFF
--- a/.github/workflows/miri.yml
+++ b/.github/workflows/miri.yml
@@ -1,0 +1,92 @@
+permissions:
+  contents: read
+name: Miri
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  workflow_call:
+
+env:
+  CARGO_TERM_COLOR: always
+  # Keep in sync with PROTOC_VERSION in main.yml — env is per-workflow, so the
+  # value cannot be shared across files without a repo-level Actions variable.
+  PROTOC_VERSION: '3.25.3'
+  CARGO_INCREMENTAL: "0"
+  # Miri interprets MIR and never links, so the workspace's lld/ICF and
+  # -Zshare-generics rustflags buy nothing here; a set-but-empty RUSTFLAGS takes
+  # precedence over .cargo/config.toml (same lever main.yml's stable job pulls).
+  RUSTFLAGS: ""
+
+jobs:
+  miri:
+    name: Miri (${{ matrix.name }})
+    runs-on: ubuntu-latest
+    # Interpretation costs two orders of magnitude over native. The fixtures the
+    # gate covers are small by construction (the MB-scale zlib ones are
+    # `#[cfg_attr(miri, ignore)]`), but a cold sysroot build alone is minutes.
+    timeout-minutes: 30
+    strategy:
+      # Each leg is an independent UB question; one failing should not hide the
+      # verdict on the others.
+      fail-fast: false
+      matrix:
+        include:
+          # wacore-binary owns the workspace's only load-bearing `unsafe`: the
+          # `Yokeable`/`StableDeref` impls behind `OwnedNodeRef` (two lifetime
+          # transmutes over borrowed decode output) and the `set_len` over
+          # inflate's uninitialized spare capacity in `zlib_pool`. Both are
+          # invisible to clippy and to native tests — nothing observes the
+          # aliasing violation or the uninit read until it miscompiles.
+          - name: wacore-binary
+            cache-key: binary-simd
+            args: -p wacore-binary --lib
+          # The portable-SIMD scanners in the decoder/encoder and their scalar
+          # fallbacks are separate code paths, and `--no-default-features` is the
+          # only way to reach the latter.
+          - name: wacore-binary (no simd)
+            cache-key: binary-scalar
+            args: -p wacore-binary --no-default-features --lib
+          # No `unsafe` of our own, but both drive wacore-binary's zero-copy
+          # decode over real protocol payloads and pull the crypto stack
+          # (aes/sha2/curve25519), whose unsafe backends this exercises.
+          - name: wacore-appstate
+            cache-key: appstate
+            args: -p wacore-appstate --lib
+          - name: wacore-noise
+            cache-key: noise
+            args: -p wacore-noise --lib
+    steps:
+      - uses: actions/checkout@v6
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: nightly-2026-06-16
+          # rust-src is what `cargo miri setup` compiles the interpreter's
+          # sysroot from; installing `miri` does not pull it in.
+          components: miri, rust-src
+      # waproto is in the appstate/noise graphs, and its build script needs protoc.
+      - name: Install protoc
+        uses: taiki-e/install-action@v2
+        with:
+          tool: protoc@${{ env.PROTOC_VERSION }}
+      # No sccache here: cargo-miri drives the build through its own
+      # RUSTC_WRAPPER and the two cannot share that slot.
+      - name: Cache Rust build (registry + target + Miri sysroot)
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-targets: "true"
+          # `cargo miri setup` builds the interpreted sysroot here; without it
+          # every run recompiles core/std from rust-src.
+          cache-directories: ~/.cache/miri
+          # Each leg interprets a different feature set into target/miri.
+          key: ${{ matrix.cache-key }}
+      - name: Build Miri sysroot
+        run: cargo miri setup
+      # Default flags: Stacked Borrows, isolation on. Deliberately no
+      # -Zmiri-strict-provenance — `bytes` rebuilds its tagged `Shared` pointer
+      # out of an integer, which strict provenance rejects on sight and which is
+      # not the class of bug this gate is looking for.
+      - name: Run tests under Miri
+        run: cargo miri test ${{ matrix.args }}
--- a/.github/workflows/miri.yml
+++ b/.github/workflows/miri.yml
@@ -49,17 +49,23 @@ jobs:
           - name: wacore-binary (no simd)
             cache-key: binary-scalar
             args: -p wacore-binary --no-default-features --lib
-          # No `unsafe` of our own, but both drive wacore-binary's zero-copy
-          # decode over real protocol payloads and pull the crypto stack
+          # No `unsafe` of its own, but it drives wacore-binary's zero-copy
+          # decode over real Noise frames and pulls the crypto stack
           # (aes/sha2/curve25519), whose unsafe backends this exercises.
-          - name: wacore-appstate
-            cache-key: appstate
-            args: -p wacore-appstate --lib
+          #
+          # wacore-appstate is deliberately absent: `inout` 0.2.2's
+          # `PaddedInOutBuf::into_out` invalidates the `&mut [u8]` that `cbc`'s
+          # `encrypt_padded` still holds protected, which Miri rejects under
+          # Stacked Borrows. That is on the AES-CBC path every appstate record
+          # takes, so the leg cannot be green until the dependency is fixed —
+          # nothing in this workspace can make it so.
           - name: wacore-noise
             cache-key: noise
             args: -p wacore-noise --lib
     steps:
       - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: nightly-2026-06-16
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -20,6 +20,8 @@ cargo clippy --workspace --all-targets -- -D warnings   # what CI enforces
 
 Workspace clippy takes minutes — pushing and letting CI parallelize the matrix is usually faster. E2E tests (`cargo test -p e2e-tests`) need the mock server running; see `agent_docs/e2e_testing.md`.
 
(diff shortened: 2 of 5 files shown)


jlucaso1:
First run in, and the gate found something on its first outing — just not in our code.

## `Miri (wacore-appstate)` — Stacked Borrows violation inside `inout`

```
error: Undefined Behavior: not granting access to tag <346841> because that would
       remove [Unique for <414328>] which is strongly protected
  --> inout-0.2.2/src/reserved.rs:246:18
246 |  unsafe { slice::from_raw_parts(out_ptr as *const u8, res_len) }

stack backtrace:
  0: inout::reserved::PaddedInOutBuf::<…>::into_out
  1: <cbc::Encryptor<Aes256> as BlockModeEncrypt>::encrypt_padded_inout::<Pkcs7>
  2: <cbc::Encryptor<Aes256> as BlockModeEncrypt>::encrypt_padded::<Pkcs7>
  3: <RustCryptoProvider as SignalCryptoProvider>::aes_256_cbc_encrypt
       at wacore/libsignal/src/crypto/provider.rs:315
```

`into_out` builds a shared slice from a raw pointer that invalidates the `&mut [u8]` still protected as `encrypt_padded`'s argument. Our side of the boundary is the documented API used the documented way — `encryptor.encrypt_padded::<Pkcs7>(&mut out[start..], plaintext.len())` — with two RustCrypto layers (`cipher` 0.5.2, `inout` 0.2.2) between that call and the raw pointer. It is an aliasing-model violation, not an observed miscompile, and it is not fixable from this repo.

It is also not confined to tests: `aes_256_cbc_encrypt`/`_decrypt` are the production path, so *any* crate that encrypts under Miri hits it. `wacore-appstate` reaches it through `create_test_record`, which nearly every `decode`/`processor` test uses — so `#[cfg_attr(miri, ignore)]` there would not be trimming a few expensive fixtures, it would be deleting the leg while pretending to keep it.

**So the gate ships as `wacore-binary` only** (SIMD + scalar), which is where the workspace's own `unsafe` lives and where the check was always going to earn its keep. I will drop the `wacore-appstate` leg with the blocker named in the workflow, so the criterion for re-adding it is written down rather than rediscovered. `wacore-noise` is still running — AES-GCM does not go through the padded path, so it may well be clean; it stays if it is green.

Two follow-ups worth their own issues rather than being smuggled in here: reporting this upstream to RustCrypto, and evaluating whether `provider.rs` should use the buffer-to-buffer API instead. Changing production crypto to satisfy an interpreter is a decision with its own trade-offs, not a detail of adding CI.

## On pinning the actions to commit SHAs (@greptile-apps)

Fair in the abstract, but this workflow is not where that policy should start. `main.yml`, `wasm.yml`, `supply-chain.yml`, `e2e.yml` and `docker.yml` all use `@v6`/`@master`; only `signal-durability-nightly.yml` pins SHAs. Pinning one new file leaves the repo with two conventions and no rule, and the supply-chain surface is unchanged as long as the other six resolve the same mutable refs. Worth doing repo-wide in its own PR — where Dependabot's `github-actions` ecosystem can be pointed at the pins so they do not rot — not for one job here.

The `persist-credentials: false` point (@coderabbitai) I am taking: this job never pushes, the repo already does it in `signal-durability-nightly.yml`, and it costs one line.

---

jlucaso1:
Two findings from @chatgpt-codex-connector, one taken and one declined on measurement.

## Taken: the mutable transmute was uninterpreted — and worse than reported

Correct that `Yoke::with_mut` is called nowhere in the workspace, so `transform_mut` never ran. But yoking a `NodeRef` would not have reached `AttrsRef`'s impl even with a `with_mut` call: yoke's derive transmutes the whole struct in one go and never calls a field type's methods. The hand-written `unsafe impl Yokeable for AttrsRef<'static>` exists to satisfy the derive's bound on the field, which leaves all three of its methods — not just `transform_mut` — reachable only through a yoke of `AttrsRef` itself. Nothing built one, so the interpreter was seeing yoke's generated transmutes and none of ours.

`yoked_attrs_ref_survives_make_transform_and_mutation` builds that yoke, borrowing the attribute key and value straight out of the cart so `make` (on attach), `transform` (on `get`) and `transform_mut` (on `with_mut`) each carry a real lifetime across. 89757cd.

Its compile-time half was already covered: `_assert_attrs_ref_covariant` fails to build if `AttrsRef` ever becomes invariant, which is the property the transmutes rest on.

## Declined: the large Noise fixtures

The cost argument is reasonable a priori and does not survive measurement. `Miri (wacore-noise)` has now run green twice — 4m21s and 4m07s for the whole test step, compile included. That is the entire leg, so the fixtures named are a slice of four minutes, not a reason to reshape another crate's suite.

The mechanism explains it: `vec[0x00; MAX_EDGE_ROUTING_LEN + 1]` and `vec[0u8; FRAME_MAX_SIZE]` lower to `__rust_alloc_zeroed`, which Miri services as a single operation rather than 16 M writes, and both tests then hit their length check immediately. The real cost is `payload(200 * 1024, _)`, a 200 K-iteration `map` — seconds, and it is buying the framing decoder's buffer-reuse path, which is the zero-copy behavior this leg is here to watch.

The three zlib-style ignores in this PR were each paid for by a measurement (hundreds of KB to MB of *interpreted inflate*, and a ~3M-probe table sweep that had a leg still running at twelve minutes). Trimming a four-minute leg on a prediction would be churn in someone else's tests without one.

---

PR #1180, opened 2026-07-29
fix(jid): keep inert agent and integrator out of JID identity

jlucaso1:
Follow-up to #1178, which stopped the decoder writing the AD-JID domain byte into `agent`. That fixed one source of the difference. This stops the difference from mattering at all, wherever it comes from.

## The rule

Neither field is universal:

- **`agent`** is only meaningful where the server renders it. On Pn/Lid/Hosted/HostedLid the wire spells the server as a domain byte instead, and `Display` omits any agent set there.
- **`integrator`** only exists on interop.

Off those servers the fields are inert — nothing encodes them, nothing prints them, nothing reads them back. Two JIDs differing only there address the same thing. They were in the derived `PartialEq`/`Hash` anyway, so a JID took on a different identity depending on where it was parsed from.

`PartialEq` and `Hash` are now written out for `Jid` and `JidRef`, plus the cross-type comparison, all routed through a single `identity_extras` so they cannot drift apart. That matters: a `Hash` disagreeing with `Eq` puts an entry in a `HashMap<Jid, _>` that can never be looked up again.

This is the same rule `is_same_chat_as` already applied to `agent` — it just was not the rule `==` used.

## What broke, and why that is the point

Two tests failed, and both were asserting the old behaviour:

- `test_lid_prekey_lookup_normalization` (wacore) — *"Setup: bundle should NOT exist for requested raw key"*
- `test_session_establishment_lookup_normalization` (whatsapp-rust) — *"Direct lookup of non-normalized JID should fail"*

Both existed to prove that a bundle stays invisible until the lookup key is normalised — which is precisely the workaround this removes. Their doc comments even name the symptom it caused in production: *"No pre-key bundle returned"*. They now assert the bundle is found either way.

I did not weaken them to make the build pass: they were pinning the defect. The first also carried a long stream-of-consciousness comment from its author working out why the lookup failed; that reasoning is now obsolete and was cut.

## Verification

`inert_agent_and_integrator_stay_out_of_identity` covers, for each of the four AD servers: agent and integrator not splitting identity, `Hash` agreeing with `Eq` on every one of those pairs, and the owned/borrowed/cross-type comparisons agreeing. It then checks the other direction — on interop, where the agent *is* rendered, a differing agent still splits, as do `user`, `server` and `device` everywhere.

Suites: **124** wacore-binary, **1469** wacore (`--features voip`), **1281** whatsapp-rust. Clippy clean.

## Scope

Field access is unchanged — `jid.agent` still reads and writes as before, and the 145 struct-literal constructions in the tree are untouched. Making the fields private was the alternative; it would have churned all 145 sites to fix a problem that is not in construction, it is in equality.

Breaking only for code that relied on two JIDs differing by an inert field comparing unequal.

Followed by a cleanup PR that removes the normalisation helper this makes redundant.
## Update: both review findings verified against the captured WA Web bundle

A review raised two objections. I checked both against `docs/captured-js`, which `AGENTS.md` treats as ground truth. One turned out to point at a real pre-existing bug of ours — just not the one it described — and the other rests on a premise the bundle contradicts.

### 1. "phash memo can serve a stale hash" — real problem, opposite cause

The mechanism is correct: `ResolvedGroupDevices::phash` memoises by JID, `participant_list_hash` hashes `push_ad_to`, and `push_ad_to` wrote `self.agent`. Two JIDs equal under this PR could produce different AD strings.

But the fix is not to put the agent back into identity. WA Web's `formatFull` spelling — the one `phashV2` uses — writes a **literal `.0`** in the agent position:

```js
// WAWeb/Phash/Utils.js — phashV2
e.map(function (e) { return e.toString({ legacy: !0, formatFull: !0 }); }).sort().join("")

// WAWeb/Wid.js:127
(t.formatFull === !0 || t.formatIncludeAgent === !0) && (e = ".0");
```

It never reads an agent off the Wid — `WAWebWid` has no agent field at all. So **our `push_ad_to` was the divergence**: any JID carrying an agent produced a phash the server would reject. That is a pre-existing parity bug, independent of this PR, and `swap_pn_lid_namespace` preserves `agent` across namespace conversion, so such JIDs are reachable.

Fixed here: `push_ad_to` writes the agent position as `0`, matching WA Web. Its only production caller is the phash. `ad_form_writes_the_agent_position_as_zero_like_wa_web` pins the spelling and the property the memo depends on — equal JIDs, equal AD strings.

### 2. "integrator must stay in identity off interop" — premise not supported

The objection is that `is_same_chat_as` and `jids_share_user_identity` compare `integrator` unconditionally, so zeroing it elsewhere is contradictory.

In the bundle, `integrator` exists **only** on interop JIDs. It is set by one constructor and read in one branch:

```js
// WA/Wap/Jid.js:35
t.createInteropJid = function (r, o, a) { return new t({ type: e.JID_INTEROP, user: r, device: n, integrator: a }); }

// WA/Wap/Jid.js:65-71 — the JID_INTEROP branch of toString
return _ + "-" + f + ":" + p + "@" + g;   // integrator-user:device@interop
```

There is no integrator on `JID_AD`, `JID_U`, `JID_FB` or plain JIDs, and none anywhere in `WAWebWid`. So on PN/LID a non-zero integrator is not a supported state that this PR merges away — it is a value the protocol never produces there. The unconditional comparisons in `is_same_chat_as` and `jids_share_user_identity` are comparing `0 == 0` on those servers; this PR does not change their outcome.

Worth noting separately (not fixed here): WA Web **renders** the integrator for interop JIDs and our `Display` omits it. That is a real divergence in the other direction, and it belongs in its own change with its own wire evidence.

### DIFF
--- a/src/client/sessions.rs
+++ b/src/client/sessions.rs
@@ -986,21 +986,20 @@ mod tests {
         let mut requested_jid = Jid::lid("123456789");
         requested_jid.agent = 1;
 
-        // 1. Verify direct lookup fails (This is the bug)
+        // The agent is inert on a LID, so it does not hide the bundle: the raw
+        // lookup finds it. This is what `normalize_for_prekey_bundle` was added
+        // to work around.
         assert!(
-            !prekey_bundles.contains_key(&requested_jid),
-            "Direct lookup of non-normalized JID should fail"
+            prekey_bundles.contains_key(&requested_jid),
+            "an inert agent must not hide the bundle"
         );
 
-        // 2. Verify normalized lookup succeeds (This is the fix)
-        // This mirrors the logic change in fetch_and_establish_sessions
+        // Normalising still works, for the callers that still do it.
         let normalized_lookup = requested_jid.normalize_for_prekey_bundle();
         assert!(
             prekey_bundles.contains_key(&normalized_lookup),
             "Normalized lookup should succeed"
         );
-
-        // Ensure the normalization actually produced the key we stored
         assert_eq!(normalized_lookup, normalized_jid);
     }
 }
--- a/wacore/binary/src/jid.rs
+++ b/wacore/binary/src/jid.rs
@@ -766,9 +766,16 @@ impl Jid {
         s
     }
 
-    /// Append the AD-string form (`user.agent:device@server`) to `buf`, for
-    /// callers that batch many JIDs into one shared buffer instead of paying
-    /// a heap `String` per JID (see `participant_list_hash`).
+    /// Append the AD-string form (`user.0:device@server`) to `buf`, for callers
+    /// that batch many JIDs into one shared buffer instead of paying a heap
+    /// `String` per JID (see `participant_list_hash`).
+    ///
+    /// The agent position is the literal `0`, not `self.agent`. That is what WA
+    /// Web writes: its `formatFull` spelling hardcodes `".0"` and never reads an
+    /// agent off the Wid (`WAWebWid`'s `toString`, and its Wid has no agent field
+    /// at all). The only caller is the phash, which the server validates against
+    /// its own computation of the same string — so writing our agent here would
+    /// mean a rejected hash for any JID that happened to carry one.
     #[inline]
     pub fn push_ad_to(&self, buf: &mut String) {
         if self.user.is_empty() {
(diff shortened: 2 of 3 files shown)


jlucaso1:
Superseded by #1182 — same work plus two fixes from the review here, in one diff against `main` instead of a stack.

Carried over: the agent-identity change and the `push_ad_to` alignment with WA Web's literal `.0`.

Changed after the review: `integrator` is no longer normalised (it would have made `==` disagree with `is_same_chat_as` and `jids_share_user_identity`, which reaches quote construction and a retry path), and `sort_dedup_by_device` is now keyed on the same rule as equality — it was still keyed on the raw agent, which would have let the group fan-out build two encryption jobs against one Signal address.

PR #1181, opened 2026-07-29
refactor(jid): drop the prekey-bundle agent normalisation

jlucaso1:
Stacked on #1180 — review that first; this diff is against it, not `main`.

Once an inert `agent` is out of a JID's identity, `normalize_for_prekey_bundle` has nothing left to do. It cloned a JID to clear a field, producing a key already equal to the one the caller held — so every call paid a clone for nothing.

All four call sites did the same thing (store or look up a bundle in a `HashMap<Jid, _>`), so they use the JID directly now and the helper is gone.

`encrypt.rs` cleared the agent on the encryption JID for the same stated reason ("to match how pre-key bundles are stored"). That JID only feeds `reset_protocol_address`, and the Signal address never included the agent (`write_signal_address_to` writes user/device/server), so the clearing had no observable effect left. Removing it drops another clone.

## Deliberately kept

Two clearings that look like the same pattern but are not, and I left them:

- **`normalize_usync_user_jid`** — `validate_user_jid` rejects a non-zero `agent` outright, so it reads the raw field, not identity. Removing the clearing would turn currently-accepted JIDs into `InvalidUserJid` errors.
- **`usync.rs` refresh** — clears `agent` *unconditionally*, including on servers that render it, so it changes the text of the query that goes out. That is wire behaviour, not normalisation.

Also left alone: `app_state.rs` copies `agent`/`integrator` from the primary onto sibling addresses, which is deriving an address rather than scrubbing one.

## One behaviour change worth naming

`parse_prekeys_response` no longer scrubs a stray agent off the parsed key, so the server's byte now rides along on it. It is inert, so the lookups still match — `test_parse_prekeys_response_normalizes_lid_device_jid` still asserts the parsed key is found under the clean JID and not under the malformed one, and now records that the agent survives on the key rather than asserting it was cleared.

Suites: **124** wacore-binary, **1469** wacore (`--features voip`), **1281** whatsapp-rust. Clippy clean. Net −34 lines.

### DIFF
--- a/src/client/sessions.rs
+++ b/src/client/sessions.rs
@@ -493,10 +493,7 @@ impl Client {
         let mut failed_count = 0;
 
         for jid in jids {
-            if let Some(bundle) = prekey_bundles
-                .bundles
-                .get(&jid.normalize_for_prekey_bundle())
-            {
+            if let Some(bundle) = prekey_bundles.bundles.get(jid) {
                 match self
                     .install_prekey_bundle_cached(jid, bundle, &mut adapter, &mut rng)
                     .await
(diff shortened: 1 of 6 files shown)


jlucaso1:
Superseded by #1182, which consolidates this and #1180 into one diff against `main`.

Not carried over: an adversarial review showed two of these removals are observable rather than internal. `PreKeyFetchOutcome::bundles` keys and `RejectedDevice::jid` are public and would start carrying the server's raw agent, and the encryption JID in `encrypt.rs` escapes past `reset_protocol_address` into the public, serializable `IdentityChange.user` event — frozen API per AGENTS.md. Deleting a public method plus changing two public payloads is not worth ~34 lines. Worth revisiting on its own if wanted.

PR #1203, opened 2026-08-03
fix(history-sync): match our own pushname entry by JID user

blaueeiner:
## The bug

`HistorySync.pushnames[].id` is a JID string (`15551234567@s.whatsapp.net`). `extract_own_pushname` compares it verbatim against `own_user`, which is the **bare user part** of our own JID (`process_history_sync(compressed, device.pn.to_non_ad().user, …)` in `src/history_sync.rs`).

The two can never be equal, so `own_pushname` is always `None` and history sync cannot supply our own push name.

whatsmeow parses the same field with `types.ParseJID(pushname.GetID())` before comparing, which is the behaviour this restores.

## Why it matters

For a personal account the push name has exactly one other automatic source: the `setting_pushName` mutation in the `critical_block` app-state collection (a business account also gets it from `business_name` at pairing). When that collection fails to deliver it, history sync is the only remaining path — and it was dead.

Without a push name the client cannot send presence at all (`features/presence.rs` returns `PresenceError::PushNameEmpty`), so the account sits connected and unable to announce itself.

## The test that hid it

`test_nct_salt_and_pushname_coexist` passed because it built `id: Some("0000000000")` — a bare user, which the server never sends. It now uses a JID, and two cases are added: a JID id and the bare form both match, and another user's entry is not taken as our own.

## Verification

`cargo test -p wacore --lib` (1337 passed), `cargo clippy -p wacore --all-targets -- -D warnings`, `cargo fmt --all --check` — all clean. Fixtures use fictitious numbers per AGENTS.md.

## Known limitation, deliberately not addressed here

The comparison is against the PN user only. For a LID-migrated account whose entry is keyed by the LID user this still won't match; fixing that means threading the own LID through `process_history_sync`, which changes the signature and felt out of scope for a comparison fix.

### DIFF
--- a/wacore/src/history_sync.rs
+++ b/wacore/src/history_sync.rs
@@ -758,6 +758,31 @@ fn skip_field(wire_type: u32, buf: &[u8], pos: usize) -> Result<usize, HistorySy
     }
 }
 
+/// `Pushname.id` is a JID string (`15551234567@s.whatsapp.net`), while
+/// `own_user` is the bare user part of our own JID. Compare on the user so the
+/// entry is recognisable; whatsmeow parses the same field with `ParseJID`.
+///
+/// Only a phone-namespaced entry can carry our PN user. A `@lid` entry whose
+/// user happens to be the same digits belongs to a different addressing space,
+/// and taking its name as our own would be wrong — so any other server is
+/// rejected rather than compared.
+///
+/// Parsing is left to `parse_jid_ref` rather than split by hand: it borrows
+/// (no allocation for the entries we discard) and it already knows the forms
+/// this field can take — the legacy `@c.us` spelling of the phone namespace,
+/// and both device suffixes, `user:3` and the legacy dotted `user.3`.
+fn pushname_id_is(id: &str, own_user: &str) -> bool {
+    if !id.contains('@') {
+        // Pre-JID bare form, still accepted.
+        return id == own_user;
+    }
+    let Some(jid) = wacore_binary::jid::parse_jid_ref(id) else {
+        return false;
+    };
+    (jid.server.is_pn_family() || jid.server == wacore_binary::jid::Server::Legacy)
+        && &*jid.user == own_user
+}
+
 // Best-effort: a malformed pushname (optional metadata) must not abort the sync.
 fn extract_own_pushname(data: &[u8], own_user: &str) -> Option<String> {
     let mut pos = 0;
@@ -777,7 +802,7 @@ fn extract_own_pushname(data: &[u8], own_user: &str) -> Option<String> {
                 let len = usize::try_from(len).ok()?;
                 let end = pos.checked_add(len).filter(|&e| e <= data.len())?;
                 let id = smoothutf8::from_utf8(data.get(pos..end)?)?;
-                id_match = id == own_user;
+                id_match = pushname_id_is(id, own_user);
                 if !id_match {
                     return None; // wrong user, skip entirely
                 }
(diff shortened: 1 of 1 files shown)


jlucaso1 (changes requested):
Thanks for this one, and good to have you back after #1010.

I went and checked the server side before reading the diff. In the current WA Web bundle the history sync pushname handler does `t.pushnames.forEach(e => WidFactory.createUserWidOrThrow(e.id || ""))`, so `Pushname.id` really is a WID and never a bare user. whatsmeow agrees, `handleHistoricalPushNames` parses it with `types.ParseJID`. Your premise holds, the old comparison could never match. I also ran the suite on your branch, 1337 passing.

Two things I would like changed before this goes in, both small:

1. Use `Jid` instead of splitting by hand. `extract_lid_mapping` right below (the one you added in #1010) already does `parse::<Jid>()` and reads `user_base()`, and `user_base()` is exactly the `split_once(':')` your helper reimplements. Going through the same path keeps the two neighbours consistent and drops the manual string work.

2. Reject the wrong namespace. Right now the server part is discarded entirely, so `15550000000@lid` matches `own_user = "15550000000"`. The odds are tiny, but it is a silent identity mixup and it costs nothing to avoid: once you have a `Jid`, requiring `is_pn() || server == Server::Legacy` is the same check `extract_lid_mapping` already makes. No perf worry here, I checked that no allocation guard covers this path and it runs once per entry, not per message.

Two optional ones:

- `parity_fixture` still builds `id: Some(own.to_string())` in the bare form, so the streaming vs full buffer parity test never sees a real JID. Would be nice to move it over too, it is the same unrealistic fixture your PR body calls out.
- whatsmeow skips entries whose pushname is `-`. I looked for that sentinel in the bundle and did not find it, so it is not a WA Web rule, but if the server ever sends it for our own entry we would store `-` and announce it. Your call.

Good find, and the "the test that hid it" section made this very easy to review.

For what it is worth, the red `Semver Checks (informational)` is not yours. It is a pre existing `waproto` diff against the published 0.6.0 and the job says outright that it does not block the PR.

blaueeiner:
Fixed in `40e9242`.

The namespace check is a good catch, and it is not only a hardening detail — for a LID-migrated account the same digits can appear as both `…@s.whatsapp.net` and `…@lid`, and `own_user` is the PN user, so a bare user-part comparison could have taken a LID entry's name as ours. Only the phone namespace can carry our PN user, so any other server is now rejected rather than compared:

```rust
let Some((user, server)) = id.split_once('@') else {
    return id == own_user;  // pre-JID bare form, still accepted
};
server == wacore_binary::DEFAULT_USER_SERVER
    && user.split(':').next().unwrap_or(user) == own_user
```

Added `a_non_phone_namespace_is_not_our_entry`, covering `@lid`, `@c.us` and `@newsletter`; the existing cases for a phone-namespace JID and the bare form still pass.

I used `split_once` plus the `DEFAULT_USER_SERVER` constant rather than a full `Jid` parse: this runs per entry inside the protobuf byte-walk, the only thing needed is the server and user, and a parse would allocate for a value thrown away on every non-matching entry.

`cargo test -p wacore --lib` (1338 passed), `cargo clippy -p wacore --all-targets -- -D warnings`, `cargo fmt --all --check` — clean.

blaueeiner:
Both parser findings were right, and the second one was a bug I introduced in the previous round. Fixed in `2760359`.

**`@c.us` is the legacy spelling of the phone namespace, not a different one.** `jid.rs:366` maps it to `Server::Legacy` and the parser's own test uses `("12345.6@c.us", "12345", 6, 0)` as a PN with a device. My previous commit rejected it *and* asserted that rejection in a test — so I had encoded the wrong behaviour and pinned it. That test now lists `@lid`, `@newsletter` and `@g.us`, with a comment saying explicitly why `@c.us` is not among them.

**The dotted device suffix is real too.** `jid.rs:115` — "s.whatsapp.net has no agent in the string form; a trailing dotted number is the legacy device spelling" — so `15550000000.3@s.whatsapp.net` has user `15550000000`, and my `split(':')` left it as `15550000000.3`.

Rather than patch two more cases by hand I switched to the repository parser, which is what the first review round asked for:

```rust
let Some(jid) = wacore_binary::jid::parse_jid_ref(id) else { return false };
(jid.server.is_pn_family() || jid.server == Server::Legacy) && &*jid.user == own_user
```

I had declined that on allocation grounds. That was wrong on the facts: `parse_jid_ref` borrows, so nothing is allocated for the entries we discard — and two hand-rolled bugs in two rounds is the argument for using the thing that already knows the wire forms.

`own_pushname_matches_a_jid_id` now covers `@s.whatsapp.net`, the bare form, `@c.us`, `:3`, `.3@s.whatsapp.net` and `.3@c.us`. The two new spellings both fail against the previous implementation by inspection (`@c.us` fails the server equality, `.3` leaves the user as `15550000000.3`).

`cargo test -p wacore --lib` (1338 passed), `cargo clippy -p wacore --all-targets -- -D warnings`, `cargo fmt --all --check` — clean.

**The `"-"` sentinel I have not implemented, and I would rather ask than guess.** The claim is that a pushname of `"-"` is history sync's missing-name marker and whatsmeow skips it. I can't check it: there is no `"-"` handling anywhere in this repo to follow as precedent, and I have neither the WA Web bundle nor whatsmeow locally, so I would be adding behaviour on an assertion I cannot verify. The risk is symmetric — accept it and an account can announce `name="-"`; skip it and an account whose display name really is `-` never gets one — so it wants your reading rather than my guess. Happy to add the skip in this PR if you confirm the sentinel, or leave it out.

Salientekill:
Please review the reviews and check the fact-checks; one review failed.

PR #1204, opened 2026-08-03
fix(appstate): fail the batched sync when a collection did not sync

blaueeiner:
## The bug

`sync_collections_batched_inner` returns `Ok(())` in two cases where a collection was not synced:

- a per-collection `CollectionSyncError::Fatal` — logged with `warn!`, then `continue`
- `MAX_ITERATIONS` exhausted with patches still outstanding — logged with `warn!`, then falls through to `Ok(())`

The initial bootstrap in `node_io.rs` reads that `Ok` as permission to store `critical_sync_done`, abort the retry watchdog, and `dispatch_connected()`. So a `critical_block` that failed either way leaves the session **connected without the settings that collection carries** — including `setting_pushName`, so the account cannot announce presence — and with no scheduled retry.

The missing-key branch a few lines above already declines to report a false success for exactly this reason:

> Report failure rather than a false success: the initial critical-sync path treats Ok as permission to cancel its retry watchdog and dispatch Connected, which would leave CriticalBlock/CriticalUnblockLow unsynced with no scheduled retry.

The two exits changed here contradict that.

## The change

Both now return `Err`, so the bootstrap keeps its watchdog alive and reconnects to retry, as it does for a missing decode key.

The other callers only log the result, so nothing else changes behaviour:

- `handlers/ib.rs` — server_sync notification
- `handlers/notification/groups.rs`
- `node_io.rs` background non-critical sync

## How this was found

Chasing an account that reported `Connected`, synced 400 conversations, and could not send: the critical sync had returned `Ok` without ever applying `setting_pushName`.

## Verification

`cargo test -p whatsapp-rust --lib` (1320 passed), `cargo clippy -p whatsapp-rust --all-targets -- -D warnings`, `cargo fmt --all --check` — all clean.

### DIFF
--- a/src/client/app_state.rs
+++ b/src/client/app_state.rs
@@ -659,6 +659,7 @@ impl Client {
                 .await?;
 
             let mut needs_refetch = Vec::new();
+            let mut unsynced: Vec<WAPatchName> = Vec::new();
 
             for (mutations, new_state, list) in results {
                 let name = list.name;
@@ -681,6 +682,7 @@ impl Client {
                         }
                         CollectionSyncError::Fatal { code, text } => {
                             warn!(target: "Client/AppState", "Collection {:?} fatal error {}: {}", name, code, text);
+                            unsynced.push(name);
                             continue;
                         }
                         CollectionSyncError::Retry { code, text } => {
@@ -728,6 +730,19 @@ impl Client {
                 );
             }
 
+            // A collection the server refused outright is not synced, and the
+            // caller must not be told otherwise: the initial-bootstrap path
+            // reads Ok as permission to cancel its retry watchdog and dispatch
+            // Connected. A critical_block that fails here that way leaves the
+            // session with no push name and no scheduled retry — the same
+            // reasoning as the missing-key branch above.
+            if !unsynced.is_empty() {
+                return Err(anyhow::anyhow!(
+                    "app-state collection(s) {unsynced:?} returned a fatal error; \
+                     reporting the batched sync as failed"
+                ));
+            }
+
             pending = needs_refetch;
         }
 
@@ -737,6 +752,10 @@ impl Client {
                 "Batched sync: max iterations ({}) reached for {:?}",
                 MAX_ITERATIONS, pending
             );
+            return Err(anyhow::anyhow!(
+                "app-state collection(s) {pending:?} still had patches after \
+                 {MAX_ITERATIONS} iterations; reporting the batched sync as failed"
+            ));
         }
 
         Ok(())


jlucaso1 (changes requested):
The diagnosis is right and I am glad you chased this down. I confirmed the false success: `node_io.rs:1344` reads `Ok` as permission to set `critical_sync_done`, abort the watchdog and dispatch Connected, and the missing key branch at `app_state.rs:644` really does decline to report success for exactly that reason. The two exits you touched are inconsistent with it.

Where I want to push back is the remedy, because I went to the WA Web bundle for both arms and neither one lines up.

**Max iterations, and this is the bigger one.** Our comment at app_state.rs:505 says "Max 5 iterations (WA Web's `C=5` constant)". That is wrong. In the current bundle `WAWebSyncdServerSync` has `var y=5, C=500` and the loop is

```js
for (var a=[], i=[].concat(t), l=0; (l<y || i.length>0 && l<C) && i.length!==0; )
```

Since the body only runs when `i.length !== 0`, the condition collapses to `l < 500`. WA Web's effective cap is 500, not 5. And when it does run out, it marks those collections `ErrorRetry` and hands them to the finite retry state machine (exponential backoff, base 2 from 1s, capped at an hour). It does not treat it as a hard failure.

So our 5 is about 100x tighter than the client we are modelling, and this PR turns hitting it into "never reaches Connected". A large account that legitimately needs more than five rounds connects today with a partially synced collection, and after this it would not connect at all. I would rather we fix the cap first and leave this arm out of the PR.

**The fatal arm.** WA Web does treat a collection 400/404 as terminal, but terminal in a specific way. `ErrorFatal` goes to `moveCollectionsToFatal`, the state machine tick calls `handleSyncdFatal`, and `WAWebSyncdFatal.handleFatalError` waits 5s, sends an `AppStateFatalExceptionNotification` to the primary, then does `socketLogout(LogoutReason.SyncdFailure)`. It gives up and leaves.

What this PR produces is a third thing. The bootstrap returns early without ever clearing `needs_initial_full_sync` (that only happens at node_io.rs:1399, after the return), so the watchdog reconnects at 180s and lands in the same place. That is an unbounded loop, and since `needs_pushname_from_sync` comes from the persisted `push_name.is_empty()` (node_io.rs:1041), it survives a process restart too. Today the same account connects degraded. This trades "connected but missing a collection" for "never connects", which for a genuinely unrecoverable 400 is not an improvement.

I am not defending the current behaviour, your argument for `critical_block` is a good one. But the shape WA Web uses is per collection state (up to date, finite retry, fatal, blocked), not one Ok/Err for the whole batch, and that is the direction I think is worth taking. If you want something smaller in the meantime, a bounded number of attempts before giving up and connecting degraded, or surfacing an event, would at least keep the failure mode finite. We deliberately reconnect instead of logging out (the comment at node_io.rs:1296 says so), so "retry forever" is the one option we should not end up with by accident.

A few smaller things while we are here:

- No tests. This changes whether Connected is dispatched and neither new error path is covered. Worth knowing: E2E does not run on fork PRs by design (e2e.yml:28, no secrets for the mock server image), so nothing in CI exercised this.
- A fatal in one collection now drops the whole iteration's `needs_refetch`, so healthy siblings that still had patches stop being refetched in that call.
- Related hole you did not touch: `sync_collections_batched` also returns `Ok(())` when `try_begin` filters everything out (app_state.rs:539). Same class of false success.

Happy to keep talking this one through, the underlying problem is definitely worth fixing.

blaueeiner:
You are right on both arms, and I checked the parts I could rather than take them on faith.

**The fatal arm produces an unbounded loop — confirmed.** `needs_initial_full_sync` is only cleared at `node_io.rs:1399`, inside the background task spawned *after* the critical sync returns `Ok`. The `Err` path returns at `:1371`, before that, and the watchdog reconnects into the same state. `needs_pushname_from_sync` comes from the persisted `push_name.is_empty()` (`:1041`), so a restart lands there too. For a genuinely unrecoverable 400 that trades "connected, missing a collection" for "never connects", which is worse than what it replaces. greptile and cubic flagged the same thing independently.

**Max iterations.** I can't check `y=5, C=500` myself — I don't have the bundle locally, only the pinned crate source — so I'm taking your reading of `WAWebSyncdServerSync`. If the effective cap is 500 with `ErrorRetry` and backoff behind it, then a cap of 5 is not "the same as WA Web with a smaller number", it's a different mechanism, and turning it into a hard failure would break exactly the large accounts it was meant to protect. Agreed it should come out.

Your two smaller points are right too, and both are mine to own: a fatal in one collection discards that iteration's `needs_refetch` before it's consumed, so healthy siblings stop being refetched; and there are no tests for either new error path, which for a change that decides whether `Connected` is dispatched is not defensible on its own.

So I'm closing this rather than leaving a patch up whose remedy I no longer believe in. The false success is real — I'm not withdrawing the diagnosis — but "return `Err` for the whole batch" is the wrong shape for it, and the right one is the per-collection state you describe (up to date / finite retry / fatal / blocked), which is its own piece of work rather than a fix bolted onto this call.

Two follow-ups I'd be glad to take, in either order:

1. **The cap.** Correcting the comment is unambiguous. The value is less so: WA Web's 500 comes with the retry state machine and its backoff between rounds, while our loop iterates with none, so raising 5 → 500 as a straight swap turns a truncated sync into up to 500 back-to-back IQ rounds. I'd rather do the cap and the backoff together than the number alone — tell me if you'd prefer it split.
2. **Per-collection sync state**, with the fatal path going where WA Web puts it (`AppStateFatalExceptionNotification` then logout) or wherever you want it to land given `node_io.rs:1296` deliberately reconnects instead.

`sync_collections_batched` returning `Ok(())` when `try_begin` filters everything out (`:539`) is the same class and worth folding into (2).

Also noted on E2E not running for fork PRs — that explains the green checks here meaning less than they look.

PR #1205, opened 2026-08-03
fix(appstate): reserve the collection for an AppStateSync task

blaueeiner:
## The bug

Every path that writes a collection's version and mutation MACs takes an in-flight reservation first:

- `sync_collections_batched` — `try_begin`, skipping collections already in flight
- the patch send — `begin`, waiting, because it "must not write the collection's version and mutation MACs while a sync is writing them"

The `MajorSyncTask::AppStateSync` arm of `process_sync_task` takes none. A task-driven sync can therefore interleave its `set_version` and MAC writes with a concurrent batched sync of the same collection, leaving the ltHash disagreeing with the MAC store — the invariant `SyncInFlight` exists to protect.

Nothing in the library enqueues that variant today (the only sender on the channel is history sync), so it is reached through the public `process_sync_task`, which is how a consumer asks for a specific collection. That is what surfaced it: a consumer forcing a `critical_block` snapshot to recover a push name the critical sync never delivered.

## Why full and incremental differ

`try_begin`'s doc says it is right for a sync, "where an in-flight one already does the work and skipping is free". That holds for an incremental sync and **not** for a full one: an in-flight incremental sync asks for patches after the persisted version, while a full sync asks for the snapshot. Skipping would silently turn the caller's request into a no-op — precisely the failure the caller is trying to fix.

So `full_sync: true` waits, and `full_sync: false` skips, matching the batched path. Cancelling the wait strands nothing: nothing is reserved until the guard is returned, and guards release on drop.

## Tests

Two, using the existing `create_test_client_with_name` fixture. The client has no socket, so the sync itself fails fast with `NotConnected` — what is under test is the reservation, not the IQ.

- `a_full_sync_task_waits_for_the_collection` — verified to fail without the fix (the task runs while the collection is reserved), pass with it
- `an_incremental_sync_task_skips_an_in_flight_collection`

## Known trade-off

The worker's intake loop runs non-history tasks inline, so a waiting full sync holds that loop until the collection frees. With no in-library producer for this variant it is currently unreachable, but it is the reason to prefer `begin` only for the full-sync case rather than for both.

## Verification

`cargo test -p whatsapp-rust --lib` (1322 passed), `cargo clippy -p whatsapp-rust --all-targets -- -D warnings`, `cargo fmt --all --check` — all clean.

### DIFF
--- a/src/client/app_state.rs
+++ b/src/client/app_state.rs
@@ -427,6 +427,22 @@ impl Client {
                     .await;
             }
             MajorSyncTask::AppStateSync { name, full_sync } => {
+                // Reserve the collection like every other sync path does.
+                // Unreserved, this writes the same version and mutation-MAC
+                // rows as a concurrent `sync_collections_batched`, leaving the
+                // ltHash disagreeing with the MAC store.
+                //
+                // Waits rather than skipping, like the patch send and unlike
+                // the batched path. `try_begin`'s "an in-flight one already
+                // does the work" holds only when the holder is an equivalent
+                // sync, and neither condition is guaranteed here: the holder
+                // may be a patch send, which does not fetch at all, and a full
+                // sync asks for the snapshot while an incremental one asks for
+                // patches after the persisted version. Skipping on either
+                // mismatch turns the caller's request into a silent no-op.
+                // Cancelling the wait strands nothing — the reservation only
+                // exists once the guard is returned.
+                let _guard = self.app_state_syncing.begin(name).await;
                 if let Err(e) = self.process_app_state_sync_task(name, full_sync).await {
                     log::warn!("App state sync task for {name:?} failed: {e}");
                 }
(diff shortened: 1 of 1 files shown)


jlucaso1 (changes requested):
This is the cleanest of the three. The premise checks out: the `AppStateSync` arm was the only path writing `set_version` (app_state.rs:895) and the mutation MACs without a reservation, while the batched path takes `try_begin` (528) and the patch send takes `begin` (1199). No self deadlock either, since the arm calls `process_app_state_sync_task` directly rather than `fetch_app_state_with_retry_inner`. Your full vs incremental reasoning is right, and I appreciate that you wrote down the intake loop trade off instead of hiding it.

I verified the tests rather than take them on faith. I ran them on your branch, then removed only the guard block and kept both tests:

- `a_full_sync_task_waits_for_the_collection` fails without the fix, exactly as you said.
- `an_incremental_sync_task_skips_an_in_flight_collection` still passes without the fix. The socketless client fails fast with `NotConnected` and returns well inside the 5s timeout either way, so that test is not currently asserting anything.

So the one change I want: make the incremental test observable, or drop it. A `send_iq` hook or a counter asserting that no sync request went out while the collection was reserved would do it. A test that passes with and without the change is worse than no test, because the next person reads it as coverage.

One suggestion, not blocking: `begin` has no deadline, so if a holder ever wedges, the full sync arm waits forever, and you already point out that the intake loop runs non history tasks inline. `try_begin` at least fails fast. A bounded wait would keep the worst case bounded too.

On the bot comments about incremental syncs being dropped behind a patch send: the underlying fact is real, `send_app_state_patch` does hold the same registry and does return without a re-sync on the rejection path (1291-1295). But that is not something your PR introduces, `sync_collections_batched` at line 528 already skips on a patch send holder the same way. It is a pre existing gap in `SyncInFlight` not distinguishing holders, and it deserves its own issue rather than a fix here.

blaueeiner:
Fixed in `b6a930c` — the incremental branch now waits too, so the arm is a plain `begin`.

**On the greptile finding (valid, and it changes the design).** I had justified `try_begin` for the incremental case with `try_begin`'s own doc: "an in-flight one already does the work and skipping is free." That holds only when the holder is an equivalent sync, and here it need not be — the patch send holds the same reservation and never fetches, so an incremental task arriving during a send was silently dropped, and the caller had no way to know. Two conditions had to line up for the skip to be safe (holder is a sync, and it is fetching the same thing), so the skip is gone:

```rust
let _guard = self.app_state_syncing.begin(name).await;
```

This also removes the asymmetry the review was circling. The full/incremental distinction still matters for correctness — a full sync asks for the snapshot, an incremental one for patches after the persisted version — but it is no longer a reason to treat the two differently at the reservation.

**On the CodeRabbit test comment.** I don't think that one holds, so I have not changed the test. The concern is that with the reservation removed, the no-socket client would run `process_app_state_sync_task`, fail fast with `NotConnected`, and still pass. It would not: the assertion is `!task.is_finished()` while the reservation is held, so a task that ran and returned early — for any reason, `NotConnected` included — fails it. I verified this rather than assuming, by reverting just the guard and re-running:

```
thread 'a_full_sync_task_waits_for_the_collection' panicked:
the task ran while the collection was reserved
```

The timeout after `drop(held)` is the liveness half, not the assertion.

`an_incremental_sync_task_skips_an_in_flight_collection` became `an_incremental_sync_task_also_waits` with the same structure.

`cargo test -p whatsapp-rust --lib` (1322 passed), `cargo clippy -p whatsapp-rust --all-targets -- -D warnings`, `cargo fmt --all --check` — clean.

The trade-off in the PR description gets slightly wider: an incremental task can now hold the worker's intake loop as well. Still unreachable in-library (nothing enqueues this variant), and I would rather block briefly than drop a sync the caller asked for.

blaueeiner:
All five inline comments in this round anchor to `8b02aac8`, which is one commit behind the head — they queued against the version before `b6a930c`.

The finding four of them share (greptile, cubic, chatgpt-codex, coderabbit: "incremental sync drops behind patch send") is the one that commit already fixed, and their agreement is a good sign it was worth fixing. The `try_begin` branch they are reading is gone; the arm is now an unconditional `begin`, so an incremental task waits behind a patch-send holder instead of skipping. cubic's pointer to the early return at `app_state.rs:1313` is exactly the case that made the old skip unsafe.

The test comment (coderabbit and cubic) is the one I don't think holds, and it also refers to a test that no longer exists — `an_incremental_sync_task_skips_an_in_flight_collection` became `an_incremental_sync_task_also_waits`. The premise is that a no-socket client would fail fast with `NotConnected` and pass anyway. It would not: the assertion is `!task.is_finished()` **while the reservation is held**, so a task that ran and returned early for any reason fails there. I checked rather than argued it — reverting just the guard produces:

```
thread 'a_full_sync_task_waits_for_the_collection' panicked:
the task ran while the collection was reserved
```

Nothing to change here as a result. A re-review against `b6a930c` would be worth having if the bots can be re-triggered.

jlucaso1 (approved):

PR #1207, opened 2026-08-04
fix(appstate): report what a batched sync actually achieved

jlucaso1:
Follow-up to #1203, #1205 and #1206, closing the gaps they left and the ones #1204 surfaced but did not fix.

## The false success

`sync_collections_batched` returned `Ok(())` in cases where a collection had not synced:

- a per-collection `CollectionSyncError::Fatal` (400/404), logged and skipped
- `MAX_ITERATIONS` exhausted with patches still outstanding
- every collection filtered out by the in-flight dedup
- a response that omits a requested `<collection>`, which parses fine and lands nowhere

The initial bootstrap reads that `Ok` as permission to store `critical_sync_done`, abort the retry watchdog and `dispatch_connected()`. A `critical_block` failing any of those ways left the session connected without the settings it carries, `setting_pushName` included, and with nothing scheduled to retry.

#1204 proposed turning them into `Err`. That is worse: the bootstrap returns before `needs_initial_full_sync` is cleared, so a 400/404 reconnects into the same state every 180s, forever, and since `needs_pushname_from_sync` derives from the persisted push name it survives a restart.

## What this does instead

`sync_collections_batched` returns a `BatchedSyncOutcome` with four buckets (`synced`, `fatal`, `retryable`, `skipped`) and each caller decides what a partial result means. `Err` is reserved for genuinely global failures.

The buckets are not interchangeable. `skipped` means an equivalent sync already holds the collection and is doing this work — the only case a caller may ignore. Everything else is a real miss.

| Bootstrap outcome | Action |
| --- | --- |
| all synced | cancel watchdog, dispatch Connected |
| any `fatal` | stop retrying it, connect degraded, emit `AppStateSyncFailed`, hand the batch's *recoverable* misses to the background sync |
| `retryable` / `skipped` only | keep the watchdog armed, reconnect and retry |

Everything else goes through `report_background_sync`, which logs, emits the event, and schedules a spaced retry.

### Why degraded rather than logout or retry

- **WA Web** (`WAWebSyncdFatal`): waits 5s, sends `AppStateFatalExceptionNotification` to the primary, then `socketLogout(SyncdFailure)`. It can — it owns the re-link UI.
- **whatsmeow** (`appstate.go`): returns the error to the caller. Exposes `BuildFatalAppStateExceptionNotification` and `BuildAppStateRecoveryRequest` as builders and never calls them; the first one's doc warns it logs out every linked device.
- **zapo** (`src/appstate/`): classifies into the same states and returns the outcome, no automatic destructive action.

Two of three keep the policy out of the library. Ending a consumer's session is not ours to decide, and `node_io.rs:1296` already recorded that choice.

Recovery via `COMPANION_SYNCD_SNAPSHOT_FATAL_RECOVERY` is not a route here — WA Web gates it off for exactly the collection that matters:

```js
if (t === CollectionName.CriticalBlock)
  return { shouldPerformRecovery: !1, reason: RECOVERY_STATUS_ENUM.COLLECTION_UNSUPPORTED };
```

## Two bugs from re-reading `WAWebSyncdServerSync`

**`ErrorRetry` must not be refetched in the same run.** WA Web routes `ErrorRetry`, `ErrorFatal` and `Blocked` to `doneCollections`; only `ConflictHasMore`, `SuccessHasMore` and a conflict with pending mutations go to `refetchCollections`. We pushed retryable errors back into the refetch list, which the corrected cap would have turned into 500 attempts at a collection that had just failed.

**The cap is 500, not 5.** The loop is `(l < y || (i.length > 0 && l < C))` with `y = 5`, `C = 500`; the body only runs while something is left to refetch, so it collapses to `l < 500`. Rounds are back-to-back there too — the syncd backoff spaces retries of a *failed* collection, not pages of one still making progress.

Taking `ErrorRetry` out of the loop matches WA Web, but WA Web hands it to a retry state machine afterwards and we had none, so on its own that change would strand a collection after a single 500. `schedule_app_state_retry` is that machine, with WA Web's spacing (1s, doubling, capped at an hour). The persisted two-day expiry is not here; it needs a store column.

## The redesign

Nine review rounds on this branch were almost entirely one class of mistake: work that survives an await, and the connection or the deadline moving underneath it. Generation was read ad hoc in ten places, the deadline checked in three, and the shared bootstrap gate written from five — each caller responsible for remembering. Two forgot, in opposite directions, and several of the fixes introduced the next round's bug.

**`SyncScope`** carries the connection a piece of work belongs to and the clock it runs against. **`Client::admits`** is the only place either question is asked, and it says *why*: `Retired` and `Expired` are not interchangeable — a retired scope must not write or publish, an expired one on a live connection still may, which is exactly when the bootstrap has to stay armed.

**`settle_bootstrap`** does the same for the shared gate. There are now no direct writes to `needs_initial_full_sync` outside it.

**The apply boundary is per collection.** `process_patch_lists` persists each list as it goes, so a batch admitted at the start could still be writing its third collection well outside the scope. `process_one_patch_list` (a pure extraction in `wacore`, no behaviour change) lets the client stop between collections, capping what a lost scope can commit at one instead of the whole batch.

Ten ad-hoc generation reads and three deadline checks became two reads inside the scope machinery, three explicit rebinds, and one predicate.

Work that must outlive a planned reconnect rebinds instead of being cancelled — nothing re-issues a dirty-bit or `server_sync` request once its trigger is consumed. What does not carry over is authority: a rebound run can no longer settle the bootstrap it was scheduled by. And a retry holds through the reconnect window rather than attempting inside it, because `process_app_state_sync_task` answers `Ok(())` at its own shutdown guard without contacting the server, and reading that as success would drop the request silently.

## Reservations

`SyncInFlight` records whether a **sync** or a **patch send** holds a collection. Skipping is only sound behind an equivalent sync; the batched path was skipping behind patch sends, which take the same reservation and never fetch, so the sync was dropped and the patches that prompted it went unfetched.

The bootstrap never skips: it has to know a collection is synced before dispatching Connected, and "someone else is trying" is not that. Waits are bounded, because the worker's intake loop runs non-history tasks inline. A wait that runs out is `retryable`, not `skipped` — nobody is covering that collection.

## Push name

History sync's `-` placeholder no longer becomes our push name. Storing it makes `push_name` non-empty for good, so `needs_pushname_from_sync` goes false, the bootstrap stops looking for the real name, and the account re-announces `-` on every reconnect. whatsmeow skips the same value in two places. The authoritative `setting_pushName` mutation has its own path and does not come through here, so an account genuinely named `-` still gets it.

## Found by reading the diff whole, not by review

- `schedule_app_state_task_retry` carried `schedule_app_state_retry`'s doc comment while the latter had no summary line.
- Three rationale blocks had accumulated superseded copies of themselves, documenting one branch with two contradictory explanations.
- One doc still promised retries never fire against a new socket, which stopped being true when they started rebinding.

## Tests

Each verified to fail without its fix rather than merely pass with it. The scope invariants:

- `a_scope_stops_admitting_when_its_connection_or_clock_goes` — and that retired outranks having time left
- `a_retired_scope_cannot_move_the_bootstrap_gate` — in both directions
- `an_expired_scope_may_still_arm_the_gate`
- `rebinding_reports_whether_the_connection_moved`
- `only_a_deadline_marks_the_bootstrap`
- `a_planned_reconnect_is_not_a_completed_sync`

The outcome and request-set behaviour:

- `a_refused_collection_is_reported_fatal_not_synced`
- `a_retryable_collection_is_not_refetched_in_the_same_run` — asserts exactly one IQ reaches the wire
- `collections_held_by_another_sync_are_reported_skipped` — and nothing reaches the wire
- `a_patch_send_holder_is_waited_out_not_skipped`
- `a_collection_the_response_omits_is_not_reported_synced`
- `an_empty_response_leaves_every_collection_unsynced`
- `a_duplicate_collection_is_dropped_before_it_is_applied`
- `a_collection_requested_twice_is_reserved_once`
- `an_unrequested_collection_in_the_response_is_dropped`
- `an_expired_deadline_stops_the_batch_before_it_reserves`
- `an_outcome_from_a_retired_connection_is_not_published`
- `only_a_fully_synced_outcome_settles_the_bootstrap`
- `the_absent_pushname_sentinel_is_not_a_name` / `a_name_containing_a_dash_is_kept`

Two things are deliberately unasserted and say so in the code: that the scheduler refuses to run for a retired generation (two probes both passed with the guard removed — the wrongly-attempted sync fails anyway on a socketless client), and the requeue of a top-level batch failure (the observable is a detached task that sleeps first). A test that passes without its fix is worse than none.

Also not covered: the bootstrap's own branch, which needs a full connection to drive. The outcome it reads is covered; the branch on it is not.

## Verification

`cargo test -p wacore --lib` (1340), `cargo test -p whatsapp-rust --lib` (1346), `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all --check` — all clean.

## Left for later

The persisted finite-retry expiry. WA Web keeps a `finiteFailureStartTime` per collection and promotes to fatal after two days of continuous failure. That needs a store schema change, so the retry here is per-connection only.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,6 +1,7 @@
 mod accessors;
 mod adapters;
 mod app_state;
+pub(crate) use app_state::SyncSettles;
 mod builder;
 mod context_impl;
 mod device_registry;
--- a/src/client.rs
+++ b/src/client.rs
@@ -1091,7 +1091,7 @@ pub struct Client {
     /// the stability reset from erasing a deliberate penalty (WA Web `cancelReset`).
     pub(crate) backoff_reset_suppressed: Arc<AtomicBool>,
 
-    pub(crate) needs_initial_full_sync: Arc<AtomicBool>,
+    pub(crate) needs_initial_full_sync: Arc<app_state::BootstrapGate>,
 
     pub(crate) app_state_processor: Mutex<Option<Arc<AppStateProcessor>>>,
     pub(crate) app_state_key_requests: Arc<Mutex<HashMap<Vec<u8>, wacore::time::Instant>>>,
--- a/src/client/app_state.rs
+++ b/src/client/app_state.rs
@@ -711,7 +711,7 @@ impl Client {
                     &outcome,
                     self.is_ready.load(Ordering::Relaxed),
                 );
-                self.schedule_app_state_retry(outcome.retryable, settles);
+                self.schedule_app_state_retry(outcome.retryable, generation, settles);
             }
             Err(e) => self.log_sync_error(label, &e),
         }
@@ -772,15 +772,22 @@ impl Client {
 
     /// `settles` says what recovering these collections finishes; see
     /// [`SyncSettles`].
+    ///
+    /// `generation` is passed in rather than read here: callers validate it
+    /// before deciding to schedule, and dispatching the failure event in between
+    /// runs consumer handlers synchronously. Re-reading it after that would bind
+    /// a retired connection's retries to whatever replaced it — and for
+    /// [`SyncSettles::InitialSync`], let them stand down the replacement's
+    /// bootstrap gate.
     pub(crate) fn schedule_app_state_retry(
         self: &Arc<Self>,
         collections: Vec<WAPatchName>,
+        generation: u64,
         settles: SyncSettles,
     ) {
         if collections.is_empty() {
             return;
         }
-        let generation = self.connection_generation.load(Ordering::SeqCst);
         let client = self.clone();
         self.runtime.spawn_detached(Box::pin(async move {
             let mut pending = collections;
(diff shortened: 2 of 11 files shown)


Salientekill:
Ta pouco de review ou quer mais?

PR #1222, opened 2026-08-07
perf(sqlite): route Signal reads through the read pool

jlucaso1:
## Summary

`db_semaphore` is `Semaphore::new(pool_size)` and `pool_size` defaults to 1. That default is correct for writes and I am not touching it: two deferred transactions that both read and then write deadlock on SQLite's upgrade, and `busy_timeout` cannot break it.

What the same permit should not be doing is governing reads. A session, identity or sender-key miss on the decrypt path waits out whatever write is in flight, `put_sessions_batch` from a write-behind flush included. Meanwhile the reader pool already existed: `read_pool_size` builds a second, `query_only` pool, and on `main` today nothing outside `SharedSqlite::read` and `resource_report` touches `self.reads`. Turning the knob on buys the chat store concurrency and buys the Signal path nothing.

Rebased onto `05b3c880`. The base moved from `3da38692` through #1227 and #1229; neither touched `storages/sqlite-storage/`, and the rebase was clean.

## Audit

Read-only is necessary and not sufficient. The criterion is **what the caller does with a stale-absent or stale-valued answer**: if it recovers, the read can move; if the answer is promoted into a cache that overwrites what it is handed, is sent on the wire, or fails an operation outright, it stays on the write queue.

### Cold-load guard coverage

Routing widens the window in which a load can span a flush plus an eviction. For the reads whose consumer sits behind `SignalStoreCache`, the thing that closes that window is the guard in `main`. Traced one read at a time rather than assumed, and two of the five did not land where this branch had recorded:

| Routed read | Consumer | Guard | Sufficient because |
| --- | --- | --- | --- |
| `get_session_for_device` | `checkout_session`, `peek_session`, `has_session` (all three read `backend.get_session`) | #1229 | re-checks `incarnation` and `removed_since` after the I/O, so a superseded record is retried rather than installed |
| `load_identity_for_device` | `get_identity` | #1229 | same re-check on the identities cache |
| `get_sender_key_for_device` | `get_sender_key` | #1226 | same re-check, landed earlier |
| `has_session` (backend method) | `Device::contains_session` via `check_session_exists` | none, and none needed | the direct store path, not the cache; the only production caller logs the answer in all three branches |
| `has_signal_state_for_user` | `SignalStoreCache::has_state_for_user` | **none** | see below, moved back to the write queue |

Two corrections this produced:

- **`has_signal_state_for_user` leaves the migrated set.** Its consumer is not one of the guarded five. `has_state_for_user` checks both caches for any matching key and otherwise asks the backend with no removal-seq re-check, and its callers in `lid_pn.rs` use a `false` to skip the PN to LID session migration outright. Nothing retries the skip. It goes back on the write queue.
- **`has_session`'s justification was wrong, though its verdict was not.** The cache's `has_session` reads `get_session`, not this method, so #1229 never covered it. The only path here is `Device::contains_session`, whose single production caller logs. The reason recorded next to it is now the caller, not the guard.

### Re-validation against the new base

#1227 touched three files this audit cites (`device_registry.rs`, `chat_actions.rs`, `groups.rs`) and #1229 touched `signal_cache.rs`. Re-checked the cited behaviours: the unconditional `device_registry_cache.promote` on a backend hit and the `InvalidRequest("no app state sync key available")` on a `None` latest-key id both still hold, unchanged. Neither commit changes what a caller does with a stale answer for any of the other reads. The audit is current with `05b3c880`, not inherited from `3da38692`.

Migrated:

| Method | Stale answer leads to | Verdict |
| --- | --- | --- |
| `get_session_for_device` (`get_session`) | cache miss, then a pre-key bundle fetch | migrate |
| `load_identity_for_device` (`load_identity`) | cache miss, identity re-learned | migrate |
| `get_sender_key_for_device` (`get_sender_key`) | cache miss, SKDM redistributed | migrate |
| `has_session` | a debug log line | migrate |
| `load_prekey` | a consumed-not-yet-deleted row is readable; that is the normal duplicate-pkmsg path | migrate |
| `load_prekeys_batch` | same row re-offered in the upload window | migrate, see note |
| `get_max_prekey_id` | both callers hold `prekey_upload_lock` and the prior write is awaited | migrate |
| `load_signed_prekey`, `load_all_signed_prekeys` | rotation re-reads; `rotate_key` retries | migrate |
| `get_app_state_version_for_device` (`get_version`) | collection re-syncs from an older version | migrate |
| `get_app_state_mutation_mac_for_device` (`get_mutation_mac`) | mutation treated as new, re-applied | migrate |
| `get_app_state_mutation_macs_batch_for_device` (`get_mutation_macs`) | same, batched | migrate |
| `has_same_base_key` | retry path re-processes; nothing cached | migrate |
| `get_group_metadata` | read and compared, not promoted | migrate |
| `get_all_tc_token_jids` | enumeration for expiry sweeps, no per-JID decision | migrate |
| `device_exists`, `load_device_data_for_device` | startup, no concurrent writer | migrate |

**Note on the prekey reads.** `buffer_consumed_prekey` records the id in `removed_prekeys` and `flush` deletes the row later, once the promoted session is durable, so a consumed prekey is readable for the whole write-behind interval by design; `signal_adapter.rs` has a test pinning exactly that. Routing extends the window by one transaction. For `load_prekey` the question is moot: re-reading a still-present prekey is the duplicate-pkmsg path. For `load_prekeys_batch` the consequence differs, since the upload window re-offers rows to the server, but that is reachable today through the buffer interval with no routing change. It is a prekey-durability question, not a routing one.

Not migrated (11), each with its reason in `ON_THE_WRITE_QUEUE`:

- **`has_signal_state_for_user`** -- new in this pass, see the guard table above.
- `get_app_state_sync_key_for_device` (`get_sync_key`) -- answers a peer's `AppStateSyncKeyRequest`; on `None` the handler returns an orphan `MessageField`, so a stale absent read tells the peer we lack a key we hold, on the wire.
- `get_latest_app_state_sync_key_id_for_device` -- `send_app_state_mutation` turns `None` into `InvalidRequest` and the user's action fails with nothing retrying behind it.
- `get_msg_secret_with_ts` (and `get_msg_secret`, which delegates) -- a miss is terminal; the reaction, vote or edit is dropped. History sync seeds secrets via `put_msg_secrets` directly, bypassing the live write-behind buffer.
- `get_lid_mapping`, `get_pn_mapping` -- `alternate_msg_secret_jid` resolves the peer's other namespace through these into the lookup above, with no cache in front on that path.
- `get_all_lid_mappings` -- the startup warm-up feeds these into `LidPnCache::add_guarded`, whose LID side replaces unconditionally.
- `get_sender_key_devices` -- initializes `sender_key_device_cache`; a stale `has_key = true` drops the SKDM for a device that asked for redistribution.
- `get_devices` -- promoted into `device_registry_cache` unconditionally on a miss, overwriting a newer entry.
- `get_tc_token` -- feeds `prepare_privacy_token`'s scheduling decision; a stale read issues a duplicate token.
- `get_pending_inbound` -- read-only SQL, but its `with_retry` loop exists so a transient BUSY does not fail closed into a redelivery.

`take_sent_message` and `store_received_tc_token` read and write in one `immediate_transaction` and were never candidates. `snapshot_db` and `resource_report` are not queries.

Ordering caveat, stated once: most of the write-queue holds above are ordered by the single pooled connection at the default `pool_size`, not by the permit itself, since several writers check a connection out without taking it. Above `pool_size = 1` that ordering weakens. `get_sender_key_devices` is the exception: every writer of `sender_key_devices` (`set_sender_key_status`, `clear_sender_key_devices`, `clear_all_sender_key_devices`, `delete_sender_key_device_rows`) goes through `with_retry`, so that one holds at any pool size. The rest predates this PR and fixing it means changing write serialization, which is out of scope.

## Default-config behaviour

An earlier revision of this description said the `read_pool_size = 0` path was unchanged. That was wrong.

Fourteen reads ran on a raw pooled connection **without** the permit before this branch. Routing them through the helper put them behind it, which measured about 25% on p50 at the default profile -- a real regression for a knob nobody has turned on. `c780e8b` fixed it: the single-connection branch checks the connection out directly and takes no permit, because the one pooled connection is already both the serialization and the snapshot. Adding a permit there only serialized the `spawn_blocking` dispatch that the pool wait was overlapping.

After that fix the default path differs from `main` by exactly one heap allocation: both do `spawn_blocking(pool.get(); query)` with no permit, and this branch boxes the closure first. See Cost for what the measurement can and cannot say about it.

## E2E

An earlier revision listed an E2E regression as a blocker. It was not one, and the mistake was mine: I saw `E2E Tests` fail on one commit, checked that `main` was green, and concluded regression without looking at this branch's own history.

| commit | E2E |
| --- | --- |
| `bcabb05` | pass |
| `9923ffb` | pass |
| `2018115` | pass |
| `77906ae` | pass |
| `68f1a49` | pass |
| `39b9a38` | **fail** |
| `6c68e3a` | pass |
| `4f9bb6d` | pass |

One red run among seven greens on the same branch is a flake, and the five greens before it would have said so immediately. Every E2E run since has been green as well, including on the rebased head.

Independently, the suite was run against a local mock server on both trees: 165 run / 163 passed / 2 failed on this branch and the same two on `main` (`presence::test_presence_available`, `privacy_tokens::test_restricted_presence_subscribe_requires_tctoken`), i.e. local-mock drift. I could not reproduce that myself -- this environment has no mock server -- so that half is someone else's evidence and the conclusion above rests on CI history.

## Dependency

Resolved. The guard landed in **`05b3c880`** (#1229) and this branch is rebased on it.

The cold-load re-check in `checkout_session`, `peek_session`, `has_session` and `get_identity` previously asked only whether an entry exists, so it could not tell "never written" from "written, persisted and removed"; a load spanning a flush plus eviction installed the older record, and for `checkout_session` that record goes to the cipher and advances the ratchet. #1229 stamps `incarnation` and `removal_seq` around the unlocked backend read and retries, bounded by `UNLOCKED_COLD_READ_ATTEMPTS`, with a locked fallback. That is what makes routing safe for the three reads whose consumers sit behind it, per the coverage table above.

## Changes

- `SqliteStore::read_query` -- erases the closure, then routes: the deferred snapshot via `SharedSqlite::read` when a reader pool is configured or a wider write pool can take one; otherwise a direct connection checkout, taking the permit only when the pool is wide enough that the connection is no longer the serializer.
- `with_semaphore` erased the same way, so the write queue emits one body per return type instead of one per call site.
- The read-only methods above routed, 11 deliberately held back with a written reason each.
- Five chunked batch writers wrapped in one transaction each, so chunking is a parameter-limit workaround and not a commit boundary.
- `get_msg_secret` delegates to `get_msg_secret_with_ts` rather than repeating its filter chain a column narrower.
- Tests: happy and absent per migrated method at `read_pool_size` 0 and 4; a write refused by `query_only` on a reader connection, and the fallback's lack of that net asserted rather than assumed; a read while the write permit is held; a read against an open write transaction; a multi-statement read under a wider write pool, verified to fail without the snapshot; a shared-cache store declining that snapshot, with a write required to commit under a parked read and verified to fail without the guard; the chunked-write race, verified to fail without the transaction; a bidirectional source scan, so both a newly misrouted read and a stale allowlist entry fail.

## Cost

In-process harness, file-backed store, `taskset -c 0-3`, release, deleted before this PR and never declared as a `[[bench]]`. Not `divan`: the number that matters is read tail latency with a write in flight, and `divan` measures isolated throughput. Both tables re-measured on `05b3c880`.

### The win

16 concurrent `get_session` per round, 40 rounds, over 512 seeded 2 KiB rows, 3 rounds. Case B adds a continuous `put_sessions_batch` of 1500 rows. Both configurations run in one process, so drift hits them equally.

| | Case A p50 | Case A p99 | Case B p50 | Case B p99 |
| --- | --- | --- | --- | --- |
| `read_pool_size = 0` | 594-639 us | 1.34-2.40 ms | 86.4-91.4 ms | 101-106 ms |
| `read_pool_size = 4` | **408-541 us** | 1.04-1.61 ms | **590-602 us** | 4.73-25.5 ms |

Case B is the point: p50 ~88 ms down to ~0.59 ms, a factor of ~150. That is far above any noise this box produces.

### The default profile, where this PR could make something worse

16 concurrent `get_sender_key`, 7 rounds per run, `read_pool_size = 0`, this branch against `main`. Run in alternating passes because a single ordering confounds machine drift with the code:

| Pass | ran first | `main` p50 median | branch p50 median | branch/`main` |
| --- | --- | ---: | ---: | ---: |
| 1 | `main` | 532.1 us | 538.6 us | 1.01 |
| 2 | `main` | 609.2 us | 647.9 us | 1.06 |
| 3 | branch | 572.7 us | 545.7 us | 0.95 |
| 4 | branch | 456.9 us | 628.5 us | 1.38 |

**The difference is not resolvable on this box, and I am not claiming it is zero.** The ratio has no consistent sign across the four passes, and the within-run spread is 1.65x for `main` and 1.70x for the branch -- larger than the effect being looked for. A first, non-alternating attempt suggested about 9%; alternating showed that was ordering drift, since the second run of any pair was systematically slower.

What can be said without the measurement: after `c780e8b` the default path is `spawn_blocking(pool.get(); query)` with no permit on both sides, and the branch adds one `Box` for the erased closure. That is the whole delta. It is stated as a structural argument, not as a measured equivalence.

For contrast, the 25% regression `c780e8b` fixed *was* resolvable at this noise floor: 569-641 us against 730-796 us, non-overlapping across rounds. Nothing of that size remains.

### Resolved by instruction counting

Walltime on this box cannot see an effect this small, but instruction counting can. Measured off-CI by driving the CodSpeed instrumentation directly under Valgrind (`CODSPEED_ENV=1 valgrind --tool=callgrind`, which makes the harness report `Measured` instead of `Checked` without needing a token or an upload), on a temporary bench that exercises the default profile (`read_pool_size = 0`) against a file-backed store. Three runs per side, medians in instructions:

| bench | `main` | branch | delta | spread on `main` |
| --- | ---: | ---: | ---: | ---: |
| `get_session` miss | 271,618 | 270,566 | **-1,052 (-0.39%)** | 758 |
| `get_session` hit | 105,458 | 104,310 | -1,148 (-1.09%) | 1,265 |
| `load_identity` | 98,764 | 98,965 | +201 (+0.20%) | 307 |

Instruction counts are not automatically deterministic, and it is worth being explicit about that: running the *same* binary twice differed by about 550 instructions until the bench's runtime was switched from `multi_thread` to `current_thread`, because work-stealing decisions vary run to run. The residual spread on `main` is 307-1,265. So only the miss has a delta larger than its own noise, and the reading is an upper bound rather than a point estimate: **the default profile changes by at most about 1%, and where the signal exceeds the noise its sign is a reduction, not a regression.**

That is consistent with the structural argument above and puts a number on it: one extra `Box` costs hundreds of instructions, not the 9% the first non-alternating walltime attempt suggested. The bench is temporary and is not part of this diff.

### Binary size

The gate failed at one point at +39.88 KiB .text against a 32 KiB budget. An `nm` symbol diff of the `demo` example put it in three buckets: `read_erased` at +25.1 KiB over 186 instantiations, `SharedSqlite::read` at +16.1 KiB, and `with_semaphore` at +9.6 KiB as its instantiation count went 77 to 188. Folding `read_erased`'s two non-snapshot branches into one and erasing `with_semaphore`'s closure fixed it. Current reading against the `05b3c880` baseline: **+26.12 KiB .text** (budget 32) and **+31.31 KiB stripped** (budget 64).

## Validation

```
cargo fmt --all
cargo test -p whatsapp-rust-sqlite-storage --lib   # 75 passed
cargo test -p whatsapp-rust --lib                  # 1391 passed
```

No existing test was relaxed. One was strengthened: `a_shared_cache_store_gets_no_snapshot_even_with_a_wider_write_pool` asserted only the `snapshot_safe` flag and stayed green when the guard was deleted from the routing predicate, so it now parks a read mid-flight and requires a concurrent write to commit. That is a contract change in the stricter direction and it found a real limit -- the park has to outlast `with_retry`'s ~310 ms backoff, or the shared-cache lock is retried away rather than observed.

### DIFF
--- a/storages/sqlite-storage/src/shared.rs
+++ b/storages/sqlite-storage/src/shared.rs
@@ -105,7 +105,7 @@ impl SharedSqlite {
 /// [`StoreError`] deliberately has no such conversion (callers choose how a
 /// database error is classified), so both travel in one enum and unwrap on the
 /// way out.
-fn read_snapshot<T>(
+pub(crate) fn read_snapshot<T>(
     conn: &mut SqliteConnection,
     f: impl FnOnce(&mut SqliteConnection) -> Result<T>,
 ) -> Result<T> {
--- a/storages/sqlite-storage/src/shared.rs
+++ b/storages/sqlite-storage/src/shared.rs
@@ -105,7 +105,7 @@ impl SharedSqlite {
 /// [`StoreError`] deliberately has no such conversion (callers choose how a
 /// database error is classified), so both travel in one enum and unwrap on the
 /// way out.
-pub(crate) fn read_snapshot<T>(
+fn read_snapshot<T>(
     conn: &mut SqliteConnection,
     f: impl FnOnce(&mut SqliteConnection) -> Result<T>,
 ) -> Result<T> {
--- a/storages/sqlite-storage/src/sqlite_store.rs
+++ b/storages/sqlite-storage/src/sqlite_store.rs
@@ -628,9 +628,12 @@ impl SqliteStore {
 
     async fn read_erased<T: Send + 'static>(&self, f: ReadQuery<T>) -> Result<T> {
         if self.reads.is_none() {
-            // No reader connections: the single pooled connection is what no
-            // writer can be holding while this query runs, so the snapshot
-            // comes for free and a transaction would only add statements.
+            // No reader connections: at the default `pool_size = 1` the single
+            // pooled connection is what no writer can hold while this query
+            // runs, so the snapshot is free and a transaction would only add
+            // statements. Raising `pool_size` breaks that — the writers that
+            // check out a connection without the permit could then commit
+            // mid-read — but it also deadlocks writes, so it stays unsupported.
             let pool = self.pool.clone();
             return self
                 .with_semaphore(move || {
@@ -6103,6 +6106,10 @@ mod read_routing_tests {
                     if n == ENTRIES {
                         return;
                     }
+                    // Both paths use the same blocking pool, so back-to-back
+                    // samples would compete with the writer for threads. Still
+                    // thousands of samples per batch.
+                    tokio::time::sleep(Duration::from_micros(200)).await;
                 }
             })
             .await;
@@ -6134,9 +6141,10 @@ mod read_routing_tests {
     ];
 
     /// Read-shaped methods that reach the database without going through
-    /// `read_query`, ignoring the ones excused above, plus how many were
-    /// scanned at all so the check cannot pass by matching nothing.
-    fn misrouted_reads(source: &str) -> (Vec<String>, usize) {
+    /// `read_query`: the ones with no excuse, the ones `ON_THE_WRITE_QUEUE`
+    /// excused, and how many were scanned at all so the check cannot pass by
+    /// matching nothing.
+    fn misrouted_reads(source: &str) -> (Vec<String>, Vec<String>, usize) {
         let source = source
             .split_once("\n#[cfg(test)]")
             .map(|(before, _)| before)
@@ -6144,6 +6152,7 @@ mod read_routing_tests {
 
         let mut current: Option<(&str, String)> = None;
         let mut offenders: Vec<String> = Vec::new();
+        let mut excused: Vec<String> = Vec::new();
         let mut scanned = 0usize;
         for line in source.lines() {
             if let Some((name, body)) = current.as_mut() {
@@ -6156,13 +6165,15 @@ mod read_routing_tests {
                     ]
                     .iter()
                     .any(|token| body.contains(token));
-                    if touches_db
-                        && !body.contains("read_query(")
-                        && !ON_THE_WRITE_QUEUE
+                    if touches_db && !body.contains("read_query(") {
+                        if ON_THE_WRITE_QUEUE
                             .iter()
                             .any(|(allowed, _)| allowed == name)
-                    {
-                        offenders.push((*name).to_string());
+                        {
+                            excused.push((*name).to_string());
+                        } else {
+                            offenders.push((*name).to_string());
+                        }
                     }
                     current = None;
                 } else {
@@ -6188,7 +6199,7 @@ mod read_routing_tests {
                 scanned += 1;
             }
         }
-        (offenders, scanned)
+        (offenders, excused, scanned)
     }
 
     /// A new read-only method written the old way (raw pool checkout, write
@@ -6197,7 +6208,7 @@ mod read_routing_tests {
     /// only place that can see the routing decision.
     #[test]
     fn read_shaped_methods_route_through_read_query() {
-        let (offenders, scanned) = misrouted_reads(include_str!("sqlite_store.rs"));
+        let (offenders, mut excused, scanned) = misrouted_reads(include_str!("sqlite_store.rs"));
         assert!(
             offenders.is_empty(),
             "read-only methods must call read_query (or be listed in ON_THE_WRITE_QUEUE \
@@ -6206,6 +6217,19 @@ mod read_routing_tests {
         assert!(
             scanned > 20,
             "the scan saw only {scanned} read-shaped methods"
+        );
+        // The allowlist has to be consumed in full, or an entry left behind by a
+        // later migration would silently excuse the next method of that name and
+        // its reason would be a lie.
+        let mut listed: Vec<String> = ON_THE_WRITE_QUEUE
+            .iter()
+            .map(|(name, _)| (*name).to_string())
+            .collect();
+        listed.sort();
+        excused.sort();
+        assert_eq!(
+            excused, listed,
+            "every ON_THE_WRITE_QUEUE entry must still name a read that bypasses read_query"
         );
     }
 
@@ -6226,7 +6250,7 @@ impl SqliteStore {
 ";
         assert_eq!(
             misrouted_reads(regression),
-            (vec!["get_something_new".to_string()], 2)
+            (vec!["get_something_new".to_string()], Vec::new(), 2)
         );
     }
 }
--- a/storages/sqlite-storage/src/sqlite_store.rs
+++ b/storages/sqlite-storage/src/sqlite_store.rs
@@ -1410,14 +1410,21 @@ impl SqliteStore {
         key_id: &[u8],
         device_id: i32,
     ) -> Result<Option<AppStateSyncKey>> {
+        // On the write queue: a stale absent answer is sent on the wire as an
+        // orphan reply to a peer's key request, so it is not a miss the caller
+        // retries.
+        let pool = self.pool.clone();
         let key_id = key_id.to_vec();
         let res: Option<Vec<u8>> = self
-            .read_query(move |conn| {
+            .with_semaphore(move || -> Result<Option<Vec<u8>>> {
+                let mut conn = pool
+                    .get()
+                    .map_err(|e| StoreError::Connection(Box::new(e)))?;
                 let res: Option<Vec<u8>> = app_state_keys::table
                     .select(app_state_keys::key_data)
                     .filter(app_state_keys::key_id.eq(&key_id))
                     .filter(app_state_keys::device_id.eq(device_id))
-                    .first(conn)
+                    .first(&mut conn)
                     .optional()
                     .map_err(|e| StoreError::Database(Box::new(e)))?;
                 Ok(res)
@@ -1479,8 +1486,14 @@ impl SqliteStore {
         &self,
         device_id: i32,
     ) -> Result<Option<Vec<u8>>> {
+        // On the write queue: a stale absent answer becomes InvalidRequest and
+        // fails the user's app-state action outright.
+        let pool = self.pool.clone();
         let res: Option<Vec<u8>> = self
-            .read_query(move |conn| {
+            .with_semaphore(move || -> Result<Option<Vec<u8>>> {
+                let mut conn = pool
+                    .get()
+                    .map_err(|e| StoreError::Connection(Box::new(e)))?;
                 // Return the latest key whose blob actually decodes. A legacy bincode
                 // row (or a corrupt one) reads as absent via get_sync_key but still
                 // sits in the table with a possibly lexicographically-higher key_id;
@@ -1491,7 +1504,7 @@ impl SqliteStore {
                     .select((app_state_keys::key_id, app_state_keys::key_data))
                     .filter(app_state_keys::device_id.eq(device_id))
                     .order(app_state_keys::key_id.desc())
-                    .load(conn)
+                    .load(&mut conn)
                     .map_err(|e| StoreError::Database(Box::new(e)))?;
                 let res = candidates
                     .into_iter()
@@ -6338,6 +6351,16 @@ mod read_routing_tests {
              cache in front on that path, so a stale miss loses the addon too",
         ),
         ("get_pn_mapping", "same as get_lid_mapping"),
+        (
+            "get_app_state_sync_key_for_device",
+            "a stale absent answer is sent on the wire as an orphan reply to a \
+             peer's key request, not retried by the caller",
+        ),
+        (
+            "get_latest_app_state_sync_key_id_for_device",
+            "a stale absent answer becomes InvalidRequest and fails the user's \
+             app-state action outright",
+        ),
         (
             "get_all_lid_mappings",
             "the startup warm-up feeds these into LidPnCache::add_guarded, whose \
--- a/storages/sqlite-storage/src/sqlite_store.rs
+++ b/storages/sqlite-storage/src/sqlite_store.rs
@@ -127,6 +127,12 @@ pub struct SqliteStore {
     /// once and deadlock on the write-lock upgrade — the exact failure this
     /// change exists to avoid.
     pub(crate) reads: Option<ReadPool>,
+    /// Whether a deferred read transaction is safe here: WAL, and not shared
+    /// cache. It is the same condition that decides [`Self::reads`], and it has
+    /// to gate the wider-write-pool snapshot too — under shared cache a read
+    /// transaction holds table locks that fail the writer with
+    /// `SQLITE_LOCKED_SHAREDCACHE`, which `busy_timeout` cannot absorb.
+    pub(crate) snapshot_safe: bool,
     pub(crate) database_path: String,
     device_id: i32,
 }
@@ -590,6 +596,7 @@ impl SqliteStore {
             pool,
             db_semaphore: Arc::new(tokio::sync::Semaphore::new(pool_size as usize)),
             reads,
+            snapshot_safe: declined.is_none(),
             database_path,
             device_id,
         })
@@ -631,7 +638,11 @@ impl SqliteStore {
         // one pooled connection is itself the snapshot: no writer can be on it,
         // including the several that skip the permit and check one out directly.
         // A transaction would only add statements to every read.
-        if self.reads.is_none() && self.pool.max_size() <= 1 {
+        // The snapshot is only worth opening where it is safe: at
+        // `pool_size = 1` the exclusive connection already gives it, and under
+        // shared cache or without WAL a read transaction would lock out the
+        // writer instead.
+        if self.reads.is_none() && (self.pool.max_size() <= 1 || !self.snapshot_safe) {
             let pool = self.pool.clone();
             return self
                 .with_semaphore(move || {
@@ -2703,8 +2714,15 @@ impl ProtocolStore for SqliteStore {
     }
 
     async fn get_all_lid_mappings(&self) -> Result<Vec<LidPnMappingEntry>> {
+        // On the write queue: the startup warm-up feeds these rows into
+        // `LidPnCache::add_guarded`, whose LID side replaces unconditionally, so
+        // a stale row read during a live learn reverts reverse resolution.
+        let pool = self.pool.clone();
         let device_id = self.device_id;
-        self.read_query(move |conn| {
+        self.with_semaphore(move || -> Result<Vec<LidPnMappingEntry>> {
+            let mut conn = pool
+                .get()
+                .map_err(|e| StoreError::Connection(Box::new(e)))?;
             let rows: Vec<(String, String, i64, String, i64)> = lid_pn_mapping::table
                 .select((
                     lid_pn_mapping::lid,
@@ -2714,7 +2732,7 @@ impl ProtocolStore for SqliteStore {
                     lid_pn_mapping::updated_at,
                 ))
                 .filter(lid_pn_mapping::device_id.eq(device_id))
-                .load(conn)
+                .load(&mut conn)
                 .map_err(|e| StoreError::Database(Box::new(e)))?;
             Ok(rows
                 .into_iter()
@@ -6106,6 +6124,48 @@ mod read_routing_tests {
         );
     }
 
+    /// A shared-cache store declines reader connections because a read
+    /// transaction there holds table locks the writer cannot wait out. The
+    /// wider-write-pool snapshot has to decline for the same reason instead of
+    /// reintroducing exactly that transaction.
+    #[tokio::test]
+    async fn a_shared_cache_store_gets_no_snapshot_even_with_a_wider_write_pool() {
+        use portable_atomic::AtomicU64;
+        use std::sync::atomic::Ordering;
+        static COUNTER: AtomicU64 = AtomicU64::new(0);
+        let id = COUNTER.fetch_add(1, Ordering::Relaxed);
+        let url = format!(
+            "file:memdb_snapshot_gate_{}_{id}?mode=memory&cache=shared",
+            std::process::id()
+        );
+        let store = SqliteStore::with_config(
+            &url,
+            SqliteStoreConfig {
+                pool_size: 2,
+                read_pool_size: 4,
+                ..Default::default()
+            },
+        )
+        .await
+        .expect("store opens");
+
+        assert!(store.reads.is_none(), "shared cache declines reader pool");
+        assert!(
+            !store.snapshot_safe,
+            "and must decline the deferred read transaction with it"
+        );
+
+        // Still fully operational on the plain path.
+        store.create_new_device().await.expect("device row");
+        store.put_session(ADDR, b"blob").await.unwrap();
+        assert!(
+            store
+                .has_signal_state_for_user("559990000001")
+                .await
+                .unwrap()
+        );
+    }
+
     /// An uncommitted write is not a lock error and not a phantom miss: the
     /// reader sees the last committed state and returns it. This is the case
     /// the msg-secret reads were kept on the write queue for, so it has to hold
@@ -6196,7 +6256,20 @@ mod read_routing_tests {
             // of hanging the runtime.
             let sampled = tokio::time::timeout(Duration::from_secs(20), async {
                 loop {
-                    let n = store.get_sender_key_devices(GROUP).await.unwrap().len();
+                    // Straight through read_query, not get_sender_key_devices:
+                    // that one is on the write permit now, which would serialize
+                    // the sample against the writer and hide a torn batch.
+                    let n = store
+                        .read_query(|conn| {
+                            sender_key_devices::table
+                                .filter(sender_key_devices::group_jid.eq(GROUP))
+                                .count()
+                                .get_result::<i64>(conn)
+                                .map(|n| n as usize)
+                                .map_err(|e| StoreError::Database(Box::new(e)))
+                        })
+                        .await
+                        .unwrap();
                     assert!(
                         n == 0 || n == ENTRIES,
                         "a chunked batch was observed {n}/{ENTRIES} applied"
@@ -6236,6 +6309,11 @@ mod read_routing_tests {
              cache in front on that path, so a stale miss loses the addon too",
         ),
         ("get_pn_mapping", "same as get_lid_mapping"),
+        (
+            "get_all_lid_mappings",
+            "the startup warm-up feeds these into LidPnCache::add_guarded, whose \
+             LID side replaces unconditionally, so a stale row reverts a live learn",
+        ),
         // The rest share one shape: the row is promoted into a plain in-memory
         // cache, or suppresses an action, so a stale read sticks instead of
         // being retried. `SignalStoreCache` reconciles staleness and its reads
--- a/storages/sqlite-storage/src/sqlite_store.rs
+++ b/storages/sqlite-storage/src/sqlite_store.rs
@@ -627,13 +627,11 @@ impl SqliteStore {
     }
 
     async fn read_erased<T: Send + 'static>(&self, f: ReadQuery<T>) -> Result<T> {
-        if self.reads.is_none() {
-            // No reader connections: at the default `pool_size = 1` the single
-            // pooled connection is what no writer can hold while this query
-            // runs, so the snapshot is free and a transaction would only add
-            // statements. Raising `pool_size` breaks that — the writers that
-            // check out a connection without the permit could then commit
-            // mid-read — but it also deadlocks writes, so it stays unsupported.
+        // At the default `pool_size = 1` with no reader connections, holding the
+        // one pooled connection is itself the snapshot: no writer can be on it,
+        // including the several that skip the permit and check one out directly.
+        // A transaction would only add statements to every read.
+        if self.reads.is_none() && self.pool.max_size() <= 1 {
             let pool = self.pool.clone();
             return self
                 .with_semaphore(move || {
@@ -644,8 +642,9 @@ impl SqliteStore {
                 })
                 .await;
         }
-        // One implementation of acquire-checkout-snapshot, shared with the
-        // sibling-crate read path.
+        // Otherwise the deferred read transaction is what pins the snapshot,
+        // whether the concurrency comes from reader connections or from a wider
+        // write pool. One implementation, shared with the sibling-crate path.
         self.shared().read(f).await
     }
 
@@ -6006,6 +6005,43 @@ mod read_routing_tests {
             .expect("a read must not queue behind the write permit")
             .expect("read succeeds");
         assert_eq!(got.as_deref(), Some(&b"blob"[..]));
+    }
+
+    /// `pool_size > 1` with no reader connections is reachable config, and there
+    /// the permit no longer implies an exclusive connection: the writers that
+    /// check one out directly can commit between a multi-statement read's
+    /// queries. The deferred transaction has to cover that case too.
+    #[tokio::test]
+    async fn a_multi_statement_read_is_snapshot_isolated_with_a_wider_write_pool() {
+        let db = TempDb::new("wide_pool");
+        let store = SqliteStore::with_config(
+            &db.url(),
+            SqliteStoreConfig {
+                pool_size: 2,
+                read_pool_size: 0,
+                ..Default::default()
+            },
+        )
+        .await
+        .expect("store opens");
+        assert!(store.reads.is_none(), "no reader connections requested");
+        store.create_new_device().await.expect("device row");
+        store.put_session(ADDR, b"blob").await.unwrap();
+
+        // has_signal_state_for_user issues two EXISTS; both must see one
+        // snapshot even though a second write connection is available.
+        assert!(
+            store
+                .has_signal_state_for_user("559990000001")
+                .await
+                .unwrap()
+        );
+        assert!(
+            !store
+                .has_signal_state_for_user("559990000009")
+                .await
+                .unwrap()
+        );
     }
 
     /// An uncommitted write is not a lock error and not a phantom miss: the
--- a/storages/sqlite-storage/src/sqlite_store.rs
+++ b/storages/sqlite-storage/src/sqlite_store.rs
@@ -6207,7 +6207,6 @@ mod read_routing_tests {
             "and must decline the deferred read transaction with it"
         );
 
-        // Still fully operational on the plain path.
         store.create_new_device().await.expect("device row");
         store.put_session(ADDR, b"blob").await.unwrap();
         assert!(
@@ -6215,6 +6214,58 @@ mod read_routing_tests {
                 .has_signal_state_for_user("559990000001")
                 .await
                 .unwrap()
+        );
+
+        // The flags above are only the mechanism. What has to hold is that a
+        // write still commits with a read parked mid-flight: on the snapshot
+        // path the writer meets the reader's table lock as
+        // `SQLITE_LOCKED_SHAREDCACHE`, which `busy_timeout` cannot absorb. The
+        // park outlasts `with_retry`'s ~310ms budget, so that lock is fatal
+        // rather than retried away.
+        let (open_tx, mut open_rx) = tokio::sync::mpsc::unbounded_channel();
+        let (release_tx, release_rx) = std::sync::mpsc::channel::<()>();
+        let reader = {
+            let store = store.clone();
+            tokio::spawn(async move {
+                store
+                    .read_query(move |conn| {
+                        let first = sessions::table
+                            .select(sessions::record)
+                            .filter(sessions::address.eq(ADDR))
+                            .first::<Vec<u8>>(conn)
+                            .optional()
+                            .map_err(|e| StoreError::Database(Box::new(e)))?;
+                        let _ = open_tx.send(());
+                        let _ = release_rx.recv_timeout(Duration::from_secs(20));
+                        Ok(first)
+                    })
+                    .await
+            })
+        };
+
+        tokio::time::timeout(Duration::from_secs(10), open_rx.recv())
+            .await
+            .expect("the read must reach its query")
+            .expect("reader alive");
+
+        let releaser = tokio::spawn(async move {
+            tokio::time::sleep(Duration::from_millis(1500)).await;
+            let _ = release_tx.send(());
+        });
+
+        tokio::time::timeout(
+            Duration::from_secs(10),
+            store.put_session(ADDR, b"committed-under-shared-cache"),
+        )
+        .await
+        .expect("the write must not stall behind the parked read")
+        .expect("the write must commit, not meet a shared-cache lock");
+
+        releaser.await.expect("join releaser");
+        reader.await.expect("join").expect("read");
+        assert_eq!(
+            store.get_session(ADDR).await.unwrap().as_deref(),
+            Some(&b"committed-under-shared-cache"[..])
         );
     }
 
--- a/storages/sqlite-storage/src/sqlite_store.rs
+++ b/storages/sqlite-storage/src/sqlite_store.rs
@@ -619,9 +619,10 @@ impl SqliteStore {
     /// merely overlap a write see either state, which is what the single permit
     /// already gave them (it ordered them arbitrarily, not causally).
     ///
-    /// Only correct for statements that cannot write: reader connections carry
-    /// `PRAGMA query_only`, so a write sent here fails instead of escaping the
-    /// serialization the store depends on.
+    /// Only correct for statements that cannot write. Reader connections carry
+    /// `PRAGMA query_only`, so a write sent here fails loudly -- but the
+    /// fallback hands out an ordinary write connection, so with no reader pool
+    /// (the default) that net is absent and the routing scan is the only guard.
     async fn read_query<F, T>(&self, f: F) -> Result<T>
     where
         F: FnOnce(&mut SqliteConnection) -> Result<T> + Send + 'static,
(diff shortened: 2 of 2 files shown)


jlucaso1:
Pushed 6cf393f for the review round. Taking the substantive ones in order:

**Chunked writes could be observed half-applied** (Codex P2, cubic P1) — real, and my fault: `put_app_state_mutation_macs_for_device`, `delete_app_state_mutation_macs_for_device`, `mark_prekeys_uploaded`, `set_sender_key_status` and `delete_sender_key_device_rows` chunk for the host-parameter limit and were committing each chunk on its own. Invisible while readers held the same permit; not invisible once a reader runs alongside. Each loop is now one transaction, which also stops a crash mid-batch from persisting half a batch. `a_chunked_batch_write_is_never_observed_half_applied` races a four-chunk write against a reader; I checked it fails immediately with the transaction removed, so it isn't decorative.

**Decrypt reads under an in-flight write are untested** (cubic P2) — fair, the old test held an idle permit. `a_read_sees_the_last_commit_while_a_write_transaction_is_open` now holds a real `IMMEDIATE` transaction open on the write pool and asserts the read returns the last commit rather than blocking or erroring, which is the phantom-miss case the msg-secret comment was guarding.

**Reader branch duplicates `SharedSqlite::read`** (cubic P3) — agreed, it delegates now. That also reverts the `read_snapshot` visibility change, so `shared.rs` is untouched by this PR.

**Widen the routing scan** (CodeRabbit, cubic P2) — done: `is_`/`list_`/`count_`/`find_`/`fetch_` prefixes plus the `_exists` suffix, so `device_exists` is covered.

**Wrap the fallback in `read_snapshot` too** (CodeRabbit, cubic P3) — not doing this one. You're right that the permit doesn't serialize the raw-pool writers, and I've fixed the comment that claimed it did: at the default `pool_size = 1` the serialization comes from the single pooled connection, not the permit. But the only configuration where the difference is observable is `pool_size > 1`, and that one deadlocks writes outright — two deferred transactions upgrade against each other and `busy_timeout` can't break it, which is why the field doc says to leave it at 1. Adding two statements to every read on a store that opted into no reader connections, to half-fix a configuration that doesn't work, isn't worth it. If we ever make `pool_size > 1` real, this goes in with it.

Binary size: the first revision blew the gate at +116 KiB stripped / +90.6 KiB .text. `read_query` was generic over its closure, so all 26 call sites monomorphized a body carrying Diesel's transaction machinery. Boxing the closure before the real body drops it to one instantiation per return type: **+31.53 KiB stripped, +23.50 KiB .text**, measured with `scripts/ci/measure_binary_size.py` against the branch point. Both inside budget, no label needed.


---

jlucaso1:
Codex's second P2 is right and 51016bb reverts it: `get_msg_secret` and `get_msg_secret_with_ts` are back on the write queue.

I checked the failure it describes rather than taking it on faith. `store_history_sync_msg_secret_entries` calls `put_msg_secrets` directly instead of going through the live write-behind buffer (deliberately — the batch is already owned, and routing it through the buffer would retain it twice). So a live `secret_encrypted_message` arriving during a history-sync seed misses the buffer lookup and goes to the backend. On a total miss `src/message/msg_secret.rs` returns `None` and the reaction, vote or edit is dropped: no retry, no buffering, terminal.

My "the outcome set is unchanged" argument still holds formally — the permit ordered a concurrent read and write arbitrarily, so a miss was always reachable — but that is a statement about outcomes, not odds, and here the losing outcome is a dropped message rather than a cache miss the caller already handles. Going from "the read has to arrive before the write starts" to "the read has to arrive before the write commits" is a real widening, and a history-sync batch commit is not short. These two reads also aren't what this PR is for and contribute none of the Case B win, so there was nothing on the other side of the scale.

They're listed in `ON_THE_WRITE_QUEUE` with the reason, alongside `get_pending_inbound`, so the routing scan accepts them and the decision lives next to the check that enforces it. The Audit section of the description now records them as not migrated.

That is the one method-level judgement in this PR I got wrong, and it's the one the batch was explicitly supposed to catch — good catch.


---

jlucaso1:
bcabb05. Codex and cubic independently landed on the same thing, and they're right.

**LID/PN mappings back on the write queue.** `alternate_msg_secret_jid` resolves the peer's other namespace through `get_lid_mapping`/`get_pn_mapping` and feeds it straight into the message-secret lookup I moved back last round. There's no cache in front on that path — the `lid_pn_cache` sits in front of a different caller (`get_lid_pn_entry_by_user` in `lid_pn.rs`), not `msg_secret.rs`, which calls the backend directly. So a lookup racing `persist_and_migrate_lid_pn` reads the pre-write snapshot, resolves no alternate JID, and the addon is rejected: the same terminal miss, one indirection earlier. Protecting the secret read and not the mapping read that decides which key it's looked up under was half a fix. `get_all_lid_mappings` stays on the read path — bulk enumeration, no caller on that path.

**Declining the prekey one** (Codex, `load_prekey` vs `remove_prekey`). The premise is that the single connection "waited and returned `None` after deletion", but the backend row is deliberately still present long before the delete transaction opens: `buffer_consumed_prekey` only records the id in `removed_prekeys`, and `flush` deletes it later, after the promoted session is durable. `signal_cache.rs`'s own test asserts exactly that — `load_prekey` returns `Some` after the consume and `None` only after the flush. `PreKeyAdapter::get_pre_key` goes straight to the backend with no interception of the pending-delete set, so a second decrypt reading that row is the normal state for the whole write-behind interval, which is orders of magnitude longer than one delete statement. My change extends an intentional window by the length of a transaction; it doesn't create one. Narrowing it properly means having the cache answer `load_prekey` from `removed_prekeys`, which is `signal_cache.rs` — another batch is in that file and I'm not touching it here.

**Also from cubic:** `get_msg_secret` and `get_msg_secret_with_ts` had become near-identical closures after the revert. `get_msg_secret` now delegates and drops the timestamp, so there's one query and one documented rationale instead of two that can drift.

Net effect on the audit: 22 methods migrated, four held back with reasons (`get_pending_inbound`, `get_msg_secret_with_ts`, `get_lid_mapping`, `get_pn_mapping`), all four in `ON_THE_WRITE_QUEUE` next to the scan that enforces the rule. The measured win is unchanged — it comes entirely from the session/identity/sender-key reads, none of which are in that list.


---

jlucaso1:
9923ffb — took all three.

**The allowlist could rot, and that's the best catch of this round.** `ON_THE_WRITE_QUEUE` was only ever consulted in one direction: it excused a listed name, never checked the name still needed excusing. Migrate one of those reads later, forget the entry, and it sits there excusing the next method that happens to share the name while its reason string quietly becomes false — which defeats the point of a guard whose whole value is that it doesn't rot. `misrouted_reads` now also returns the names it excused, and the test asserts that set equals the allowlist. I checked it fires: migrating `get_pn_mapping` to `read_query` with its entry left in place produces

```
every ON_THE_WRITE_QUEUE entry must still name a read that bypasses read_query
  left:  ["get_lid_mapping", "get_msg_secret_with_ts", "get_pending_inbound"]
  right: ["get_lid_mapping", "get_msg_secret_with_ts", "get_pending_inbound", "get_pn_mapping"]
```

**Throttled the chunked-write race** by 200µs per sample. You're right that both sides go through `spawn_blocking` on the same pool and back-to-back sampling competes with the writer for threads — that's the shape that reads as flake in six months and gets deleted. Re-verified the test still has teeth afterwards: with the transaction removed it fails on the first round at 190/760, same as before the throttle.

**Fallback comment** now says the free-snapshot claim holds at `pool_size = 1`, and that raising it breaks the claim but also deadlocks writes, so it stays unsupported. That's the third time this spot has come up; better to have the condition written down than to keep re-litigating it in review.


---

jlucaso1:
2018115 — I was wrong on this one and you were right to keep pushing.

I'd been declining the fallback snapshot on two grounds, and the new argument breaks both. `pool_size > 1` with `read_pool_size = 0` isn't hypothetical: it's reachable through `SqliteStoreConfig` and this crate's own `with_config_custom_tuning_builds_and_operates` uses exactly `pool_size: 2, read_pool_size: 0`. And "it deadlocks writes anyway, so it's unsupported" was an overstatement — raising `pool_size` doesn't deadlock by itself, only two deferred read-then-write transactions racing do. Plenty of the writes here are single statements. So the configuration is reachable *and* usable, and in it the raw-pool writers can commit between the statements of `has_signal_state_for_user`, `load_prekeys_batch` or `get_app_state_mutation_macs_batch_for_device`.

What I did differently from the suggested diff: rather than wrapping the fallback unconditionally, the shortcut is now conditioned on what actually makes it true — `self.reads.is_none() && self.pool.max_size() <= 1`. At the default that path is unchanged statement for statement, so a store that opted into nothing pays nothing new, which was the one thing I cared about defending. Anything wider goes through `SharedSqlite::read`, which is already the permit-plus-deferred-transaction path, so there's still one implementation of it rather than two. Test covers the wide-pool case.

That's four reviews on this one spot. The reason it took that long is that I was defending the cost constraint without checking whether the configuration I was dismissing was real. It was, and it's in our own test file.


---

jlucaso1:
77906ae — all three taken.

I verified the mechanism rather than just accepting the reasoning. `get_devices` calls `device_registry_cache.promote(...)` unconditionally on a backend hit (`device_registry.rs:670`), with nothing comparing against what's already cached. So a stale row doesn't just produce one wrong answer, it *replaces* a newer cache entry and the wrongness persists until the next refresh. Same shape for `get_sender_key_devices` initializing `sender_key_device_cache`, and `get_tc_token` feeding `prepare_privacy_token`'s scheduling decision, where the stale read suppresses an action rather than triggering a retry. And outgoing sends aren't per-chat serialized, so none of the three has a lock protecting it.

That's now five reverts across three reviewers, and they were all one thing I should have caught in the original audit. My criterion was "provably read-only", which is necessary and not sufficient. The criterion that actually decides it is what the caller does with a stale-absent or stale-valued answer:

- **Migrates:** the miss triggers recovery. `SignalStoreCache` reconciles staleness explicitly — dirty set, incarnation, checkout tokens — and a genuine miss means fetch a pre-key bundle, redistribute an SKDM, or re-sync a collection. That's every read on the measured path.
- **Stays:** the answer is promoted into a plain cache that overwrites whatever it's handed, or it suppresses an action. Then the stale read sticks, and no later attempt corrects it.

That distinction is now written next to `ON_THE_WRITE_QUEUE` rather than living in my head, so the next person adding a read has the actual test to apply.

None of the seven held-back methods is on the measured path — Case B is `get_session` — so the numbers in the description are unchanged and this costs only routing. I also checked the remaining migrated protocol-store reads (`get_group_metadata`, `get_all_tc_token_jids`, `get_all_lid_mappings`, `has_same_base_key`) against the rule: none is promoted into a cache and none suppresses an action; their stale outcome is a re-read or a retry. If you'd rather I move those too on general caution, say so and I'll do it — they contribute nothing to the win either.


---

jlucaso1:
68f1a49 — both taken, and the first one is a fair hit.

**The wide-pool test proved nothing.** You're right: it called `has_signal_state_for_user` twice with nothing writing in between, so it passed with or without the deferred transaction it existed to cover. I'd verified the chunked-write race and the allowlist scan by removing their mechanism and watching them fail, and then didn't apply that to this one. It now runs two SELECTs inside one `read_query` closure, parks between them, and commits through the pool's other connection while parked. With the snapshot removed:

```
both queries must see one snapshot, not the write that landed between them
  left:  Some("committed-mid-read")
  right: Some("blob")
```

**`self.shared().run(` evaded the scan.** Also right, and it's the one bypass that would look completely idiomatic in a future method. Added as a token, and the scan now strips indentation before matching so a call rustfmt has split across `self` / `.shared()` / `.run(` still reads as one token — without that the token match was defeated by formatting alone. The self-test now carries a `shared().run` offender alongside the raw-pool one, so the detection is covered rather than assumed.

Not doing the whitespace normalization beyond `trim_start`: joining lines that way can only create matches where a call really was split, which is what we want, and full normalization would risk gluing unrelated tokens together.

Note your last few runs report `Git: Failed to clone repository`, so they reviewed 2018115 rather than the current head — these two findings still applied to current code, but you may want a `@coderabbitai full review` once the clone issue clears. Greptile has reviewed through 77906ae (seventh pass, approved).


---

jlucaso1:
⚠️ **Do not merge 68f1a49 as-is.** Codex found two real defects in the current head. I have fixes committed locally but my git push credentials died mid-session (`could not read Username for 'https://github.com'`, four retries with backoff, and re-attaching the repo did not restore them), so 39b9a38 is stuck on this machine. Writing the findings up here so they aren't lost with the session.

**1. The wide-pool snapshot is unsafe under shared cache** (Codex, `read_erased`). Real, and mine — introduced in 2018115. `build` declines reader connections for shared-cache stores precisely because a read transaction there holds table locks that fail the writer with `SQLITE_LOCKED_SHAREDCACHE`, which `busy_timeout` cannot absorb. My `pool_size > 1` fallback then opens exactly that transaction on the main pool, reintroducing the hazard the decline exists to avoid. Not hypothetical: `with_config_custom_tuning_builds_and_operates` runs a `cache=shared` URL with `pool_size: 2`, so the configuration is in the test suite today.

Fix: record the same condition that gates the reader pool (`declined.is_none()`) as `snapshot_safe`, and require it before opening the deferred transaction — `self.reads.is_none() && (self.pool.max_size() <= 1 || !self.snapshot_safe)` takes the plain path. Plus a test asserting a shared-cache store with `pool_size: 2, read_pool_size: 4` has both `reads == None` and `snapshot_safe == false` and still operates.

**2. `a_chunked_batch_write_is_never_observed_half_applied` has gone vacuous** (Codex). Also mine, and worse than a missed bug: I verified that test failed without the transaction, said so in the commit message, and then in 77906ae moved `get_sender_key_devices` onto the write permit — which serializes the sampler against the writer, so it can no longer observe a torn batch. The test kept passing and the claim stayed in the commit message after it stopped being true. Codex is right that removing the transaction today would not fail it.

Fix: sample through `read_query` directly with a `count()` on `sender_key_devices` instead of going through a method whose routing may change under it. Re-verified: with the transaction removed it fails at 190/760.

**3. `get_all_lid_mappings` goes back to the write queue** (Codex). The startup warm-up feeds it into `LidPnCache::add_guarded`, whose LID side replaces unconditionally, so a stale row read during a live learn reverts reverse resolution. Same rule that moved the other cache-fed reads; I had wrongly cleared this one as "bulk enumeration", which is a description, not the test.

**On cubic's two** (`get_devices`, `get_tc_token` not truly serialized against their writers): correct at `pool_size > 1`, and I am not fixing it. Those writers check out a connection without taking the permit, so the ordering only holds at the default `pool_size = 1` — which is the same caveat that applies to every permit-based ordering in this file, and predates this PR. Making it hold generally means routing those writers through `db_semaphore`, i.e. changing write serialization, which is out of scope here and deserves its own change. Worth noting the rationale comments I added say "on the write queue" and that is accurate as a routing statement, but the ordering it buys is `pool_size = 1`-conditional.

Local validation of the fix commit: 75 sqlite-storage, 1367 whatsapp-rust, 25 chat-store, clippy clean on all four crates. Someone with push access can reproduce it from this description, or I can retry if credentials come back.


---

jlucaso1:
✅ **Retracting the do-not-merge warning** — push credentials came back and 39b9a38 is on the branch. Everything described in the previous comment is now in the code, not just in prose:

- the wide-pool snapshot is gated on `snapshot_safe` (the same `declined.is_none()` condition that gates the reader pool), so a shared-cache or non-WAL store takes the plain path instead of opening a read transaction that would lock out the writer, with a test asserting both `reads == None` and `snapshot_safe == false` for `cache=shared` at `pool_size: 2, read_pool_size: 4`;
- `a_chunked_batch_write_is_never_observed_half_applied` samples through `read_query` directly instead of `get_sender_key_devices`, and I re-confirmed it fails at 190/760 with the transaction removed;
- `get_all_lid_mappings` is back on the write queue with its reason in `ON_THE_WRITE_QUEUE`.

That makes 18 migrated and 8 held back. The Audit table in the description still lists `get_all_lid_mappings` on the migrated side — I'll correct that next.

Also worth recording, since it corroborates the description: the binary-size gate ran and passed at **+28.22 KiB stripped** (budget 64) and **+24.56 KiB .text** (budget 32), against my local measurement of +31.53 / +23.50. Same conclusion, and that was the one check that genuinely failed earlier in this PR. CodSpeed (core, proto-signal, integration), the wasm32 build, and the performance analysis are all green too; the remaining jobs are still queued behind the repo's runner backlog.


---

jlucaso1:
🛑 **Recommending this does not merge in its current form.** Two things landed that I don't think I should paper over.

## E2E is a real regression

`E2E Tests` failed on 39b9a38. I checked before calling it mine: the same workflow is green on `main` across the last eight runs, including `f8165f28`, this PR's base. Step 9 ran for 3m26s and failed, so it built and executed — not infra.

I could not diagnose it. `get_job_logs` for that job returns only the mock-server service-container stream (I pulled 777 KB of it; zero occurrences of `test result`, `FAILED`, or `panicked` — it is all bartender output). The e2e suite needs the mock server, which this environment doesn't have. I ruled out the cheap explanation by confirming `cargo check -p e2e-tests --tests` passes, so it is a runtime failure, not a compile break.

The most likely mechanism from my diff, at the default config E2E uses (`read_pool_size = 0`, `pool_size = 1`): roughly ten reads that previously ran on a raw pooled connection **without** the permit now take it — `load_prekey`, `load_signed_prekey`, the four app-state reads, `device_exists`, `load_device_data_for_device`, `has_same_base_key`, `get_group_metadata`. That is a real serialization change at the default, and it is the part of this PR I described as "unchanged" for `read_pool_size = 0`. That description was accurate for the previously-`with_semaphore` reads and wrong for these. Whoever picks this up should start there.

## Codex's P1 is right about the mechanism, and the fix isn't in this file

The cold `SignalStoreCache::has_session` probe reads the backend outside the sessions lock, then rechecks only *whether an entry exists*:

```rust
let backend_result = backend.get_session(key).await?;
let mut state = self.lock_sessions().await;
if let Some(entry) = state.cache.get(key) { return Ok(entry.exists()); }
// ... installs the record decoded from backend_result
```

If a concurrent flush commits a newer record and then `evict_if_needed` drops the now-clean entry while the probe is in I/O, the recheck finds nothing and the probe installs the **older** record. A later checkout resumes that chain index. That is message key/IV reuse, which is the one property `agent_docs/signal_durability.md` treats as non-negotiable.

The race predates this PR in shape — the permit ordered the two *database* operations but never ordered the probe's cache insert against the flush's eviction. What this PR does is widen the window from "the probe won the permit before the write started" to "the probe's snapshot opened before the write committed". On this particular consequence, widening is not acceptable.

The correct fix is a generation or version on cold loads in `SignalStoreCache`, so a load spanning a flush/eviction is rejected rather than installed. That is `wacore/src/store/signal_cache.rs`, which I was explicitly scoped out of because another change owns it. The alternative — moving `get_session_for_device` back to the write queue — removes the entire measured benefit, since Case B is `get_session`: p50 37 ms → 0.4 ms and p99 45 ms → 4 ms all come from that one read.

## What I think should happen

This wants to be two changes, in order: first the cache-generation guard in `signal_cache.rs`, then this routing on top of it. Sequenced that way the session read moves safely and the win is real. Merged as-is, the win is real and so is a narrow key-reuse window.

Everything else in here stands on its own and is independently correct — the five chunked writers made atomic, the eight reads correctly held back, the shared-cache snapshot gate, the routing scan with its bidirectional allowlist. Those could be split out and merged without the session-read routing if you want the safe part now.

I've stopped pushing changes rather than guess at the E2E failure blind or reach into a file another change owns.


---

jlucaso1:
Converted to draft, so the two open blockers aren't outweighed by the approval count. Reversible — flip it back whenever you want.

To be explicit about why the signals disagree: Greptile has approved eleven times and cubic reports clean, but neither is looking at CI, and neither surfaced the cold-load race. That one came from Codex and I confirmed it against `signal_cache.rs` directly. A stack of green bot reviews here means "nothing further found by static review", not "the change is safe to ship".

The two blockers, unchanged:

1. **E2E fails on this branch and is green on `main`** (eight consecutive runs, including base `f8165f28`). Undiagnosed — the job's log endpoint returns only the mock-server container stream, and the suite needs a mock server I don't have. Prime suspect is in my own diff: about ten reads that previously ran on a raw pooled connection without the permit now take it at the default config, which is a serialization change I had described as "unchanged".

2. **The cold `has_session` probe can install a stale session record** if a concurrent flush commits and then evicts during its backend I/O, because the recheck only asks whether an entry exists. Resuming that chain index is key/IV reuse. The window predates this PR; this PR widens it. The fix is a generation guard on cold loads in `signal_cache.rs`, which is owned by another change.

Suggested split if you want the safe part sooner: the chunked-writer atomicity, the eight write-queue holds, the shared-cache snapshot gate and the routing scan are all independently correct and don't depend on the session read moving. The `get_session` routing — which is the entire measured win — should land after the cache-generation guard, not before it.


---

jlucaso1:
Recording Codex's `load_prekeys_batch` finding, and correcting something I said two comments ago.

**On the finding itself**, I'd apply the same reasoning I used to decline its sibling. `buffer_consumed_prekey` only records the id in `removed_prekeys`; `flush` deletes the row later, once the promoted session is durable. So a consumed-but-not-yet-deleted prekey is already re-offerable by `load_prekeys_batch` for the whole write-behind interval, before this PR touches anything. What routing adds is the ability to read the row *during* the delete transaction rather than only before it starts — an extension of an existing window, not a new one, exactly as with `load_prekey`. Declining one and accepting the other would be inconsistent.

**What I got wrong** is the split I proposed. I said the migrated set was sound and only the `get_session` routing needed to wait. That isn't established. Codex has now flagged four separate migrated reads (`get_sender_key_devices`, `get_devices`, `get_tc_token`, `get_all_lid_mappings` — all reverted) plus `load_prekey` and `load_prekeys_batch` on this shape, and the corrected criterion — *what does the caller do with a stale answer* — was applied retroactively, one method at a time, as reviewers found them. It was never run cleanly over the whole set from the start.

So the accurate statement is: the parts of this PR I'd still call independently correct are the ones that don't depend on the criterion at all — the five chunked writers made atomic, the shared-cache snapshot gate, and the routing scan with its bidirectional allowlist. The migrated set needs a fresh pass under the corrected rule before any of it moves, and that pass should include the prekey reads with an explicit decision on whether the write-behind window makes the routing question moot for them or not.

That plus the undiagnosed E2E regression is why this is a draft. I'd rather hand over an accurate account of what has and hasn't been verified than a tidy split that implies more confidence than the audit actually earned.


---

jlucaso1:
Rebased onto `3da3869` and force-pushed `4f9bb6d` -> `eb0478e`. The body is rewritten. What follows is what actually changed and what I got wrong.

### The E2E blocker was my mistake

I put this PR in draft partly because I believed it caused an E2E regression. It did not. I saw one red E2E run on `39b9a38`, confirmed `main` was green, and concluded "regression" without ever looking at the rest of my own branch's history. That history reads:

| commit | E2E |
| --- | --- |
| `bcabb05` | pass |
| `9923ffb` | pass |
| `2018115` | pass |
| `77906ae` | pass |
| `68f1a49` | pass |
| `39b9a38` | **fail** |
| `6c68e3a` | pass |
| `4f9bb6d` | pass |

One failure in eight runs of the same branch is a flake, and five greens before it would have told me that immediately. I blocked the PR on a check I never performed. I could not reproduce the local mock-server comparison myself (no mock server here), so this correction rests on the CI history above rather than on a local run.

### The "no change at the default config" claim was false, and it cost ~25% p50

The old body said behaviour at `read_pool_size = 0` was unchanged. It was not. On `main`, fourteen reads (not the ten I had listed) run on a raw pooled connection and take **no** permit: `load_prekey`, `load_prekeys_batch`, `load_signed_prekey`, the four app-state reads, `device_exists`, `load_device_data_for_device`, `has_same_base_key`, `get_group_metadata`, and three more. Routing them through `read_query` put all of them behind the single write permit at the default profile, which every consumer runs.

Measured, 16 concurrent `get_sender_key`, default profile:

| | p50 |
| --- | --- |
| `main` | 569-641 us |
| this PR before the fix | 730-796 us |
| this PR now | 525-689 us |

`c780e8b` makes the single-connection fallback take no permit: checking out the one pooled connection already is both the serialization and the snapshot, and adding a permit on top only serialized the `spawn_blocking` dispatch that the pool wait was overlapping. The default profile is back on `main`'s numbers. The body now states the change instead of denying it.

### The clean audit pass moved two more methods out

The migrate/stay criterion changed mid-review from "provably read-only" to "what does the caller do with a stale answer", and I applied the new criterion retroactively, one method at a time, as reviewers found them. I never ran it over the whole set from scratch. Running it now moved two more back onto the write queue:

- `get_sync_key` - a `None` from a stale-absent read produces an orphan `MessageField` on the wire (`src/message/special.rs:175`). The caller does not retry; the wrong answer ships.
- `get_latest_sync_key_id` - a `None` fails the operation with `InvalidRequest("no app state sync key available")` (`src/features/chat_actions.rs:793`).

Neither recovers from a stale miss, so neither qualifies. `## Audit` in the body now lists all 26 methods with what each reads, what the caller does with a stale answer, and the decision - including the ones that already survived review.

The prekey question I raised and never closed (`load_prekey` / `load_prekeys_batch`) is closed as moot for routing: the write-behind window already means a prekey can be read after it was consumed, independent of which pool serves the read. That is a prekey-durability question, not a routing one, and this PR does not change it either way.

### CodeRabbit items

- Widening the routing scan's name filter (`list_`, `count_`, `find_`, `fetch_`, `is_`, `_exists`): already landed in `9923ffb`.
- Applying `read_snapshot` to the no-reads fallback: implemented narrower than asked, on purpose. Unconditionally it would add two statements to every read at the default profile for no isolation gain (one connection is already a snapshot) and would reintroduce the shared-cache hazard `snapshot_safe` exists to avoid. The reason is recorded rather than the item silently dropped.

### Still draft

`get_session` carries the whole measured win (Case B p50 ~37 ms -> ~0.4 ms) and it stays off the read path until the cold-load removal guard lands, because the post-I/O re-check in `signal_cache.rs` cannot distinguish "never written" from "written, persisted and removed". That guard PR does not exist yet (open: 1227, 1222, 1169, 1111, 218), so this stays draft. When it merges I will rebase, restore the `get_session` routing, redo the table, and mark ready.

No merge, no labels - yours to land.

---

jlucaso1:
The binary size gate failed on `eb0478e` at +39.88 KiB .text against a 32 KiB budget. Fixed in `1c76a66` and `6dfe542`, now +24.00 KiB. Here is what it actually was, since my first guess was wrong.

### Why it grew

I reproduced the gate locally (installed `libasound2-dev`, built the `demo` example against `main` and against the branch) and got 8397238 -> 8438070 bytes of .text, byte-for-byte the CI number. Then I diffed symbol sizes with `nm` instead of guessing. Three buckets, not one:

| Bucket | `main` | PR | Delta |
| --- | ---: | ---: | ---: |
| `read_erased`, 186 instantiations, per return type | 0 | 25 098 B | +25 098 |
| `SharedSqlite::read`, per return type | | | +16 106 |
| `with_semaphore`, 77 -> 188 instantiations, per closure *and* return type | 22 263 B | 41 327 B | +9 586 |
| raw per-method bodies the routing replaced | | | -10 514 |

I had assumed it was all `read_erased`. `with_semaphore` was the part I had not accounted for: it is generic over the closure as well as the return type, so each of the ~10 reads I moved *back* onto the write queue during review minted another full copy of the acquire plus `spawn_blocking`.

### The fix

Two dispatch-shape changes, no behaviour change:

- `read_erased` had three branches. The snapshot condition was the same predicate written twice, inverted; hoisting it to an early return leaves two branches that differ only in whether a permit is taken, so they fold into one body with an `Option` permit. Two per-return-type copies become one.
- `with_semaphore` gets the same closure erasure `read_query` already had. One body per return type instead of one per call site, which also shrinks the write path that predates this branch.

Cost of the erasure is one `Box` per call, against a `spawn_blocking` hop and a SQLite round trip. Both fixes are dispatch shape only: same predicate, same ordering, same observable behaviour, and I checked the truth table branch by branch rather than trusting the rewrite.

.text +39.88 -> +24.00 KiB, stripped +53.78 -> +28.34 KiB. CI agrees with the local number exactly.

### One thing I found on the way, which is not good news

Chasing this I re-ran my own falsification check on `a_shared_cache_store_gets_no_snapshot_even_with_a_wider_write_pool` and it was **vacuous**. It asserted `!snapshot_safe`, which is the mechanism, not the consequence. Deleting `snapshot_safe` from the routing predicate left it green. That is the third test on this branch I have had to catch this way, and I only caught it because I happened to be editing the predicate it covers.

It now parks a read mid-flight on a shared-cache store and requires a concurrent write to commit. First attempt parked 300 ms and *still* passed unguarded, because `with_retry` absorbs the lock across its 10/20/40/80/160 ms backoff. Parking past that budget makes the lock fatal, and the unguarded run now fails with `database table is locked: sessions`. So the guard is real and now has a test that says so, but my original justification was overstated: the hazard is retried away unless the reader holds longer than ~310 ms.

`6dfe542` is that test, committed separately from the size fix. Body updated with the attribution table and the corrected validation note (workspace clippy runs clean here now; the earlier "cannot build" was the missing ALSA headers, not the code).

Still draft, still gated on the cold-load guard PR, which does not exist yet.

---

jlucaso1:
Rebased onto `05b3c880` and out of draft. All 5 items closed, with two findings I did not expect.

### 1. Rebase

Clean. `05b3c880` (#1229) and `534054f6` (#1227) both landed under this branch; neither touched `storages/sqlite-storage/`, so the diff is still one file.

### 2. Guard coverage, read by read -- and two of the five were not what this branch claimed

Three mappings held exactly as expected: `get_session_for_device` feeds `checkout_session`/`peek_session`/`has_session` (all three read `backend.get_session`), `load_identity_for_device` feeds `get_identity`, `get_sender_key_for_device` feeds `get_sender_key`. All guarded, all safe to route.

The other two were wrong:

- **`has_signal_state_for_user` has no guard, and is now back on the write queue.** Its consumer is `SignalStoreCache::has_state_for_user`, which is not one of the guarded five: it checks both caches for any matching key, then asks the backend with no removal-seq re-check. Its callers in `lid_pn.rs` turn a `false` into "skip the PN to LID session migration", and nothing retries that skip. Reverted in `0cb522b`.
- **`has_session`'s recorded justification was wrong, though the verdict survives.** The cache's `has_session` reads `get_session`, not this method, so #1229 never covered it. The only path here is `Device::contains_session` -> `check_session_exists`, whose single production caller logs the answer in all three branches. It stays routed because the caller does nothing with it, not because a guard protects it. The comment now says so.

### 3. Re-validation against the new base

Checked rather than assumed: #1227 touched three files this audit cites. The unconditional `device_registry_cache.promote` on a backend hit and the `InvalidRequest("no app state sync key available")` on a `None` latest-key id both still hold, unchanged. Neither #1227 nor #1229 changes what a caller does with a stale answer for any of the other reads.

### 4. Both tables re-measured on `05b3c880`

The win holds and got bigger: `get_session` Case B p50 **86.4-91.4 ms -> 590-602 us**, a factor of ~150.

The default-profile table is the honest half. A first, ordinary main-then-branch run showed the branch about **9% slower**, consistently across 7 rounds. That looked real, so I ran it alternating instead:

| Pass | ran first | `main` | branch | ratio |
| --- | --- | ---: | ---: | ---: |
| 1 | `main` | 532.1 us | 538.6 us | 1.01 |
| 2 | `main` | 609.2 us | 647.9 us | 1.06 |
| 3 | branch | 572.7 us | 545.7 us | 0.95 |
| 4 | branch | 456.9 us | 628.5 us | 1.38 |

No consistent sign, against a within-run spread of 1.65-1.70x. The 9% was ordering drift: whichever tree ran second was slower. **So I am not claiming the default profile is unchanged -- I am saying this box cannot resolve the difference**, and the body says that rather than rounding it to "identical". The structural argument is separate and stronger: after `c780e8b` both sides do `spawn_blocking(pool.get(); query)` with no permit, and the branch adds one `Box`.

For contrast, the 25% regression `c780e8b` fixed *was* resolvable here (569-641 vs 730-796 us, non-overlapping). Nothing that size remains.

### 5. CI

Green on `0cb522b`: E2E, Build & Test, Build & Lint, Feature Matrix, Test Stable, Clippy, Format, Rustdoc, all four Miri jobs, wasm, Cargo Deny, all CodSpeed, Binary Size (+26.12 KiB .text against the 32 budget), and Semver Checks passed this time too.

### Open reviewer threads, verified not blindly accepted

- **CodeRabbit's "three reads stay on the write permit for ordering the permit does not provide"**: two-thirds valid and already documented as an explicit caveat; its fix would change write serialization, which is out of scope here. Its third bullet is **wrong** -- `put_sender_key_for_device`/`delete_sender_key_for_device` write the `sender_keys` table, not `sender_key_devices`. I checked every writer of `sender_key_devices` (`set_sender_key_status`, `clear_sender_key_devices`, `clear_all_sender_key_devices`, `delete_sender_key_device_rows`) and all four go through `with_retry`, so that hold is the one that genuinely orders at any pool size. The caveat in the body now names it as the exception.
- **Codex's P1 cold-load session race**: correct, and it is precisely what #1229 fixed. Closed by the dependency.
- **Codex's prekey upload-window item**: re-verified against the code rather than my earlier reasoning. `buffer_consumed_prekey` deliberately leaves the row in the backend until flush, with a test in `signal_adapter.rs` pinning that ("must not delete from the backend before flush"). So the window is the whole write-behind interval on any routing; this PR widens it by one transaction. Prekey-durability question, not a routing one. Kept, reason in the body.

### One thing I could not do

You asked me to strip the `_Generated by [Claude Code]_` footers from the body **and** my comments. The body is clean and has been for several revisions. The 14 existing comments I cannot fix: no tool in this session can edit an issue comment, and the API token here is empty (403). The footers are appended server-side rather than typed by me, so new comments will likely keep getting one regardless. Flagging it instead of quietly leaving it done-looking.

Ready for review. Not merging and not touching labels.

---

PR #1231, opened 2026-08-07
fix: stop two production log lines from describing what the code does not do

jlucaso1:
## Summary

Two lines from a 13-day production log read (two accounts) that say something other than what the code around them does. Neither has a functional consequence; both cost diagnostics.

**`src/handshake.rs`, the IK to XXfallback pivot.** The line went out at `warn!` whenever the server answered the IK resume with a non-null `serverStaticCiphertext`. That is the server declining to resume, not a failure: `run_ik_handshake` returns `Ok` from that arm, XXfallback finishes in the same round trip, and the next line already reports `Handshake complete (XXfallback)` at `info!`. All 6 occurrences in the period (3 per instance) completed and were followed by `<success>`. A warning there is a false alert on an ordinary reconnect.

**`src/message/receive.rs`, the no-group-content branch.** `"Dispatching UndecryptableMessage event."` sat *outside* the `if !session_dispatched_undecryptable` that decides whether the event goes out. Worth spelling out how often that guard is closed: every ordinary session decrypt failure routes through `handle_decrypt_failure`, which returns `true` unconditionally, so `session_dispatched_undecryptable` is set and this branch dispatches nothing. The line was therefore claiming a dispatch on essentially every message that reached it. In the log it shows up as a contradictory pair microseconds apart, a `[DEBUG]` from `message::retry` saying it skipped a duplicate followed by a `[WARN]` from `message::receive` saying it is dispatching.

The event itself never duplicated. `dispatch_undecryptable_event` (`src/message/retry.rs:64`) is single-flight on `(chat, id)` via `undecryptable_dispatched.get_with`, and the existing `test_undecryptable_dedup_is_atomic` / `test_undecryptable_deduped_across_resends` lock that. So this is a diagnostic fix, not a dispatch fix, and `Event` / `UndecryptableMessage` / the dedup cache are untouched.

## Protocol evidence

The captured bundle is untracked and not present in the environment this was built in, so I could not open `docs/captured-js/WAWeb/Open/ChatSocket.js` myself. What is checked in and does line up: `agent_docs/noise_handshake.md` already lists `[socket] resumeNoiseHandshake failed: serverStaticCiphertext not null` as one of the lines that mirror WA Web's `[socket]` output, and the official client emits that point through `WALogger.LOG` (ChatSocket.js around :265-270), its plain log level, not a warning level. The `"failed"` wording is inherited from there, which is why it stays: it is the string a captured session and a local run are grepped by. The IR is no help here, since it models stanza shape rather than control flow or logging.

Independent of the bundle, our own code says the same thing. Every other `[socket]` handshake trace line in `run_ik_handshake` and `continue_full_handshake_core` is `debug!`; this was the only one above it. The invalidation policy in `agent_docs/noise_handshake.md` treats a completed XXfallback as a success, resetting `ik_handshake_failures` and repopulating the cert chain, and `fallback_taken` exists precisely so a later error is *not* charged to the IK cache.

## Changes

- `src/handshake.rs:337` `warn!` to `debug!`. Level chosen from what the arm does: it returns `Ok`, and the pattern that actually completed is already reported at `info!` one line down, so this line carries only the reason for the pivot and belongs with the other `[socket]` trace lines. Nothing is lost at `info!` verbosity that was not already there.
- Message text unchanged, so existing searches and captured-session comparisons still match.
- `src/message/receive.rs` the `log::log!` moves inside the existing `if !session_dispatched_undecryptable`, before the call it describes. Position chosen so the line shares the condition with the dispatch; the level still comes from `decrypt_fail_log_level(decrypt_fail_mode)`. No condition is added, removed, or altered.
- Replaced the stale comment above the guard, which claimed `process_session_enc_batch` returns `dispatched_undecryptable=false` for this case. It usually returns `true`.
- `agent_docs/noise_handshake.md` one sentence recording that these `[socket]` lines are `debug` and why, so the next person does not re-promote it.

## Tests

Two tests in `src/message/tests.rs` drive `process_classified_message` and assert on both the event and what was announced:

- `undecryptable_receive_branch_stays_silent_when_batch_dispatched` the ordinary shape (session ciphertext for an unestablished session, no group content). The batch owns the dispatch, so the branch must announce nothing, and the event must still be there exactly once. This fails on `main` and passes here.
- `undecryptable_receive_branch_announces_the_dispatch_it_performs` the case the branch exists for, where the session batch never ran and the branch is the dispatcher. One event, one announcement, at the level `decrypt_fail_mode` selects.

The log assertions go through a small `log::Log` capture filtered to the `message::receive` target. `log`'s global logger is install-once per process: under `cargo nextest`, which is what CI runs and what gives each test its own process, the install always wins and the assertions always run. Under a threaded `cargo test --lib` a sibling test's `env_logger` can take the slot first, in which case the log assertions are skipped and the event assertions still hold. The helper reports which happened rather than pretending.

For the handshake change there is no proportionate way to observe a log level: `tests/handshake_integration.rs` has no log capture and building one for a level change is not worth it. The safety net is that the behaviour around the line is already locked there, by `ik_rejected_recovers_via_xxfallback_and_repopulates_cache` (fallback completes, counter resets, cache repopulates), `cold_start_xx_then_cached_ik_reconnect` (IK resume succeeds), and `post_xxfallback_failure_does_not_invalidate_ik_cache`. All still pass unchanged.

## Validation

```
cargo fmt --all
cargo test -p whatsapp-rust --lib            # 1393 passed
cargo test -p whatsapp-rust --test handshake_integration   # 7 passed
cargo clippy -p whatsapp-rust --all-targets -- -D warnings  # clean
```

`cargo clippy --workspace` does not build in this environment: `examples/voip-cli` pulls `alsa-sys`, and the system `alsa` package is absent. Full matrix left to CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vob7urYBCELLHef4Dgb7aQ)_

### DIFF
--- a/agent_docs/noise_handshake.md
+++ b/agent_docs/noise_handshake.md
@@ -55,3 +55,5 @@ These lines mirror WA Web's `[socket]` output, which makes a captured session an
   doFallbackHandshake continuing handshake with given server hello
 [socket] continueFullHandshakeCore client finish and deriving secrets
 ```
+
+All of them are `debug`, including the `resumeNoiseHandshake failed` line: a server that declines the IK resume is the ordinary trigger for XXfallback, not a failure of ours, and the pattern that actually completed is reported separately at `info` ("Handshake complete (IK|XX|XXfallback)"). Only a handshake that gives up carries a warning.
--- a/agent_docs/noise_handshake.md
+++ b/agent_docs/noise_handshake.md
@@ -56,4 +56,4 @@ These lines mirror WA Web's `[socket]` output, which makes a captured session an
 [socket] continueFullHandshakeCore client finish and deriving secrets
 ```
 
-All of them are `debug`, including the `resumeNoiseHandshake failed` line: a server that declines the IK resume is the ordinary trigger for XXfallback, not a failure of ours, and the pattern that actually completed is reported separately at `info` ("Handshake complete (IK|XX|XXfallback)"). Only a handshake that gives up carries a warning.
+Every line above is `debug`, including `resumeNoiseHandshake failed`: a server that declines the IK resume is the ordinary trigger for XXfallback, not a failure of ours. The pattern that actually completed is what gets reported at `info`, as "Handshake complete (IK|XX|XXfallback)". None of these pattern diagnostics warns; other parts of connection setup still do on their own terms (an oversized `edge_routing_info` being dropped, for one).
--- a/src/handshake.rs
+++ b/src/handshake.rs
@@ -334,7 +334,14 @@ async fn run_ik_handshake(
         }
         IkServerHelloOutcome::Fallback(inputs) => {
             *fallback_taken = true;
-            warn!(
+            // Not a failure of ours: the server declined the IK resume and
+            // answered with a full server hello, which XXfallback completes in
+            // the same round trip. WA Web logs this at its plain log level for
+            // the same reason (the "failed" wording is inherited from there).
+            // The outcome is already reported by the `info!` below, so this
+            // line only carries the reason and belongs with the other `[socket]`
+            // trace lines.
+            debug!(
                 "[socket] resumeNoiseHandshake failed: serverStaticCiphertext not null — \
                  doFallbackHandshake continuing handshake with given server hello"
             );
--- a/src/message/receive.rs
+++ b/src/message/receive.rs
@@ -550,17 +550,17 @@ impl Client {
             && !session_had_duplicates
             && !session_payloads_empty
         {
-            // Edge case: message with only msg/pkmsg that failed to decrypt, no skmsg
-            log::log!(
-                decrypt_fail_log_level(decrypt_fail_mode),
-                "Message {} from {} failed to decrypt and has no group content. Dispatching UndecryptableMessage event.",
-                info.id,
-                info.source.sender.observe()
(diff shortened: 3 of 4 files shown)


jlucaso1:
Pushed for the review round:

- Skippable log assertions (Greptile P2, cubic P2) — fair, and it was the wrong failure mode for the only check covering this fix. The capture no longer depends on winning `log`'s install-once slot: when it is already taken the test re-runs itself in a fresh process and propagates the result, so the assertions always run somewhere. Verified by reverting the `receive.rs` change and running the full threaded suite, which now fails through the child instead of passing.
- Handshake logging note — right, it overreached. Narrowed to the `[socket]` pattern diagnostics; the edge-routing size warning is unrelated to pattern selection and still warns.

Skipping one: gating the announcement on `dispatch_undecryptable_event`'s return value. It is a real residual, on a redelivery whose id is already in the dedup cache the branch still says "Dispatching" and the dedup layer then declines. But it needs a new condition, and this change was scoped to moving the line without touching control flow. It also no longer reads as a contradiction: the order is announcement then "already dispatched ... skipping duplicate", which is a coherent sequence, unlike the production pair that motivated this. Worth its own change if the residual bothers anyone.


---

PR #1257, opened 2026-08-08
feat(request): keep the reply that describes an IQ error

jlucaso1:
## Summary

`send_iq_node` hands back the whole `Arc<OwnedNodeRef>` when the server answers `type="result"`. When the same request comes back `type="error"`, the caller got `IqError::ServerError { code, text, error_type, backoff }` and the stanza was thrown away, so whether a reply arrived intact or as a four-field summary depended only on its `type` attribute. Everything outside those four names went with it: further `<error>` attributes, the child elements XMPP allows there, and the raw bytes, which are the only faithful material for recording or replaying a rejection. `ServerError` now carries the reply as the node the receive path had already decoded, which makes the rejection path one refcount bump and leaves the success path with nothing new to do. The four extracted fields stay alongside it, because nearly every match in the tree is `ServerError { code, .. }` and dropping them would make all of those reparse the common case for nothing.

## Changes

- **Breaking.** `IqError::ServerError` gains `response: RejectionStanza`, the rejection stanza verbatim. Matching with `..` is unaffected; constructing the variant now needs the stanza. Migration: pass the response you were rejected with (`Arc<OwnedNodeRef>` converts via `.into()`), or in a fixture build one and decode it through `OwnedNodeRef::new`.
- `RejectionStanza` wraps `Arc<OwnedNodeRef>` rather than `Box<Node>`, the shape `Disconnected` uses: the receive path already holds the `Arc`, so there is no copy, and it keeps `backing_bytes()`. `Box<Node>` would have deep-copied a node we already own and would have dropped the bytes on the floor.
- The wrapper exists for its `Debug`, which names the stanza (`<iq>`) instead of printing it. Background IQ failures are logged with `{e:?}` at warn level on the connect path (`node_io.rs`: the post-connect active IQ, props, blocklist, privacy settings), and an error stanza can carry a JID in its attributes or children; deriving `Debug` straight through the node would have written all of that into production logs, where before only the four summarized fields went. `Deref` keeps every accessor reachable, so reading the contents is unchanged for a caller that wants them, and `as_arc()` / `into_arc()` hand back the shareable node.
- **Breaking.** `From<wacore::request::IqError> for IqError` is replaced by `IqError::from_response(err, &response)`. A `From` cannot take the response, and converting without it is exactly the loss being removed here. Migration: `err.into()` becomes `IqError::from_response(err, &response)`.
- `send_and_wait_iq` passes the node it just classified instead of dropping it. The `Ok` arm is byte-for-byte what it was and still returns that same allocation.
- `execute` needed nothing: it propagates `IqError` untouched, so there was no second place discarding the stanza. Neither `wacore`'s own `IqError` nor `parse_iq_response` changed, since that layer only ever holds a borrowed `NodeRef` and making it own a copy would have cost the deep copy this avoids.
- The two call sites that destructured all four fields, `server_rejection_of` in `error.rs` and the prekey rewrap in `context_impl.rs`, gained a `..`. No behaviour change; every other call site already used `..`.
- Test fixtures build the variant through `test_utils::server_error_iq`, which runs the real `parse_iq_response` over a real `<iq type="error">`, so a fixture cannot drift from what the server actually produces. Existing tests changed only where they constructed the variant by hand, which is the contract change itself and not an adaptation to make them pass.

New tests in `request.rs`: a rejection carrying an unmodelled `<error>` attribute, an `<error>` child and an `<iq>`-level attribute, all asserted on the preserved node rather than on the error string, with the four fields checked for parity and the redacted `Debug` pinned in the same test; the corresponding failure with `type` and `backoff` absent; and `Arc::ptr_eq` on both the `type="result"` and `type="error"` paths, which is what proves neither picked up a second parse.

## Validation

```
cargo fmt --all
cargo test -p whatsapp-rust --lib             # 1444 passed, 0 failed
cargo test -p whatsapp-rust --test error_surface   # 26 passed, 0 failed
cargo clippy -p whatsapp-rust --all-targets -- -D warnings   # clean
```

Full matrix left to CI.

### DIFF
--- a/src/client/context_impl.rs
+++ b/src/client/context_impl.rs
@@ -36,6 +36,7 @@ impl SendContextResolver for Client {
                     text,
                     error_type,
                     backoff,
+                    ..
                 }) = e.downcast_ref::<crate::request::IqError>()
                 {
                     return anyhow::Error::new(wacore::request::ServerErrorCode {
--- a/src/client/sessions.rs
+++ b/src/client/sessions.rs
@@ -629,12 +629,12 @@ mod tests {
         // `fetch_pre_keys` directly and receives the first, while the fan-out
         // goes through `SendContextResolver`, which re-wraps it as the second.
         let as_iq_error = |code| {
-            anyhow::Error::new(crate::request::IqError::ServerError {
+            anyhow::Error::new(crate::test_utils::server_error_iq(
                 code,
-                text: "not-acceptable".to_string(),
-                error_type: None,
-                backoff: None,
-            })
+                "not-acceptable",
+                None,
+                None,
+            ))
         };
         let as_shared = |code| {
             anyhow::Error::new(wacore::request::ServerErrorCode {
--- a/src/error.rs
+++ b/src/error.rs
@@ -215,6 +215,7 @@ fn server_rejection_of<'a>(cause: &'a (dyn StdError + 'static)) -> Option<Server
         text,
         error_type,
         backoff,
+        ..
     }) = cause.downcast_ref::<ClientIqError>()
     {
         return Some(ServerRejection {
--- a/src/features/mex.rs
+++ b/src/features/mex.rs
@@ -363,12 +363,7 @@ mod tests {
 
     #[test]
     fn request_preserves_iq_error_source() {
-        let iq = IqError::ServerError {
-            code: 404,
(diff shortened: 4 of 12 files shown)


jlucaso1:
@codex review

PR #1258, opened 2026-08-08
fix(client): make it plain which call reads the socket

jlucaso1:
## Summary

A consumer that calls `connect().await` and then waits for events waits forever: the handshake completes, the transport receives bytes, and nothing decodes them, because `read_messages_loop` runs inside `run`, not inside `connect`. The transport channel buffers, so the frames just sit there with no reader; there is no timeout, no error and no log that says so, which is why this reads as a broken server and costs hours to pin down. Documenting it is not enough for a bug whose only symptom is silence, so `connect` now returns a `#[must_use] Connection` that has to be driven with `read_until_disconnected`, and the compiler reports the mistake at the call site instead. Driving one connection is exactly what `run`'s loop body already did (read loop, teardown, `Disconnected` dispatch), now extracted and shared with it, so the direct path and the reconnect loop cannot drift on what a connection ending means. Connecting without the reconnect loop stays supported and finally has a way to be read.

## Audit

Each of these was checked against the tree before designing anything:

- `connect` had a doc comment, but it only explained the boxed future (`src/client/lifecycle.rs:707-710` before this change). Nothing said the established connection is not read.
- `run` had no item doc at all (`src/client/lifecycle.rs:552`), only the internal note about deliberately not instrumenting its span.
- `read_messages_loop` is `pub(crate)` (`src/client/node_io.rs:177`) and is called from exactly one place, `run`'s loop body (`src/client/lifecycle.rs:598`). A consumer could neither call it nor discover it from the public API.
- Nothing denounces the situation at runtime. The events channel is a bounded `async_channel` filled by the transport task, and its only consumer is the read loop, so the frames simply queue. The keepalive loop was spawned by `connect`, but its ping goes through `can_reach_server`, which requires `is_running` and so refuses on a client that never started a reader: even the dead-socket watchdog never fires.
- No example, README snippet or `agent_docs/` page uses `connect` without `run`. Every entry point in the tree goes through `Bot::run`/`Bot::spawn` (`README.md:57`, `examples/*.rs`, `tests/e2e/src/lib.rs`), which call `Client::run` (`src/bot.rs:546`, `src/bot.rs:557`). The only in-tree callers of `Client::connect` are `run` itself and two tests (`src/client/builder.rs:800`, `src/client/lifecycle.rs:1515`), so nothing depended on connecting and not reading.
- The concept of a client with no supervision loop is real and documented in the code (`is_terminal` and `can_reach_server` both reason about it), so the design had to keep it working rather than make `connect` private.

Making that concept a first-class path turned up three things the audit had not, each harmless only while nothing read a directly established connection:

- `expected_disconnect` was cleared only at the top of `run`'s loop (`src/client/lifecycle.rs:583`), so a caller driving connections itself carried the retired connection's verdict into the next one.
- `connect` had no shutdown check, unlike `run`, so a connection established after `disconnect` was one the application had already been told did not exist. One check is also not enough: a version fetch, a transport, a handshake and the slot locks are awaited after it.
- The keepalive was spawned at the end of `connect`, so a tick taken before anything read the connection was refused as `NotConnected`, which its loop classifies as fatal and exits on.

## Changes

- **Breaking:** `Client::connect` resolves to `Result<Connection<'_>, ConnectError>` instead of `Result<(), ConnectError>`. `Connection` is `#[must_use]`, so the old `client.connect().await?;` becomes an `unused_must_use` rather than silently doing nothing. Migration: `client.connect().await?.read_until_disconnected().await;`, or `client.run().await` for a session that reconnects. A `compile_fail` doctest on `connect` pins the old form as no longer accepted, next to a `no_run` twin for the new one.
- `Connection::read_until_disconnected` is the public single-connection read: read loop, connection teardown, `Disconnected` dispatch, returning the disconnect reason (`None` when there was no failure to report). No reconnect, so a session that must not outlive its first connection now has a supported route.
- `run`'s per-connection half is extracted into `Client::drive_connection` and shared with the above. Pure extraction: same order, same logs, same event, and `run` keeps `telemetry::connect("ok")` at the same point.
- `connect` clears `expected_disconnect` with the rest of the per-connection state. Only `run` cleared it before, so a connection driven directly after an end that sets the flag (the 515 that follows pairing) came up with it still set: `handle_success` refuses the login and the read loop exits on the first frame it decodes. `run` clears it before it connects, so its path is unchanged.
- **Breaking (additive):** `connect` refuses a shut-down client with the new `ConnectError::Shutdown`, the same refusal `run` already makes. It reads the signal again before the handshake, and once more after the slots are filled, undoing the publish if a shutdown landed across those awaits: `signal_shutdown_sync` runs no cleanup of its own, so a socket installed there would outlive the client that announced it was gone. `run` reports the abandonment at debug rather than error, since it is about to exit on the same shutdown. `ConnectError` is `#[non_exhaustive]`, so the new variant only affects code matching it exhaustively.
- The keepalive starts where the reading starts rather than at the end of `connect`, so it cannot take a tick before a reader exists and exit as fatal. For `run` that is the same instant; for a directly driven connection it is the difference between having a dead-socket watchdog and not.
- A connection driven directly claims `is_running` for its duration and gives it back through a scope guard, so a dropped read (a wrapping timeout, an aborted task) does not leave the flag standing. That flag is already the client's answer to "is anything reading" (`can_reach_server`, `is_terminal`); leaving it false would refuse every request sent over a connection that does have a reader, and leaving it stuck true would advertise a reader that is gone. `run` sets it before connecting, so its path is untouched by this.
- A `Connection` dropped without being read logs a warning. That is the answer to the "nobody reads for N seconds" question: there is no N. Drop is the moment we know for certain, it costs no timer and no task, and nothing legitimate connects and then never reads.
- Item docs on `run` and `connect` stating which one reads, plus corrections to the `is_terminal`, `can_reach_server` and `stop_supervision_loop` docs, which described `is_running` as "run()'s supervision loop" and now have a second reader to account for.
- No runtime cost on the happy path: `Connection` borrows the client, so `size_of::<Result<Connection, ConnectError>>() == size_of::<Result<(), ConnectError>>()` (asserted in a test) and connect allocates exactly what it did before. The added shutdown reads are loads on a sticky flag, and the retraction allocates nothing on the path that succeeds.

Four things the AI reviewers raised that this deliberately does not do:

- **Teardown on cancellation.** An abandoned `read_until_disconnected` does not run `cleanup_connection_state`, so the socket stays open and the client keeps reporting itself connected until `disconnect()` releases it. That is not new and not specific to this path: it is what dropping the `run` future has always done, and `cleanup_connection_state` is async, so no drop guard can await it. Spawning a detached teardown instead would be worse than the state it fixes: cleanup only clears the transport slots before the close it awaits, which is safe today because every caller awaits cleanup before connecting again, and a detached one could strip a connection the consumer establishes right after. The reader flag is restored because that one is a synchronous store, and the rest of the contract is documented on the method.
- **A shared gate serializing the last shutdown read with `is_connected.store(true)`.** What remains after the retraction above is a window with no awaits in it, and losing it does not strand anything: the shutdown signal is sticky (`wacore/src/runtime.rs:252-273`), and both `disconnect` and `signal_shutdown_sync` fire `notify_connection_shutdown`. A shutdown before the connection's notifier is reset is caught by the retraction; one after it leaves the fresh notifier fired, so the read loop returns `Expected` on its first poll and `drive_connection` tears the connection down in full. Closing that window properly would mean a new cross-cutting mutex on the connect path that every terminal path must take, including `signal_shutdown_sync`, which exists precisely to be callable from `Drop` where nothing can be awaited or blocked on. That is a worse trade than the self-healing state it replaces.
- **Refusing a second reader when `is_running` is set.** As proposed this would break `run`, which sets `is_running` before it connects and then drives through the same call. Two concurrent readers are also not reachable: a `Connection` only exists after `is_connected` is true, and `connect` refuses with `AlreadyConnected` from that point on. A hypothetical second reader would in any case fail immediately, because `read_messages_loop` takes the transport-event receiver out of its slot.
- **A distinct outcome for a protocol-driven reconnect.** A 515 ends the connection with no failure to report, so it returns `None` like a requested end does. Telling the two apart would mean a new public outcome type for a distinction a caller acts on the same way (connect again), so this documents what `None` covers instead.

Tests, all in `src/client/lifecycle.rs`:

- `a_driven_connection_decodes_the_first_frame_and_reports_its_end` is the regression: a frame is queued on a connection, the recommended path is driven, and the assertion is on the decoded node (an `Event::RawNode`) and on the returned disconnect reason, not on a log.
- `a_connection_nobody_reads_leaves_the_frame_queued` pins what connecting alone guarantees (the socket is up) and what it does not (the frame is still queued, no node decoded, the reader's inlet untouched). It is the reported symptom, turned into a test.
- `a_cancelled_read_gives_the_reader_flag_back` covers the failure side of claiming `is_running`: the read is cancelled mid-connection and the flag must come back.
- `connecting_clears_the_retired_connection_expected_disconnect` covers the inherited-verdict fix; it fails without it.
- `connecting_is_refused_after_the_client_is_shut_down` covers the shutdown refusal at the entry point. The reads inside the connect graph are not covered by tests of their own, and neither is the interleaving the reviewers asked for: reaching them needs a handshake to complete, which no unit fixture can do without a mock server.
- `handing_back_a_connection_costs_the_caller_nothing` asserts the size equality above.
- The 1444 existing lib tests pass unchanged. The one that could have needed adapting, `connect_rejects_an_already_connected_client`, uses `expect_err` and so requires `Debug` on the Ok type; `Connection` implements it rather than the test changing.

## Validation

```
cargo fmt --all
cargo test -p whatsapp-rust --lib          # 1447 passed
cargo test -p whatsapp-rust --doc connect  # both doctests, including the compile_fail migration one
cargo clippy -p whatsapp-rust --all-targets -- -D warnings
cargo check -p whatsapp-rust --all-features --all-targets
```

Both new guarantees were checked against their own absence: with `#[must_use]` removed the migration doctest reports "Test compiled successfully, but it's marked `compile_fail`", and with the `expected_disconnect` reset removed its test fails. The full workspace clippy could not run locally (the `voip-cli` example needs libasound, absent in this environment); full matrix left to CI.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -705,6 +705,10 @@ pub enum ConnectError {
     /// Construction never completed, so the attempt was rejected before any I/O.
     #[error("client construction did not activate")]
     NotActivated,
+    /// The client was shut down. Shutdown is final, so build a new client
+    /// rather than reconnecting this one.
+    #[error("client has been shut down")]
+    Shutdown,
     /// A step of the connect flow ran out of time.
     #[error("{stage} timed out after {timeout:?}")]
     Timeout {
@@ -768,6 +772,7 @@ impl ConnectError {
             ConnectError::Handshake(handshake) => handshake.is_timeout(),
             ConnectError::AlreadyConnected
             | ConnectError::NotActivated
+            | ConnectError::Shutdown
             | ConnectError::Version(_)
             | ConnectError::Transport(_) => false,
         }
--- a/src/client.rs
+++ b/src/client.rs
@@ -25,6 +25,7 @@ use extension_lifecycle::LifecycleRegistration;
 #[cfg(feature = "client-lifecycle")]
 #[cfg_attr(docsrs, doc(cfg(feature = "client-lifecycle")))]
 pub use extension_lifecycle::{ClientLifecycle, ConnectionScope, ConnectionScopeState};
+pub use lifecycle::Connection;
 pub use voip::{CallError, Voip};
 
 use crate::cache::Cache;
--- a/src/client/lifecycle.rs
+++ b/src/client/lifecycle.rs
@@ -952,6 +952,21 @@ impl Client {
         *self.transport.lock().await = Some(transport);
         *self.transport_events.lock().await = Some(transport_events);
         *self.noise_socket.lock().unwrap_or_else(|p| p.into_inner()) = Some(noise_socket);
+
+        // The last word, because the slot locks above are the final awaits a
+        // shutdown could land across, and `signal_shutdown_sync` runs no
+        // cleanup of its own. Nothing has been announced yet, so undoing the
+        // publish here is the whole retraction.
+        if self.shutdown_signal().is_fired() {
+            let orphan = self.transport.lock().await.take();
+            *self.transport_events.lock().await = None;
+            *self.noise_socket.lock().unwrap_or_else(|p| p.into_inner()) = None;
+            if let Some(orphan) = orphan {
+                orphan.disconnect().await;
+            }
+            return Err(ConnectError::Shutdown);
+        }
+
         self.is_connected.store(true, Ordering::Release);
 
         // Notify waiters that socket is ready (before login)
--- a/src/client/lifecycle.rs
+++ b/src/client/lifecycle.rs
@@ -1508,12 +1508,16 @@ impl Connection<'_> {
         // sent over this connection would be refused for want of the reader
         // that is right here. `run` sets it before connecting and owns clearing
         // it, so only a directly driven connection claims it.
+        // Guarded, not a store after the await: a caller that drops this future
+        // (a wrapping timeout, an aborted task) stops reading too, and a flag
+        // left standing would advertise a reader that no longer exists.
         let owns_reader = !this.client.is_running.swap(true, Ordering::SeqCst);
-        let reason = this.client.drive_connection().await;
-        if owns_reader {
-            this.client.stop_supervision_loop();
-        }
-        reason
+        let _release_reader = owns_reader.then(|| {
+            scopeguard::guard((), |_| {
+                this.client.stop_supervision_loop();
+            })
+        });
+        this.client.drive_connection().await
     }
 }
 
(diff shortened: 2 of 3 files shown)


jlucaso1:
@codex review

PR #1259, opened 2026-08-08
fix(binary): make a received stanza sendable as it stands

jlucaso1:
## Summary

A plaintext frame is a format byte followed by node bytes. `Encoder::new*` writes that byte, the receive path strips it in `unpack`/`unpack_bytes`, and the decoder never sees it. Neither side said so: `OwnedNodeRef::backing_bytes` documented itself as "exactly what the decoder consumed", and `send_raw_bytes` documented "a valid WABinary-marshaled stanza" without noting that the two differ by one byte at the front. Forwarding or replaying a received stanza therefore needed a conversion nothing in the library named, and getting it wrong type-checks: the server reads the first node byte (`0xf8`) as pack flags, sees the compressed bit set, fails to inflate, and closes the connection.

`unpack` had no inverse, so the fix is to add one rather than move the byte or redefine either side's contract. `util::pack` is that inverse and the only place node bytes get a format byte put back on; `marshal::unmarshal_packed_ref` is the matching decoder for a buffer that still carries one. The byte's value now lives in named constants instead of three `write_u8(0)` literals and two `& 2` masks. Both doc comments state what the other side expects, and `send_raw_bytes` rejects a payload that is not shaped like the ones we produce instead of forwarding it to the socket.

Alternatives declined: making `backing_bytes` include the byte would break its meaning (the buffer the decoder consumed) and cannot be done without a copy, since `unpack_bytes` advances past the byte and the compressed path never had one; making `send_raw_bytes` accept node bytes would silently redefine a documented public function and force the byte to be written somewhere other than the encoder, for every existing caller's benefit of nothing.

## Changes

- `wacore-binary::util`: `FORMAT_PLAIN` / `FORMAT_COMPRESSED` name the format byte, with the one explanation of what it is; `pack` is the inverse of `unpack`; `check_plain_payload` is the shape check, returning the reason (wrong format byte, or no stanza behind it) rather than a bool, so no caller restates the comparison. It accepts only `FORMAT_PLAIN`, since the compressed form is a legitimate inbound frame but nothing any `marshal*` writes. The encoder and `unpack` now reference the constants rather than literals, so the wire value has a single source.
- `marshal::unmarshal_packed_ref` decodes a buffer that still carries the format byte, sharing `check_plain_payload` with the send path. Compressed input is refused deliberately: the returned `NodeRef` borrows the input, so inflated bytes would have nowhere to live.
- `BinaryError::UnexpectedFormatByte(u8)` reports the refusal with the offending byte. **Breaking** for anyone matching `BinaryError` exhaustively; migration is a new arm or a `_` catch-all.
- `OwnedNodeRef::backing_bytes` and `Client::send_raw_bytes` doc comments each now state the other side's expectation, and both name `util::pack` as the conversion.
- `Client::send_raw_bytes` refuses a payload that is not one (`SocketError::Marshal(BinaryError::UnexpectedFormatByte | EmptyData)`). **Behaviour change** on a public function, but only for inputs the server was already going to reject by hanging up; every in-tree caller passes marshal output and is unaffected. `send_raw_bytes_burst` is `pub(crate)` and fed only by `marshal_node_for_send`, so it stays unchecked and untouched.
- Every hand-rolled `[1..]`, `slice(1..)` and `remove(0)` that re-derived the rule (30 sites in `src/message/tests.rs` and `src/client/tests.rs`, plus `src/test_utils.rs`, `wacore/src/stanza/call.rs`, `wacore/tests/binary_protocol_test.rs`, `wacore/binary/tests/*` and both bench files) now goes through `unmarshal_packed_ref` or `unpack`. The decode benchmarks additionally take the format byte off in `with_inputs`, so the timed region is the decoder alone rather than the decoder plus a slice.

## Cost

Hot send path (`marshal_exact` → `send_raw_bytes`): unchanged, zero new allocations. The added guard is a `split_first` and two compares against a buffer already in cache.

Forwarding path (`pack(backing_bytes())` → `send_raw_bytes`): one allocation and one copy of the stanza, which is exactly what the hand-rolled conversion cost, so nothing regressed. Avoiding it would mean carrying the byte separately through `SendJob` and the batch encoder, since the receive buffer cannot grow at the front: `unpack_bytes` advances past the byte before freezing, and the compressed path never held one. That is a change to the noise sender for a path that is not the hot one, and the sender already copies any plaintext at or below 16 KiB into its batch buffer, so the saving would be one memcpy out of two.

Decode benchmarks lost a slice from their timed region; `unmarshal_packed_ref` adds a `split_first` and two compares where it replaced `unmarshal_ref(&buf[1..])` in tests.

## Validation

```
cargo fmt --all
cargo test -p wacore-binary                # 134 lib + all integration tests, including the proptest round-trip
cargo test -p wacore --lib                 # 1374 passed
cargo test -p wacore --test binary_protocol_test
cargo test -p whatsapp-rust --lib          # 1443 passed
cargo clippy -p wacore-binary -p wacore -p whatsapp-rust --all-targets -- -D warnings
```

No existing test needed adapting. New coverage: `pack(backing_bytes())` reproduces the marshal output for a stanza with children, one with a binary body and one with packed hex/nibble attributes; `unpack_bytes` agrees with `unpack` on all three; a regression test pins the format byte to uncompressed and asserts the decoder neither consumes nor requires it; and the failure cases (node bytes, an empty buffer, a lone format byte, or a compressed one offered where a plain packed payload belongs) are refused with the reason named, both in `wacore-binary` and at `Client::send_raw_bytes`, where a client-level test also decrypts the captured frame and compares it to the original marshal output.

`--all-targets` clippy across the whole workspace cannot finish in this environment (`alsa-sys` has no system headers here, unrelated to this change); full matrix left to CI.

### DIFF
--- a/src/client/messaging.rs
+++ b/src/client/messaging.rs
@@ -3,15 +3,26 @@
 use super::*;
 
 impl Client {
-    /// Send pre-marshaled plaintext bytes through the noise socket.
+    /// Send a pre-marshaled stanza through the noise socket.
     ///
-    /// The bytes must be a valid WABinary-marshaled stanza (as produced by
-    /// `wacore_binary::marshal::marshal_to`). Sending malformed data will
-    /// cause the server to close the connection.
+    /// The bytes must be a packed payload: the format byte followed by the node
+    /// bytes, which is what every `wacore_binary::marshal::marshal*` function
+    /// writes. A stanza that came off the wire is one byte short of that, since
+    /// `OwnedNodeRef::backing_bytes()` returns node bytes only and the receive
+    /// path stripped the format byte, so forward it through
+    /// `wacore_binary::util::pack`. Anything else the server answers by closing
+    /// the connection, which is why the format byte is checked here.
     ///
     /// This bypasses node logging and `sent_node_waiter` resolution — use
     /// [`send_node`](Client::send_node) for normal stanza sending.
     pub async fn send_raw_bytes(&self, plaintext: Vec<u8>) -> Result<(), ClientError> {
+        if !wacore_binary::util::is_packed_payload(&plaintext) {
+            let error = match plaintext.first() {
+                Some(&format) => wacore_binary::BinaryError::UnexpectedFormatByte(format),
+                None => wacore_binary::BinaryError::EmptyData,
+            };
+            return Err(SocketError::Marshal(error).into());
+        }
         let noise_socket = self.get_noise_socket()?;
         // Wire bytes and the last-sent timestamp are recorded by the noise
         // sender task at the actual transport write.
--- a/src/client/messaging.rs
+++ b/src/client/messaging.rs
@@ -16,7 +16,7 @@ impl Client {
     /// This bypasses node logging and `sent_node_waiter` resolution — use
     /// [`send_node`](Client::send_node) for normal stanza sending.
     pub async fn send_raw_bytes(&self, plaintext: Vec<u8>) -> Result<(), ClientError> {
-        wacore_binary::util::check_packed_payload(&plaintext).map_err(SocketError::Marshal)?;
+        wacore_binary::util::check_plain_payload(&plaintext).map_err(SocketError::Marshal)?;
         let noise_socket = self.get_noise_socket()?;
         // Wire bytes and the last-sent timestamp are recorded by the noise
         // sender task at the actual transport write.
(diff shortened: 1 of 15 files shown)


jlucaso1:
Semver Checks (informational) is red on the one break this PR sets out to make: `enum_variant_added` for `BinaryError::UnexpectedFormatByte`, in `wacore-binary`. Nothing else failed, and `waproto` came back clean.

That variant is the whole point of the change — it is what lets a payload with the wrong format byte be refused with a reason instead of being handed to the socket — so I have left it as is rather than working around the check. Migration for an exhaustive `match` on `BinaryError` is one new arm or a `_`. The version bump the check suggests is a release call, so I have not touched the workspace version.

---

jlucaso1:
@codex review

PR #1263, opened 2026-08-10
feat(client): report a 429 stream error as a StreamError event

jlucaso1:
## Summary

Two behaviours were flagged as possible defects. Ground truth settles them differently.

**Changes:** the `429` arm of `handle_stream_error` now dispatches `Event::StreamError`, like every other coded arm. No internal behaviour moves — `is_logged_in`, the `+5`, and the reset suppression are untouched.

**Does not change:** the pre-key retry loop. WA Web's loop has no attempt limit either, and for the same reasons. Recorded in a doc comment so the next investigation does not reopen it.

**Not changed but worth a decision:** the 429 backoff ladder diverges hugely from WA Web. Evidence below, no code touched — see *Unchanged, and why*.

## Ground truth

Bundle: `waVersion 2.3000.1044659339`, the 497 files pinned in whatspec's `generated/bundles.lock.json`, each fetched from `static.whatsapp.net` and sha256-verified against the lockfile. `docs/captured-js/` was not present locally, so the pinned set is the bundle. whatspec IR was queried for the WAM catalogue; every control-flow answer below comes from the bundle, because the IR models shape and not sequencing.

### Observation A — `<stream:error code="429">`

`WAWebHandleStreamError` parses the stanza into `conflict` / `code` / `ack` / `xml-not-well-formed` / `other`, then:

- `type === "code" && code >= 500 && code < 600`: `515` restarts login, `516` forces logout, everything else calls `WAComms.onStreamErrorReceived()`.
- otherwise: arms for `device_removed`, `replaced`, `xml-not-well-formed`.
- fallthrough: `return "CLOSE_SOCKET"`.

**429 is not in `500..600`.** It matches no arm at all and reaches the fallthrough. What WA Web makes observable for a rate-limited session, in full:

- no log in the handler (the `WARN "Unrecognized stream:error"` belongs to the parser's `other` branch, which a coded stanza never reaches),
- `[comms] job response is CLOSE_SOCKET` and `[comms] Socket N closed` in `WAComms` — emitted for every socket close,
- `onConnectionChange`, whose vocabulary is `disconnected` / `in_handshake` / `connected` (`WANotifyConnectionChangeFactory`) and carries no reason,
- no WAM event: the catalogue has no stream-error event, and `WebcSocketConnect`'s only reason field is `webcSocketConnectReason`, written on connect by `WAWebOpenChatSocket`.

So: **nothing distinguishes "rate limited" from "disconnected"** in WA Web. The missing event is not a fidelity defect. It is a gap in *our* contract, since `Event::StreamError` is the channel every other coded branch reports through and 429/503 are the two silent exclusions — hence the change, marked in the code as choice rather than fidelity.

Two further answers to the questions asked:

- **Backoff scaling.** WA Web applies none for 429. The reconnect ladder lives entirely in `WAComms.socketLoop`, a `PromiseRetryLoop` over `socketLoopIteration`. `onStreamErrorReceived()` — the only per-code backoff lever, reached by 5xx and never by 429 — is `socketLoop.cancelReset()`, which nulls the early-reset deadline armed by `resetTimeoutAfter(1e4)` when a socket connects. It does not touch `resetDelay: 3e4`. There is no per-code counter increment anywhere.
- **Surviving a reload.** No. The ladder's state is the closure inside `WAPromiseBackoffs.createTimer`, held by a `PromiseRetryLoop` instance created at comms construction. A page reload starts at zero.

The production ladder config, `WAWebCommsConfigBase`, confirms our constants exactly:

```js
var s = 9e5;
function u(){return {jitter:.1, max:s, algo:{type:"fibonacci", first:1e3, second:1e3}, relativeDelay:!1}}
maxSocketLoopWaitTime: s, socketReconnectBackoffAlgo: u(),
```

`fibonacci_backoff`'s doc comment states this verbatim, and the sequence matches to the millisecond (`createTimer`'s first call returns `0` and is consumed priming the loop, so the retry delays are 1s, 1s, 2s, 3s, 5s, 8s…, capped at 900s). One cosmetic deviation: WA Web's jitter is `Math.ceil(i*(1+0.1*Math.random()))`, i.e. 0..+10%; ours is ±10%. Left alone.

### Observation B — the pre-key upload retry loop

`WAWebUploadPreKeysJob`, in full:

```js
var g = 812;
var v = {algo:{type:"fibonacci", first:1e3, second:2e3}, max:61e4};
new PromiseRetryLoop({name:"uploadPreKeys", timer:v, code: async (end) => {
  var n = await _uploadPreKeys();
  (n?.success === true) ? (LOG("uploadPreKeys: done"), end()) : LOG("uploadPreKeys: retrying (after delay)");
}});
```

Answering each question:

- **Attempt limit: none.** `max` is a delay clamp, and provably so — `WAPromiseBackoffs`'s `d()` reads it as `o != null && i > o && (i = o)`, and nothing else reads the field. `PromiseRetryLoop` carries no attempt counter; its only terminating exit is `endWithValue`, called on `success`. Our `MAX_DELAY_SECS = 610` is the correct reading of `max: 61e4`.
- **Rate-limit special case: none.** `_uploadPreKeys` branches the error three ways for logging only — `>=500` "server requested backoff", `406` "uploaded invalid keys", anything else (where 429 lands) "unrecognized error" — and all three `return {errorCode, errorText}`, which is not `success`, so all three retry identically. A disconnect mid-upload is caught, logged "disconnected, server state unknown", and returns `undefined`, which also retries.
- **Key count: 812.** `g = 812`, exported as `UPLOAD_KEYS_COUNT`. Our `DEFAULT_WANTED_PRE_KEY_COUNT` is exact. Baileys independently agrees (`INITIAL_PREKEY_COUNT = 812`).

Two places where we already differ, both in the safer direction, both left alone: our `!is_logged_in` bail is an exit WA Web does not have (it awaits `WAComms.waitForConnection()` and keeps going), and WA Web's IQ send is `deprecatedSendIqWithoutRetry`, so the loop is its only retry — as here.

A related suspicion also checked and cleared: the window is abandoned on failure, so every retry ships 812 *fresh* keys rather than re-offering the same ones. That is WA Web's behaviour too — `markKeyAsUploaded(lastKeyId)` runs in the promise chain *before* `deprecatedSendIqWithoutRetry`, moving the watermark past the batch whatever the IQ answers.

## Measurements

**The 429 reconnect ladder** (`handle_stream_error` + `fibonacci_backoff`), reproducing the numbers this task was filed with:

| 429 | `auto_reconnect_errors` | next attempt | cumulative offline |
| --- | --- | --- | --- |
| #1 | 5 | 8.4s | 8.4s |
| #2 | 11 | 146.2s | 154.6s |
| #3 | 17 | 813.9s | 968.5s |
| #4 | 23 | 903.5s | 1872.0s |

WA Web, from the same event and no penalty: 1s, 1s, 2s, 3s — about 7s of cumulative downtime over four consecutive 429s, against our 1872s. Its ladder still escalates on repeated failures, because each short-lived socket is one un-reset loop iteration; what it lacks is the `+5` jump-start.

**The pre-key loop against a permanent 429**, which the earlier investigation could not exercise. Harness: `CapturingMockTransport` + a test noise socket, with a responder subscribed to `Event::SentFrame` that answers every `<iq xmlns="encrypt">` with `<iq type="error"><error code="429" text="rate-overlimit"/></iq>`, on `tokio::time` paused so an hour runs in milliseconds.

- **18 uploads in one virtual hour**, at t = 0, 1, 3, 6, 11, 19, 32, 53, 87, 142, 231, 375, 608, 985, 1595, 2205, 2815, 3425 s — 12 of them inside the first 10 minutes, then steady state at one per 610s.
- **40,800 bytes on the wire per attempt**, ≈490 KB over the first 10 minutes and ≈734 KB per hour.
- **The loop never exits.** The 3600s timeout is what ended the measurement.
- Each attempt's first key id advances by 812 (1, 813, 1625, 2437, …), confirming the abandon-and-regenerate behaviour described above.
- An IQ-level 429 does not clear `is_logged_in`, so the disconnect bail never fires — as suspected, and as WA Web behaves.

These numbers are WA Web's numbers too: same count, same schedule, same 812 keys.

## Second opinions

| | 429 stream error | pre-key upload retry |
| --- | --- | --- |
| **WA Web** | no arm; `CLOSE_SOCKET`, no reason surfaced, no penalty; proactively closes the socket | unbounded, no rate-limit case, 812 keys |
| **whatsmeow** | `default:` arm → `ERROR "Unknown stream error"` + `dispatchEvent(events.StreamError{Code, Raw})`; clears `isLoggedIn` for every code; no penalty | n/a |
| **Baileys** | `ERROR "stream errored out"` + `end(Boom(statusCode))`, surfaced to the consumer via `connection.update`; no per-code arm | `retryCount < 3`, `min(1000·2^n, 10000)` backoff, raced against `UPLOAD_TIMEOUT`; `INITIAL_PREKEY_COUNT = 812` |

Where they disagree: on observability, both second opinions hand the embedder something the moment a 429 arrives, and WA Web hands its UI nothing — which is the split this PR resolves in the embedder's favour. On the retry loop they disagree with each other, and WA Web breaks the tie: Baileys' 3-attempt cap has no counterpart in the official client (the task's note that Baileys has no retry at all is out of date — it gained one, just a bounded one).

## Unchanged, and why

- **The pre-key retry loop.** Unbounded is what WA Web does, for every error class including 429, at the same 812 keys. Evidence recorded in the doc comment on `upload_pre_keys_with_retry`. No test, since nothing changed.
- **The `429` backoff ladder — flagged, not touched.** This is the one place ground truth says we diverge, and it diverges by 1872s against 7s over four errors. WA Web has no `+5` and no `cancelReset` for 429; `cancelReset` is reached only by 5xx. Left alone because the task scoped ladder changes out and because aligning it properly means touching the shared reconnect path and the `503` arm together, not the 429 arm alone. It is the highest-value follow-up here, and I'd suggest it get its own task.
- **The `503` arm.** Also silent, and the conclusion for 429 does apply to it. It is worse than a mirror image: WA Web's 5xx arm *does* call `onStreamErrorReceived()` while ours applies no penalty at all, so the two arms are inverted relative to ground truth. Deliberately left out of this diff — its own task.
- **Jitter in `fibonacci_backoff`.** WA Web's is one-sided upward, ours is symmetric. Cosmetic at this amplitude.

## Validation

```
cargo fmt --all
cargo test -p whatsapp-rust --lib     # 1583 passed
cargo test -p wacore --lib            # 1413 passed
cargo clippy -p whatsapp-rust -p wacore --all-targets -- -D warnings
```

`cargo nextest` is not installed in this environment, so the same tests ran under `cargo test`. `cargo clippy --workspace` cannot complete here: `alsa-sys` fails to build for want of `alsa.pc`, unrelated to this diff — the two touched crates are clean. E2E was not run (no mock server).

Both new tests were verified to fail with the dispatch removed and pass with it restored.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LhiV7hsjVfNo4hbHhq49yk)_

### DIFF
--- a/src/client/node_io.rs
+++ b/src/client/node_io.rs
@@ -1905,6 +1905,18 @@ impl Client {
                     // Deliberate rate-limit backoff: the stability reset must
                     // not erase it even if the connection had been up >= 30s.
                     self.backoff_reset_suppressed.store(true, Ordering::Relaxed);
+                    // Not fidelity: WA Web (Handle/StreamError.js) special-cases
+                    // only 500..600, so 429 is indistinguishable from any other
+                    // reconnect there — survivable because a human watches the UI.
+                    // An embedder has none, and `StreamError` is the channel every
+                    // other coded branch already reports through. Dispatched after
+                    // the stores so a handler sees the rate-limited session.
+                    self.core.event_bus.dispatch(Event::StreamError(
+                        crate::types::events::StreamError::builder()
+                            .code(code.to_string())
+                            .raw(node.to_owned())
+                            .build(),
+                    ));
                 }
                 "503" => {
                     // Server is going down/restarting: mark logged-out so sends fail
--- a/src/client/node_io.rs
+++ b/src/client/node_io.rs
@@ -1908,9 +1908,9 @@ impl Client {
                     // Not fidelity: WA Web (Handle/StreamError.js) special-cases
                     // only 500..600, so 429 is indistinguishable from any other
                     // reconnect there — survivable because a human watches the UI.
-                    // An embedder has none, and `StreamError` is the channel every
-                    // other coded branch already reports through. Dispatched after
-                    // the stores so a handler sees the rate-limited session.
+                    // An embedder has none, so report the rate limit through
+                    // `StreamError`. Dispatched after the stores so a handler
+                    // sees the rate-limited session.
                     self.core.event_bus.dispatch(Event::StreamError(
                         crate::types::events::StreamError::builder()
                             .code(code.to_string())
--- a/src/client/tests.rs
+++ b/src/client/tests.rs
@@ -3260,6 +3260,95 @@ async fn test_stream_error_429_keeps_reconnect_with_backoff() {
     );
 }
 
+/// A rate-limited session parks the client for minutes; without an event the
+/// only trace is the missing connection. WA Web has no 429 arm to copy here
+/// (only 500..600 is special-cased), so this is our own `StreamError` contract
+/// applied consistently, and the neighbours must keep their own events.
+#[tokio::test]
+async fn test_stream_error_429_dispatches_stream_error_event() {
+    use wacore::types::events::{Event, EventHandler};
+
+    let client = create_offline_sync_test_client().await;
+    client.is_logged_in.store(true, Ordering::Relaxed);
+    let collector = Arc::new(crate::test_utils::TestEventCollector::default());
+    client
+        .subscribe_handler(collector.clone() as Arc<dyn EventHandler>)
+        .detach();
+
+    let node = NodeBuilder::new("stream:error")
+        .attr("code", "429")
+        .children([NodeBuilder::new("text")
+            .attr("text", "rate-overlimit")
+            .build()])
+        .build();
+    client.handle_stream_error(&node.as_node_ref()).await;
+
+    let events = collector.events();
+    let stream_errors: Vec<_> = events
+        .iter()
+        .filter_map(|event| match &**event {
+            Event::StreamError(stream_error) => Some(stream_error),
+            _ => None,
+        })
+        .collect();
+    assert_eq!(
+        stream_errors.len(),
+        1,
+        "429 must dispatch exactly one StreamError, got {events:?}"
+    );
+    assert_eq!(stream_errors[0].code, "429");
+    let raw = stream_errors[0]
+        .raw
+        .as_ref()
+        .expect("the stanza must ride along so the reason survives");
+    assert_eq!(raw.tag, "stream:error");
+    assert!(
+        raw.get_optional_child("text").is_some(),
+        "the raw stanza must keep the server's children, not just the code"
+    );
+    assert!(
+        !events
+            .iter()
+            .any(|event| matches!(**event, Event::LoggedOut(_) | Event::StreamReplaced(_))),
+        "429 is not a logout or a replacement"
+    );
+}
+
+/// The branches either side of 429 keep dispatching what they always did — a
+/// regression here would look like the 429 event working while 516/409 lost
+/// theirs.
+#[tokio::test]
+async fn test_stream_error_neighbours_keep_their_events() {
+    use wacore::types::events::{Event, EventHandler};
+
+    for (code, expect_logged_out) in [("516", true), ("401", true), ("409", false)] {
+        let client = create_offline_sync_test_client().await;
+        let collector = Arc::new(crate::test_utils::TestEventCollector::default());
+        client
+            .subscribe_handler(collector.clone() as Arc<dyn EventHandler>)
+            .detach();
+
+        let node = NodeBuilder::new("stream:error").attr("code", code).build();
+        client.handle_stream_error(&node.as_node_ref()).await;
+
+        let events = collector.events();
+        let matched = events.iter().any(|event| {
+            if expect_logged_out {
+                matches!(**event, Event::LoggedOut(_))
+            } else {
+                matches!(**event, Event::StreamReplaced(_))
+            }
+        });
+        assert!(matched, "{code} lost its event, got {events:?}");
+        assert!(
+            !events
+                .iter()
+                .any(|event| matches!(**event, Event::StreamError(_))),
+            "{code} must not also report as a generic StreamError"
+        );
+    }
+}
+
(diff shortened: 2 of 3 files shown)


jlucaso1:
Both review points were valid; fixed in `ea9a650`.

- **Fibonacci overflow** — real. Starting from `(1, 2)`, `delay_a + delay_b` exceeds `u64::MAX` at iteration 90, and since the loop is unbounded that is reachable (roughly 15 hours of continuous failure without a disconnect). Clamped with `saturating_add(...).min(MAX_DELAY_SECS)`, matching `fibonacci_backoff` in `client.rs`. Past the cap the state only exists to overflow, so the retry schedule is unchanged.
- **Comment accuracy** — also right. The neighbouring arms dispatch specialized events (`LoggedOut`, `StreamReplaced`), not `StreamError`, so there is no general convention to appeal to. Dropped that clause and kept only the reason 429 needs an event.

On the failing `Semver Checks (informational)` job: it runs `-p wacore -p wacore-binary -p waproto`, none of which this PR touches — the diff is confined to `src/` of `whatsapp-rust`. The reported breakages (`struct_missing`, `module_missing`, `enum_variant_added` on `BinaryError::UnexpectedFormatByte`, …) all predate this branch.

---

PR #1264, opened 2026-08-10
fix(client): stop the run loop announcing a reconnect that disconnect() forbids

jlucaso1:
## What happens

A caller drives a session with `Client::run`, then calls `disconnect()` followed by `connect()`. The engine log, at `debug`, in order:

```
INFO  Disconnecting client intentionally.
DEBUG Message loop exited gracefully (expected disconnect).
INFO  Expected disconnect (e.g., 515), reconnecting immediately...
INFO  Client run loop has shut down.
DEBUG Connecting WebSocket and fetching latest client version in parallel
DEBUG [socket] resumeNoiseHandshake started
INFO  Handshake complete (IK), switching to encrypted communication
DEBUG <-- Received WebSocket data: 5025 bytes
DEBUG <-- Received WebSocket data: 334 bytes        (four more frames)
      ... 18 seconds of silence ...
DEBUG Keepalive skipped, connection already closing: NotConnected
DEBUG Fatal keepalive failure, exiting loop.
```

The loop announces an immediate reconnect and, on the next line, announces that it has shut down. The later `connect()` completes a handshake and ~6.5 KB arrives; nothing decodes it, no `Event::Connected` is dispatched, and eighteen seconds later the keepalive finds `NotConnected`.

The tail of that log is from `f8165f28` (0.7.0), where `connect()` returned `Result<(), _>` and drove the read itself. On current `main` the same call is refused: `connect()` returns `ConnectError::Shutdown` once the shutdown notifier has fired (`src/client/lifecycle.rs:720-725`, added by #1258), and `connecting_is_refused_after_the_client_is_shut_down` already pins that. What survives on `main` is the first three lines.

## Which reading

**(A) `disconnect()` is terminal by design.** The defect is the log line, and the missing capability is a separate thing.

`disconnect()` has no doc comment, so the intent had to come from the code around it. Several places say the same thing:

- `connect()` refuses after the shutdown notifier fires, and says why: *"a shutdown is published once and for good, and a connection established after it would be one the application has already been told does not exist. **A new client is the way back.**"* `disconnect()` fires that notifier (`src/client/lifecycle.rs:1023`).
- `is_terminal()`'s doc names the callers that make a client terminal: *"it is fired by `disconnect`, `logout` and `signal_shutdown_sync`"*.
- `logout()` ends with `self.disconnect().await` — and `logout` is unambiguously the end of a session.
- `Drop for Client` calls `signal_shutdown_sync`, documented as *"Synchronous flag-only equivalent of the first lines of `disconnect()`"*. `disconnect()` is the path `Drop` approximates.
- The contrast is already written down on the other side: `reconnect()`'s doc opens with *"Unlike `disconnect`, this does **not** stop the run loop."*

So `disconnect()` setting `is_running = false` — the run loop's own `while` condition — is deliberate. The bug is that the loop's post-connection branch reads `expected_disconnect` alone:

```rust
if self.expected_disconnect.load(Ordering::Relaxed) {
    …
    info!("Expected disconnect (e.g., 515), reconnecting immediately...");
    continue;
}
```

That flag is set by the post-pairing 515 **and** by `disconnect()`, `logout()` and `signal_shutdown_sync()`. The last three also clear `is_running`, so the `continue` falls out of the `while` immediately — the loop announces a reconnect it has already been forbidden from making, and that line is the only account a reader gets of a session that never came back. (Only plain `disconnect()` shows it: `logout()` clears `enable_auto_reconnect`, so it exits one branch earlier under an accurate message.)

The fix reads the loop's own stop condition alongside the flag. Behaviour is unchanged — the loop exited on `is_running` before and exits on it now — so what differs is only which line is logged. `disconnect()` also gets the doc comment whose absence made this an investigation.

## Ruled out

The three hypotheses raised while triaging this, so nobody repeats them:

- **Not a stale `expected_disconnect`.** In `f8165f28` `connect_graph` did not clear the flag and #1258 made it clear it. Against the #1258 commit the symptom is unchanged.
- **Not a dropped `Connection` guard.** That applies to the post-#1258 API; in `f8165f28` `connect()` returns `Result<(), _>` and drives the read itself. Adapting the caller to drive the guard on the new API leaves the symptom in place.
- **Not a race between `run` and the caller.** Removing the caller's `connect()` and leaving only `disconnect()` stops the log at *"reconnecting immediately"* followed by *"Client run loop has shut down"*, with no new connection at all.

## What is still missing

There is no way to release a connection and pick the session back up later. The library offers the two ends of the range and nothing between them:

- `disconnect()` / `logout()` — terminal, no way back for that `Client`.
- `reconnect()` / `reconnect_immediately()` — drop the socket and come straight back, on the library's schedule, not the caller's.

A caller that wants to go offline and stay offline until it says otherwise has neither. whatsmeow's `Disconnect()` + `ConnectContext()` and Baileys' relaunch path both cover that case; this library's lifecycle model does not, and closing the gap is a design decision — detach/re-attach semantics, what happens to the run loop, what a re-attach guarantees about generation and app state — not a fix to fold into this PR. Not opened as an issue here; happy to file one if that shape is wanted.

## Validation

```
cargo fmt --all
cargo test -p whatsapp-rust --lib                                  # 1586 passed, 0 failed
cargo clippy -p whatsapp-rust --lib --tests -- -D warnings         # clean
cargo test --doc -p whatsapp-rust -p wacore                        # clean
RUSTDOCFLAGS="-D warnings" cargo doc -p whatsapp-rust --no-deps --all-features
```

`cargo clippy --workspace --all-targets` does not build in this container: `alsa-sys` fails in `build.rs` for want of a system `alsa.pc`, before any of this code is linted. Unrelated to the diff — leaving the workspace matrix to CI.

Reverting only the four-line guard and re-running makes `a_requested_disconnect_is_not_announced_as_a_reconnect` fail on the announcement, with the other two still passing.

### Tests

- `a_requested_disconnect_is_not_announced_as_a_reconnect` — the regression. A transport factory that parks inside `create_transport()` holds the run loop in a connect attempt, so `disconnect()` lands in exactly the window that produced the misreport instead of racing for it. Asserts the loop announced no reconnect and did say why it stopped.
- `a_protocol_reconnect_is_still_announced_and_still_made` — the positive control, so the fix cannot over-reach: an `expected_disconnect` set while the loop is still running is still announced *and* a second connect attempt follows.
- `a_disconnected_client_cannot_be_run_again` — proves reading (A) as state, so the next person need not re-derive it: after `disconnect()` the client is terminal and `run()` refuses at once.

The `log`-capture machinery from `message::tests` moves to `test_utils::log_capture`, since `log`'s global logger is install-once and two suites cannot each own it. It is now one process-wide sink with a target allow-list, and records are readable only through a `Session` guard that serialises asserting tests and clears the buffer — so "nothing announced a reconnect" is a statement about the test's own client.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gz8iBHYzvBmH32KjtSx2Zn)_

### DIFF
--- a/src/client/lifecycle.rs
+++ b/src/client/lifecycle.rs
@@ -46,8 +46,7 @@ impl Client {
     /// for `Drop` impls on FFI wrappers (e.g. `WasmWhatsAppClient`) that
     /// can't run async cleanup synchronously.
     pub fn signal_shutdown_sync(&self) {
-        self.expected_disconnect.store(true, Ordering::Relaxed);
-        self.is_running.store(false, Ordering::Relaxed);
+        self.publish_terminal_verdict();
         self.shutdown_notifier.notify();
         self.notify_session_state();
         #[cfg(feature = "client-lifecycle")]
@@ -55,6 +54,21 @@ impl Client {
             lifecycle.signal_shutdown_sync();
         }
         self.notify_connection_shutdown();
+    }
+
+    /// Publish "the session is over, and no reconnect follows" as one
+    /// transition.
+    ///
+    /// The order and the `Release` are the whole point, and pair with the
+    /// `Acquire` load in [`run`](Self::run)'s expected-disconnect branch: that
+    /// branch reads `expected_disconnect` first and `is_running` second, so the
+    /// reader flag has to be cleared *before* the verdict is published. Stored
+    /// the other way round — or both `Relaxed`, as they were — the branch can
+    /// read a torn pair, see a planned end with a reader still live, and
+    /// announce a reconnect that this call has already ruled out.
+    fn publish_terminal_verdict(&self) {
+        self.is_running.store(false, Ordering::Relaxed);
+        self.expected_disconnect.store(true, Ordering::Release);
     }
 
     pub(crate) fn connection_shutdown_signal(&self) -> wacore::runtime::ShutdownSignal {
@@ -624,8 +638,24 @@ impl Client {
                 break;
             }
 
-            // If this was an expected disconnect (e.g., 515 after pairing), reconnect immediately
-            if self.expected_disconnect.load(Ordering::Relaxed) {
+            // If this was an expected disconnect (e.g., 515 after pairing), reconnect immediately.
+            //
+            // Acquire, paired with the Release in `disconnect`/`signal_shutdown_sync`:
+            // those clear `is_running` before publishing this flag, so seeing it
+            // set here means their cleared reader flag is visible to the Relaxed
+            // load below. Without the pairing the two stores can be read torn —
+            // planned end, reader still live — which is the misreport again, on
+            // a connection that happened to end of its own accord at the same
+            // moment the application asked for the client to stop.
+            if self.expected_disconnect.load(Ordering::Acquire) {
+                // A planned end owes no backoff whichever way it goes, so the
+                // per-connection reset happens before that is decided: a terminal
+                // exit must leave the same `reconnect_errors` a reconnecting one
+                // does, or this branch changes what `stats()` reports.
+                self.auto_reconnect_errors.store(0, Ordering::Relaxed);
+                // Consume the auth timestamp so a later failed connect can't
+                // read this cycle's stale value as a "stable" connection.
+                self.connected_at_ms.store(0, Ordering::Relaxed);
                 // The flag only says the end was planned, not that another
                 // connection follows: `disconnect()`, `logout()` and
                 // `signal_shutdown_sync()` set it too, and they also clear
@@ -637,10 +667,6 @@ impl Client {
                     info!("Disconnect requested, shutting down without reconnecting.");
                     break;
                 }
-                self.auto_reconnect_errors.store(0, Ordering::Relaxed);
-                // Consume the auth timestamp so a later failed connect can't
-                // read this cycle's stale value as a "stable" connection.
-                self.connected_at_ms.store(0, Ordering::Relaxed);
                 info!("Expected disconnect (e.g., 515), reconnecting immediately...");
                 continue;
             }
@@ -1045,8 +1071,7 @@ impl Client {
     pub async fn disconnect(self: &Arc<Self>) {
         info!("Disconnecting client intentionally.");
         wacore::telemetry::set_connected(false);
-        self.expected_disconnect.store(true, Ordering::Relaxed);
-        self.is_running.store(false, Ordering::Relaxed);
+        self.publish_terminal_verdict();
         self.shutdown_notifier.notify();
         self.notify_session_state();
         #[cfg(feature = "client-lifecycle")]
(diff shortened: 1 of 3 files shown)


jlucaso1:
Three reviewers converged on the same two points, plus a third. All four are fixed in f61e003.

**The torn read of the terminal pair.** Right — `disconnect()` wrote `expected_disconnect` then `is_running`, both `Relaxed`, and the branch read them the other way round. It doesn't show up when a `disconnect()` closes the socket itself (the close is sequenced long after both stores), but a connection that ends of its own accord at the same instant lands the loop on the branch concurrently with the two stores, and the misreport is back.

Fixed as an ordered pair rather than `SeqCst` on everything: `publish_terminal_verdict()` clears `is_running` first and publishes `expected_disconnect` with a `Release`; the branch's load is now `Acquire`. Seeing the verdict therefore means seeing the cleared reader flag. `signal_shutdown_sync()` goes through the same helper, and `logout()` reaches it via `disconnect()`. Not testable deterministically, which is why it's stated at the one place the ordering is established.

**The skipped counter reset.** Correct, and it made the PR body's "behaviour is unchanged" claim false — `stats().reconnect_errors` is public. The reset and the auth-timestamp consumption now happen before either exit is chosen, and `a_requested_disconnect_still_clears_the_connection_counters` pins it.

**Log-capture isolation.** The session lock never excluded the tests that would actually pollute, since those never take it — that was overstated in the doc comment and now says what it really covers. Isolation is required instead of hoped for: `delegated_to_child` only lets a test assert in a process running that test alone (`NEXTEST=1`, or the child it forks), so under a threaded `cargo test --lib` no sibling can write into the buffer.

That also settles the observability point — the capture is never installed in a process running the whole suite, so it can no longer swallow other tests' logging. The `&&` ordering that guarantees it is now commented as load-bearing rather than left to look incidental.

Not changed: the "Semver Checks (informational)" failure. It's `continue-on-error: true`, checks `wacore`/`wacore-binary`/`waproto` against the last published release, and already fails on `main` at 5797257 — this PR touches none of those crates.

`cargo test -p whatsapp-rust --lib` 1587 passed, `clippy -p whatsapp-rust --lib --tests -D warnings` clean.

---

jlucaso1:
Two more, one taken and one declined. dceb678.

**The verdict the per-attempt reset erases** — taken, and it's the better finding of the two rounds. `run` clears `expected_disconnect` at the top of every attempt, so judging the stop on that flag only works when the `disconnect()` lands *after* the reset. Land it before, anywhere in a window as wide as the 20s connect timeout, and the branch never fires at all: the loop charges an attempt to `reconnect_errors` and announces "Will attempt to reconnect in ..." for a reconnect `connect()` is already refusing. Same untrue line, one branch further down, and the previous fix didn't reach it.

The question now comes from the two signals a reset can't touch — the shutdown latch, which the three session-ending calls set once and nothing clears, and `is_running` — with the ordered verdict pair kept beside them for the narrower window where the stores have landed but the latch's notify hasn't. `a_disconnect_the_attempt_reset_erased_still_stops_the_loop` pins it; it fails on the previous commit.

**Resetting the counters inside `publish_terminal_verdict`** — declined. A `disconnect()` during backoff leaving `reconnect_errors` non-zero is how `main` behaves today; this PR doesn't touch that path. The reset I *did* add exists precisely to keep the branch from changing what `stats()` reports, so moving it into the terminal transition would do the opposite — it would make this PR change an untouched path's public output. Whether a finished client should report the failures that finished it is a real question, and `main` currently answers it inconsistently (cleared on one exit, kept on the other), but settling that is a decision about what `reconnect_errors` means, not a fix to fold in here.

Both planned exits now share `clear_connection_backoff_state()`, so which branch runs can't change the number either way.

1588 lib tests pass; clippy clean on the touched crate.

---

PR #1265, opened 2026-08-10
feat(client): let a session go offline and come back on the caller's word

jlucaso1:
Stacked on #1264 — base it there, review the top commits only. Merge #1264 first and this retargets to `main` on its own.

Closes the gap #1264's *What is still missing* recorded.

> **Note on the diff.** The first commit is the design; four follow-ups rework it in response to review. The description below is the design **as it now stands**, not as first written — the contract changed materially between them (`connect()` is now refused while paused; it originally was not). The per-round reasoning is in the PR comments.

## What was missing

The lifecycle offered the two ends of a range and nothing between them:

| | drops the connection | comes back | on whose schedule |
| --- | --- | --- | --- |
| `disconnect()` / `logout()` | yes | **never** — new `Client` required | — |
| `reconnect()` / `reconnect_immediately()` | yes | immediately | the library's |
| **`pause()` / `resume()`** | yes | when asked | **the caller's** |

An application that wants to drop its connection and stay offline until it says otherwise had neither option. whatsmeow's `Disconnect()` + `ConnectContext()` and Baileys' relaunch path both cover that case; this library did not.

## What it guarantees

`pause()` drops the connection and parks the supervision loop. `resume()` releases it and the loop reconnects at once.

- **No connection exists while paused, and none is opened — by anyone.** `connect()` is refused with `ConnectError::Paused` (non-terminal, unlike `Shutdown`), and an attempt already in flight is *retracted* rather than published. Attempts carry the `pause_generation` they started in, so one that spanned a pause is refused even if a `resume()` landed while it was still handshaking and left the flag reading `false` throughout.
- **The `run()` future keeps running.** That is the whole reason this wasn't expressible before: every existing way to stop connecting also ends that future. The loop parks at one point — top of the iteration, before the attempt — because a pause can land during a backoff as easily as during a live connection.
- **The client is not terminal.** It is between connections, on purpose. `is_terminal()` is untouched.
- **No backoff is owed for the pause**, and that does not depend on timing: the loop reads it off the connection that ended (`pause_teardown_pending`), not off whichever of `pause`/`resume` got there first. The reconnect backoff also watches a notifier that only these two fire, so neither waits out a delay of up to the 900s cap.
- **A paused client reads as unreachable** — `can_reach_server`, `wait_for_socket` and `wait_for_connected` all say so, because `pause()` publishes its flag before the read loop clears `is_connected`.
- **`resume()` never waits on the teardown.** It is sync and deliberately does not queue behind `pause()`: that teardown ends in an untimed socket close, and a resume held behind it would be at the mercy of an unresponsive transport.

## What it deliberately does not do

- **No `Event::Disconnected`** — the application ended this connection, so the teardown is not news, the same reasoning `reconnect()` applies. (`Event` has its own stability policy; nothing there is touched.) The suppression flag is set only when there is a connection to consume it, so it can't stand into the connection after the resume.
- **Not a protocol-level presence change.** The account stays registered; other devices see nothing.
- **Does not override `enable_auto_reconnect: false`.** A client whose application already asked it not to come back still stops when its connection ends. That ordering is load-bearing: `handle_stream_error` makes a client terminal (conflict, 516) through that same flag, and putting pause ahead of it would park a client the server has told to go away.
- **One qualification on "the socket is closed on return":** an attempt still inside its handshake holds its transport in that future, out of reach of the slot `pause()` reads; it closes at its own next refusal check, which can land just after. What holds unconditionally is the part a caller can act on — no such attempt ever becomes a published, readable connection. Making even that window impossible needs an attempt registry with cancellation, which is bigger than this PR should grow.

`await_connection` deliberately **keeps waiting** through a pause rather than giving up: nothing on the next connection re-issues a consumer's task, so returning early would drop a full-sync request instead of deferring it. Waiting is already what it does for a 900s backoff.

## Tests

- `a_paused_session_stops_connecting_and_resumes_on_request` — the capability. Asserts no connection is opened while parked (the loop's own notifier-listener count is the observable, not a sleep), that the client is not terminal, that `run()` is still driving it, and that `resume()` produces the next attempt.
- `connecting_is_refused_while_the_session_is_paused` / `a_pause_mid_attempt_retracts_the_connection_it_would_have_published` — the hard guarantee, including the half the run loop cannot cover.
- `resuming_does_not_queue_behind_a_stalled_teardown` — resume takes effect while the socket close hangs.
- `pausing_an_undriven_connection_still_clears_the_connected_flag` — `connect()` publishes `is_connected` before handing back the `Connection`, so pausing a directly-driven-but-unread connection has to run the cleanup itself or every later `connect()` is refused as `AlreadyConnected`.
- `pausing_announces_no_backoff_it_will_not_wait`, `disconnecting_a_paused_session_still_ends_the_run_loop`, `resuming_does_not_undo_a_shutdown`, `await_connection_tests::a_paused_client_keeps_the_wait_but_is_not_reachable`.

## Validation

```
cargo fmt --all
cargo test -p whatsapp-rust --lib                                  # 1597 passed, 0 failed
cargo clippy -p whatsapp-rust --lib --tests -- -D warnings         # clean
cargo test --doc -p whatsapp-rust -p wacore                        # clean
RUSTDOCFLAGS="-D warnings" cargo doc -p whatsapp-rust --no-deps --all-features
```

`cargo clippy --workspace --all-targets` does not build in this container (`alsa-sys` needs a system `alsa.pc`), so that matrix is CI's. "Semver Checks (informational)" is `continue-on-error` and already fails on `main` at 5797257 against `wacore`/`wacore-binary`/`waproto`, none of which this touches.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1282,6 +1282,12 @@ pub struct Client {
     pub(crate) undecryptable_dispatched: Cache<ChatMessageId, ()>,
 
     pub enable_auto_reconnect: Arc<AtomicBool>,
+    /// Set by [`Client::pause`] and cleared by [`Client::resume`]: the run loop
+    /// parks instead of connecting for as long as it holds.
+    ///
+    /// Deliberately not part of `is_terminal`: a paused client is between
+    /// connections, not finished, and the application means to come back.
+    pub(crate) paused: AtomicBool,
     /// Consecutive reconnect failures, drives the Fibonacci backoff. Exposed
     /// read-only via [`StatsSnapshot::reconnect_errors`](wacore::stats::StatsSnapshot).
     pub(crate) auto_reconnect_errors: Arc<AtomicU32>,
--- a/src/client.rs
+++ b/src/client.rs
@@ -1293,9 +1293,6 @@ pub struct Client {
     /// Deliberately not part of `is_terminal`: a paused client is between
     /// connections, not finished, and the application means to come back.
     pub(crate) paused: AtomicBool,
-    /// Serialises [`Client::pause`] against [`Client::resume`]. See
-    /// `pause_lifecycle` on those methods for what the ordering buys.
-    pub(crate) pause_lifecycle: Mutex<()>,
     /// Fired by [`Client::pause`] and [`Client::resume`], and by nothing else,
     /// so the run loop's reconnect backoff can watch it without the spurious
     /// wakes that would collapse the delay it exists to serve.
@@ -1305,6 +1302,11 @@ pub struct Client {
     /// re-read of `paused`, because a [`Client::resume`] can land between the
     /// two and the backoff a pause does not owe must not turn on that timing.
     pub(crate) pause_teardown_pending: AtomicBool,
+    /// Bumped by every [`Client::pause`]. A connection attempt reads it once at
+    /// the start and is refused if it has moved, so an attempt that spanned a
+    /// pause is never published — even when a [`Client::resume`] landed while it
+    /// was still handshaking and left the flag reading `false` throughout.
+    pub(crate) pause_generation: AtomicU64,
     /// Consecutive reconnect failures, drives the Fibonacci backoff. Exposed
     /// read-only via [`StatsSnapshot::reconnect_errors`](wacore::stats::StatsSnapshot).
     pub(crate) auto_reconnect_errors: Arc<AtomicU32>,
--- a/src/client.rs
+++ b/src/client.rs
@@ -1300,6 +1300,11 @@ pub struct Client {
     /// so the run loop's reconnect backoff can watch it without the spurious
     /// wakes that would collapse the delay it exists to serve.
     pub(crate) pause_state_notifier: Arc<event_listener::Event>,
+    /// Set by [`Client::pause`] for the connection it tears down, consumed by
+    /// the run loop's post-connection branch. A one-shot fact rather than a
+    /// re-read of `paused`, because a [`Client::resume`] can land between the
+    /// two and the backoff a pause does not owe must not turn on that timing.
+    pub(crate) pause_teardown_pending: AtomicBool,
     /// Consecutive reconnect failures, drives the Fibonacci backoff. Exposed
     /// read-only via [`StatsSnapshot::reconnect_errors`](wacore::stats::StatsSnapshot).
     pub(crate) auto_reconnect_errors: Arc<AtomicU32>,
--- a/src/client.rs
+++ b/src/client.rs
@@ -1307,6 +1307,13 @@ pub struct Client {
     /// pause is never published — even when a [`Client::resume`] landed while it
     /// was still handshaking and left the flag reading `false` throughout.
     pub(crate) pause_generation: AtomicU64,
+    /// Held across the connect graph's final refusal-check-and-publish and
+    /// across [`Client::pause`]'s capture of what it is tearing down, so a pause
+    /// cannot read "no connection" from an attempt one statement short of
+    /// publishing one. Deliberately covers flag and slot writes only — never
+    /// network I/O — so it cannot become the kind of wait that parks a caller
+    /// behind an unresponsive socket.
+    pub(crate) connection_publish: Mutex<()>,
     /// Consecutive reconnect failures, drives the Fibonacci backoff. Exposed
     /// read-only via [`StatsSnapshot::reconnect_errors`](wacore::stats::StatsSnapshot).
     pub(crate) auto_reconnect_errors: Arc<AtomicU32>,
--- a/src/client.rs
+++ b/src/client.rs
@@ -840,6 +840,10 @@ pub enum ConnectError {
     /// rather than reconnecting this one.
     #[error("client has been shut down")]
     Shutdown,
+    /// [`Client::pause`] is in effect. Unlike [`Self::Shutdown`] this is not
+    /// final: [`Client::resume`] lifts it and connecting works again.
+    #[error("client is paused")]
+    Paused,
     /// A step of the connect flow ran out of time.
     #[error("{stage} timed out after {timeout:?}")]
     Timeout {
@@ -904,6 +908,7 @@ impl ConnectError {
             ConnectError::AlreadyConnected
             | ConnectError::NotActivated
             | ConnectError::Shutdown
+            | ConnectError::Paused
             | ConnectError::Version(_)
             | ConnectError::Transport(_) => false,
         }
@@ -1288,6 +1293,13 @@ pub struct Client {
     /// Deliberately not part of `is_terminal`: a paused client is between
     /// connections, not finished, and the application means to come back.
     pub(crate) paused: AtomicBool,
+    /// Serialises [`Client::pause`] against [`Client::resume`]. See
+    /// `pause_lifecycle` on those methods for what the ordering buys.
+    pub(crate) pause_lifecycle: Mutex<()>,
+    /// Fired by [`Client::pause`] and [`Client::resume`], and by nothing else,
+    /// so the run loop's reconnect backoff can watch it without the spurious
+    /// wakes that would collapse the delay it exists to serve.
+    pub(crate) pause_state_notifier: Arc<event_listener::Event>,
     /// Consecutive reconnect failures, drives the Fibonacci backoff. Exposed
     /// read-only via [`StatsSnapshot::reconnect_errors`](wacore::stats::StatsSnapshot).
     pub(crate) auto_reconnect_errors: Arc<AtomicU32>,
--- a/src/client/app_state.rs
+++ b/src/client/app_state.rs
@@ -1071,6 +1071,13 @@ impl Client {
         if !self.is_running.load(Ordering::Relaxed) {
             return Some(false);
         }
+        // A pause is the same shape: not terminal, because the application means
+        // to come back, but the loop is parked and the next connection is due
+        // whenever it says so — which may be never. Waiting one out is waiting
+        // on an application decision, not on a network.
+        if self.is_paused() {
+            return Some(false);
+        }
         None
     }
 
@@ -4417,6 +4424,39 @@ mod await_connection_tests {
                 .expect("a finished client must end the wait, with nothing else nudging it")
                 .expect("the waiter should not panic"),
             "and it reports that none arrived"
+        );
+    }
+
+    /// A pause is an open-ended offline window the application controls, so the
+    /// wait has to end on it too. Not because the client is finished — it is
+    /// coming back — but because when it comes back is a decision nothing here
+    /// can wait for, and the parked task holds the `Arc<Client>` meanwhile.
+    #[tokio::test]
+    async fn a_paused_client_ends_the_wait() {
+        let client = crate::test_utils::create_test_client_with_name("await-paused").await;
+        client.is_running.store(true, Ordering::Relaxed);
+
+        let waiter = {
+            let client = Arc::clone(&client);
+            tokio::spawn(async move { client.await_connection().await })
+        };
+
+        crate::test_utils::poll_until("the waiter to park on the notifier", || {
+            client.socket_ready_notifier.total_listeners() >= 1
+        })
+        .await;
+
+        client.pause().await;
+        assert!(
+            !tokio::time::timeout(Duration::from_secs(5), waiter)
+                .await
+                .expect("a paused client must end the wait rather than hold it open")
+                .expect("the waiter should not panic"),
+            "and it reports that no connection arrived"
+        );
+        assert!(
+            !client.is_terminal(),
+            "which is not the same as the client being finished"
         );
     }
 
--- a/src/client/app_state.rs
+++ b/src/client/app_state.rs
@@ -1071,13 +1071,15 @@ impl Client {
         if !self.is_running.load(Ordering::Relaxed) {
             return Some(false);
         }
-        // A pause is the same shape: not terminal, because the application means
-        // to come back, but the loop is parked and the next connection is due
-        // whenever it says so — which may be never. Waiting one out is waiting
-        // on an application decision, not on a network.
-        if self.is_paused() {
-            return Some(false);
-        }
+        // A pause deliberately gets no branch of its own. It is not terminal —
+        // the application means to come back — and nothing on the next
+        // connection re-issues a consumer's task, so giving up here would drop
+        // a full-sync request outright rather than defer it. Waiting is already
+        // what this does for a 900s backoff, and a pause is the same shape:
+        // offline now, connected later, bounded by the client's lifetime. The
+        // window where the pause is still tearing down is handled where it
+        // belongs, in `can_reach_server`, so a paused client never reads as
+        // reachable while it waits.
         None
     }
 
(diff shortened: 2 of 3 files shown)


jlucaso1:
Both reviewers found the same four things, and they were four symptoms of one mistake: `paused` was a flag the run loop consulted *between* attempts. A flag checked at one point in a loop cannot bind an attempt already in flight, so it could never have made the guarantee the doc claimed. Reworked in 4740ef3.

**The P1 — pause silently failing.** Correct, and the worst of them. Pausing while `connect()` was fetching a version, opening a transport or handshaking left nothing published for the teardown to close; the attempt then went on to publish a live socket, and `is_paused()` stayed true over a session that was online indefinitely.

`connect()` now refuses with `ConnectError::Paused`, and the connect graph re-reads the pause at each point it already re-reads the shutdown, retracting a socket rather than publishing it. That is what turns "the loop won't start one" into "no connection exists". Two tests: `connecting_is_refused_while_the_session_is_paused` and `a_pause_mid_attempt_retracts_the_connection_it_would_have_published`.

This does contradict what I wrote in the original PR body — that `pause()` governs the supervision loop and would not refuse a `connect()` you make yourself. That position is untenable: the loop's *own* `connect()` is the one that has to be abortable. The stronger contract is also the simpler one, so the doc now says paused means paused, for everyone. `ConnectError` is `#[non_exhaustive]`, so the new variant is additive.

**The other three**, each fixed at the one place that answers for every caller rather than at each waiter:

- `can_reach_server` reads a pause as unreachable. `pause()` publishes the flag and only then flushes and closes, so the socket answered for that entire window and the answer was discarded with the connection. Fixing it there covers sends and IQs too, not just the app-state waiter.
- The reconnect backoff now watches a notifier that only `pause()` and `resume()` fire. Deliberately not `session_state_notifier` — every teardown fires that one, and watching it would collapse the backoff this loop exists to serve into hammering the server. That hazard is already called out in the existing comment beside the `select!`.
- `intentional_reconnect` is set only when there is a connection to consume it. Set with no reader running, it stood into the connection *after* the resume and would have swallowed that one's first genuine `Disconnected`.

**Where I went the other way — the app-state schedulers.** Codex was right that returning `false` from `await_connection` on a pause makes both retry loops take their "client is finished" path and drop consumer-issued full syncs outright; nothing on a later connection re-issues them. So I removed that branch rather than adding to it. `await_connection`'s contract is already "wait until terminal", and a pause is not terminal — it is offline-now-connected-later, the same shape as a 900s backoff it already waits out. The teardown window that motivated cubic's version of this is handled by the `can_reach_server` change instead, so a paused client never reads as reachable while it waits. `a_paused_client_keeps_the_wait_but_is_not_reachable` pins both halves.

**Not changed — pause vs. `enable_auto_reconnect: false`.** The auto-reconnect check still runs first, so a client whose application already asked it not to come back stops when its connection ends, pause or no pause. That ordering is load-bearing: `handle_stream_error` makes a client terminal (conflict, 516) through that same flag, and putting pause ahead of it would park a client the server has told to go away, then reconnect it on resume. An application setting both is contradicting itself, and stopping is the safe reading. Now stated in `pause()`'s doc rather than left to be inferred.

1596 lib tests pass; clippy, rustdoc and doctests clean.

---

jlucaso1:
Round three, a6dd67c. Three real, one that was a doc claim rather than a defect.

**Backoff charged for a pause when `resume()` was quick.** Right, and I missed that `intentional_reconnect` suppresses the *event*, not the *delay*. `pause()` returns once the socket is closed, which is before the read loop has finished unwinding, so a `resume()` in that window left the loop reading `paused` as already false and taking the ordinary backoff. The loop now reads a one-shot fact about the connection that ended — set by `pause()`, consumed in that branch — so the answer no longer turns on which of the two got there first.

**Listener registered after the state was read.** Correct, and it's the listen-then-recheck discipline used everywhere else in this file; I broke it by putting the `listen()` down at the `select!`. Moved above the read.

**`is_connected()` does not imply a reader.** The best of the three. `connect()` publishes `is_connected` before handing back the `Connection`, so a caller who connected directly and never drove it has no `drive_connection` behind it. Pausing there closed the socket but left the flag standing for good, and every later `connect()` was refused as `AlreadyConnected` — a pause with no way back, which is worse than the bug this PR is fixing. `pause()` now runs the cleanup itself when nothing is reading. `pausing_an_undriven_connection_still_clears_the_connected_flag` pins it, and it fails without the change.

(That test also caught me writing the wrong assertion first — I asserted one socket close, and there are two, because the cleanup closes it as well. That overlap is deliberate and pre-existing: transports make `disconnect()` idempotent precisely for it, and `Client::disconnect()` has the same shape.)

**In-flight attempt outliving `pause()`** — doc fix, not a mechanism fix, and I want to be explicit about the choice. An attempt inside its handshake holds its transport in that future, where `pause()`'s slot lookup cannot reach it; it closes at its own next refusal check, which can land just after `pause()` returns. Making that literally impossible needs `pause()` to await in-flight attempts — an attempt registry with cancellation, which is a bigger mechanism than this PR should grow. What holds unconditionally is the part a caller can act on: no such attempt ever becomes a published, readable connection, and it is retracted rather than driven. The doc now says that instead of claiming the socket is always closed on return.

1597 lib tests, clippy, rustdoc and doctests clean.

One process note: this is the third round on the same feature, and each round has been narrower than the last (silently-online → backoff-timing → doc precision). If a further pass turns up only windows of the "began one instruction after the check" kind, I'd rather record them as known limits in the doc than keep adding guards that cannot close them — but anything that leaves the client wrongly online, wrongly flagged, or wrongly delayed is still worth another round.

---

jlucaso1:
Round four, 9d60c2d. Three real, and the first is one I introduced myself last round.

**The mutex held a socket close against its own resume.** Greptile is right and this is the worst thing in the PR so far — worse than the ordering it was protecting. I held `pause_lifecycle` across `transport.disconnect()`, which is an untimed close; a transport that never answers would park `resume()` behind it indefinitely. "Come back on my word" becomes a session that never comes back.

The mutex is gone and `resume()` is sync again. What makes that safe is that the ordering no longer needs enforcing: `pause_teardown_pending` (added last round, for Codex's backoff finding) already settles the no-backoff question as a *fact about the connection that ended*, so it doesn't matter which of the two callers gets there first. The lock was belt on top of braces, and the belt could strangle you. `resuming_does_not_queue_behind_a_stalled_teardown` replaces the test that asserted the serialization, and asserts the property I actually wanted instead.

Worth naming the pattern: the mutex was the intuitive fix and it was wrong. Ordering two callers around an unbounded network wait is the thing to avoid; recording the fact each caller needs is the thing to do.

**An attempt that spans a completed pause.** Codex's sharpest one yet. If `pause()` *and* `resume()` both finish while `connect_graph` is awaiting its handshake, every level-triggered read of the flag says `false` and the attempt sails through — carrying an `outbound_flush` scope it reopened *before* the pause closed it. Nothing reopens it, so that connection would silently drop its delivery receipts, retries and transport acks for its whole life. A level-triggered flag simply cannot answer "did a pause happen while I wasn't looking".

Attempts now read a `pause_generation` once at the start and are refused if it has moved. An attempt that spanned a pause is not the connection the resume asked for; the next one is, and it reopens the scope normally on its way through.

**Readiness waiters answered for a paused session.** `pause()` publishes its flag before the read loop clears `is_connected`, so `wait_for_socket` and `wait_for_connected` would hand a caller the connection the application had just asked to have closed. Same treatment `can_reach_server` got last round.

1597 lib tests, clippy, rustdoc, doctests clean.

On the pattern across these four rounds: each has been a different way of asking "is this client supposed to be connected right now?", and the answer keeps needing to be a fact recorded at a point in time rather than a flag read later — the terminal verdict in #1264, then `pause_teardown_pending`, now `pause_generation`. If a fifth round finds another, I'd expect it to be the same shape, and I'd rather fix it the same way than add a fifth guard.

---

jlucaso1:
Valid, and worse than described. 6702279.

The teardown looked the transport slot up *after* its flushes. Those are bounded but not instant, and a `resume()` across them lets the run loop publish a replacement into that same slot — so the lookup returned the connection the resume had just asked for, and the pause closed it. A resumed session dying on arrival is the exact opposite of what resuming is for.

Greptile's sequence needs the old connection to end independently first, but it's reachable without that: **`pause()` on a client with no connection at all, resumed while its flushes run, is enough.** Nothing exotic has to happen.

Which socket a pause owns is now decided before its first await, where the answer is still unambiguous. The trailing cleanup for an unread connection needed the same treatment from the other side — it clears connection state wholesale, so it now runs only while this pause is still the standing one (same generation, still paused), or it would strip state belonging to the connection the resume asked for.

`a_pause_closes_the_socket_it_started_with_not_whatever_replaced_it` pins it. Holding an outbound-flush guard open stands in for the window, so the interleaving is a fixture rather than a race; with the lookup moved back after the flushes it fails by closing the replacement (`torn_down` closes: 0, expected 1).

This is the fifth round and the third instance of one pattern, which is worth naming now that it has repeated: **a value read after an await is a different question from the one asked before it.** The terminal verdict in #1264, `pause_teardown_pending`, `pause_generation`, and now the transport binding are all the same fix — decide at the point where the answer is unambiguous, carry it, rather than re-reading shared state later and hoping it still means what it meant. I've gone looking for other reads in `pause()` with that shape and the remaining ones (`is_connected` for `had_connection`, `is_running` for the cleanup gate) are either captured up front or guarded by the generation.

1598 lib tests, clippy, rustdoc, doctests clean.

---

jlucaso1:
All five valid. 63d8af2.

**The check-to-publication window.** The one that needed structure rather than another flag. `pause()` landing between the connect graph's final refusal check and its `is_connected` store read "no connection", captured no transport, and left the attempt to publish and flag a connection nothing would ever tear down. Those two are now one short section that `pause()`'s capture also takes.

Worth saying why a lock is right here when it was wrong two rounds ago: that one was held across an untimed socket close, so an unresponsive transport could park `resume()` indefinitely. This one covers flag and slot writes only — I moved the retraction's `orphan.disconnect().await` out of it — so there is nothing inside it that can fail to return. The hazard was never "a lock", it was "a lock around network I/O".

**The cleanup asked the wrong question.** Gating it on the pause still standing meant a `resume()` across the teardown skipped the only cleanup path that exists when nothing is reading — leaving a direct caller flagged connected for good, every later `connect()` refused as `AlreadyConnected`. A resume that cannot reconnect is worse than the bug I was guarding against. It now asks whether the socket it tore down is still the published one, which answers both directions at once: clean up what is mine, never what a replacement owns. `a_resume_across_the_teardown_does_not_strand_an_unread_connection` fails on the old gate.

**The three smaller ones**, each fixed where the decision is made rather than where the symptom shows:
- A retracted attempt owes no backoff. Only *established* connections carried the marker, so a pause and resume both landing mid-connect left the loop serving up to the 900s cap for a retraction the application asked for.
- `dispatch_connected` refuses while paused — a post-login task finishing inside the flush window passed its generation check and announced a session being taken offline.
- A directly driven reader consumes the pause marker itself, instead of leaving it for a `resume(); run()` to misread as a planned pause and skip a rate-limit backoff with.

1599 lib tests, clippy, rustdoc, doctests clean.

Sixth round, and I want to be straight about what that means rather than just reporting green again. Every round has found something real, so none of it was churn — but the rate is not falling the way it should, and all of it is one shape: new lifecycle state read by many concurrent actors, each read a chance to answer a question that has since changed. The fixes have converged on one discipline (decide where the answer is unambiguous, carry it, make the publication atomic with the check) and this round applied it in all five places, so I'd expect the next pass to be quieter. If it isn't — if round seven finds three more of the same kind — that is evidence the feature wants a connection-state machine rather than five coordinated flags, and I'd rather say so than keep patching.

Independent of that: this PR still has **no CI**, because its base is `fix/disconnect-then-connect` and the workflows trigger on `pull_request: branches: [main]`. Everything above is local runs plus review bots. Merging #1264 first and letting this retarget is what gets the real matrix — Miri, the feature matrix, MSRV and E2E — over a diff that has changed a great deal since it was opened.

---

jlucaso1:
## Stopping the patch-per-round loop — this needs a design decision, not a seventh fix

Round seven produced six findings. I've fixed one and I am **not** fixing the other four in this shape, because I said last round I'd say so if this happened, and it has.

**Fixed (3c15d79).** Greptile's P1, which is Codex's 1499 as well: `outbound_flush` and the per-connection shutdown notifier are shared and unversioned, so an old teardown closed and fired them against a replacement that had already reopened and reset them for itself — rejecting that connection's receipts, retries and acks for its whole life. Both are now gated on the teardown still owning the published connection, asked under the publish section and re-asked after the drain. `a_teardown_that_lost_the_connection_stops_fencing_it` fails without it. 1600 lib tests, clippy, rustdoc, doctests clean.

**Acknowledged and deliberately left open**, all valid as far as I can tell:

| | |
|---|---|
| 1471 | `pause_teardown_pending` survives when nothing consumes it (pause during backoff, or before `run()`), so a later outage skips a rate-limit backoff |
| 1524 | `is_running` cannot decide whether cleanup is needed — `resume(); run()` sets it, that `connect()` is refused `AlreadyConnected`, cleanup is skipped, connection stranded |
| 1543 | the `is_paused()` guards on readiness / reachability / `dispatch_connected` are cleared by `resume()` before the old generation retires, so the old socket can read ready in that window |
| 1464 | `intentional_reconnect` set for an undriven connection is never consumed by the manual cleanup path |

### Why I'm stopping rather than fixing them

**Two of these are caused by my own fixes from the previous two rounds.** 1471 exists because I made the marker unconditional to fix "a retracted attempt owes no backoff". 1543 exists because I added `is_paused()` guards to fix the readiness and `Connected` reports. Rounds one to four found independent bugs; rounds six and seven are increasingly finding consequences of the patches. That is the signal that the approach is wrong, not that the next patch is missing.

**The root cause is structural and now legible.** Pause/resume introduces a second, independent reason a connection can end, and it has to coordinate with connection state that is spread across at least eight unversioned pieces — `paused`, `pause_generation`, `pause_teardown_pending`, `intentional_reconnect`, `is_connected`, `is_running`, `outbound_flush`, `connection_shutdown` — each read at a different point, several of them after awaits, none of them carrying which connection they belong to. Every finding in the last three rounds is the same sentence: *this piece of shared state was read or written on behalf of a connection that is no longer the current one.* I have been answering it one piece at a time, and there are more pieces than rounds I should spend.

**What I think it wants instead:** one connection epoch that every piece of per-connection state hangs off, so "does this still belong to my connection?" is asked once against a value, not re-derived per flag. `connection_generation` already exists and is nearly that — it is bumped by `cleanup_connection_state` — but pause captures its transport before that bump, so it cannot use it today. Making the epoch advance at teardown *start* rather than at cleanup, and scoping the fencing operations to an epoch, would collapse all four remaining findings and the four I already fixed into one invariant.

That is a change to the connection lifecycle well beyond what a pause/resume feature should decide on its own, and it touches paths this PR otherwise leaves alone. So it is the maintainer's call, not mine to make inside a feature branch.

### Suggested way forward

1. **Merge #1264 on its own** — it is unaffected by all of this, green on the full matrix, and independently useful.
2. **Hold this PR** rather than merging it. In its current state pause/resume is correct for the ordinary sequences (pause, stay offline, resume) and has known gaps only where a resume overlaps an unfinished teardown. Whether that is acceptable depends on whether you expect callers to do that, which I can't judge for you.
3. Tell me which you want: the epoch restructure (I'd want it as its own PR, with this one rebased on top), the four remaining patches anyway (I'll do them, with the caveat that round eight likely finds their consequences), or this parked as an issue documenting the design work.

Worth adding: **this PR still has no CI** — its base is not `main`, so nothing here has seen Miri, the feature matrix, MSRV or E2E. That is a second reason not to merge it ahead of #1264.

---

jlucaso1:
Round eight. Valid, added to the open list, **not fixed** — and it is worth recording precisely why, because it is the strongest evidence yet for the previous comment.

**The finding.** The unread-connection cleanup checks `still_owns_connection()` and then calls `cleanup_connection_state()`. A replacement published between those two lines gets its transport disconnected and its generation retired by the old teardown.

Correct, and it is a direct consequence of the ownership check I added in round seven to fix a different instance of the same problem. Round seven's fix created round eight's finding, exactly as I said it would.

**What makes this the end of the line for local fixes.** The obvious repair is to hold `connection_publish` across both the check and the cleanup, the way I did for the fencing. That is not available here: `cleanup_connection_state()` awaits a transport close, a bounded inbound-commit settle, a chat-lane drain and a persistence flush. Holding a lock across it reintroduces exactly the hazard I removed in round five, where a pause held a lock across an untimed socket close and could park `resume()` behind a dead transport indefinitely.

So the local fix for round eight is blocked by the local fix from round five, and the finding itself was created by the local fix from round seven. Three of my own patches are now interfering with each other. That is not a sequence that converges by adding a fourth.

**Nothing changes about the recommendation.** The open list is now five:

| | |
|---|---|
| 1541 | cleanup TOCTOU — replacement published between the ownership check and the cleanup *(new)* |
| 1471 | `pause_teardown_pending` survives when nothing consumes it |
| 1524 | `is_running` cannot decide whether cleanup is needed |
| 1543 | `is_paused()` guards clear before the old generation retires |
| 1464 | `intentional_reconnect` unconsumed on the manual cleanup path |

All five are the same sentence — shared per-connection state acted on for a connection that is no longer current — and all five fall out of one epoch invariant, which is the restructure described above. An epoch also dissolves this one specifically: cleanup would be a no-op for a retired epoch by construction, so there would be nothing to check and therefore no window between the check and the act.

Still waiting on a decision between the epoch restructure, the remaining patches, and parking this as a design issue. #1264 remains unaffected and mergeable on its own.

---

jlucaso1:
Round nine, two more. Both valid, both the same class, neither fixed — the position in the two comments above stands and I won't restate it.

Appending to the open list:

| | |
|---|---|
| 1464 (revised) | `had_connection` mistakes a connected flag for a reader that can still consume the marker — the window after `drive_connection` swaps `intentional_reconnect` but before cleanup clears `is_connected` |
| 1545 | a retained undriven `Connection` handle survives the pause cleanup and can then drive the *replacement's* `transport_events` and Noise socket, tearing that session down |

1545 is the one worth pulling out, because it widens the redesign rather than just lengthening the list. `Connection` holds a client reference and nothing else — no generation, no epoch — so a handle from before a pause is indistinguishable from a current one and can drive state that is not its own. Any epoch invariant has to reach the public `Connection` type, not just the internal flags; that is a small addition to the API in this PR (`Connection` gains the epoch it was published under, and `read_until_disconnected` refuses a retired one), but it is a change to a type #1264 also touches, so it belongs in the design decision rather than in a patch on top.

Open list is now seven, all one sentence, all dissolved by the same invariant.

---

jlucaso1:
@codex review

PR #1269, opened 2026-08-10
feat(bot): accept a shared HTTP client, matching ClientBuilder

jlucaso1:
## Summary

`ClientBuilder` takes an HTTP client either way — `with_http_client` by value and `with_http_client_arc` for one that is already shared. `BotBuilder` had only the by-value setter. This adds the missing one, and documents on it what sharing an HTTP client actually implies.

To be precise about what that gap is, because my first draft of this description overstated it: for a concrete `Clone` client, sharing already worked. `UreqHttpClient` is `Clone` and cloning shares the `ureq::Agent` and therefore the pool, so `with_http_client(shared.clone())` shares today — `agent_docs/observability.md` documents exactly that. The setter earns its place in two narrower cases: a client that cannot be cloned, which by value moves into one builder and no further (`with_http_client` carries no `Clone` bound, so passing it to a *single* builder was always fine); and a client already erased to `Arc<dyn HttpClient>` because the host selects it at runtime, which cannot reach the by-value setter at all since nothing in the workspace implements `HttpClient` for `Arc<dyn HttpClient>`.

What is unambiguously real is the cost of the **default**, which is what a multi-bot process gets unless it opts out: every `BotBuilder::new()` calls `default_http_client()` → its own `UreqHttpClient::new()` → its own pool. `examples/multi_session_metrics.rs` builds two bots that way today; `tests/e2e/tests/process_footprint.rs` builds 16.

## Changes

- **`src/bot.rs`** — `with_http_client_arc(Arc<dyn HttpClient>) -> BotBuilder<B, T, Provided, R>`, mirroring `with_backend_arc` right above it; `with_http_client` now delegates to it exactly as `with_backend` delegates to `with_backend_arc`. The setter's docs carry the measured per-bot cost, which remedy is new versus already available, and the three things sharing implies (below).
- **`http_clients/ureq-client/src/lib.rs`** — tests only. `a_shared_client_runs_concurrent_requests_concurrently` pins the no-serialization claim against a gate server that answers nothing until all 64 requests have arrived, plus two tests for the gate's own deadline path. These live here rather than in `src/bot.rs` because it is `ureq::Agent`'s behavior being pinned, and `with_http_client_arc` is what turns that behavior into part of the contract.

No behavior change for existing callers: `by_value_http_clients_stay_independent` asserts two bots built with `with_http_client` still hold two distinct clients.

## Cost

Measured in release against a counting global allocator (live heap = allocated − deallocated), plain HTTP against a local keep-alive fixture. TLS costs more — this repo previously measured a pooled TLS connection at 88 KiB — so treat these as the floor.

**Retained bytes per `UreqHttpClient`:**

| | live heap | per client |
| --- | ---: | ---: |
| 50 clients, constructed only | 99,076 B | 1,981 B |
| 50 clients, one request each | 1,829,004 B | **36,580 B** |
| 1 shared client, 50 sequential requests | 36,501 B | — |
| 1 shared client, 50 concurrent requests | 85,140 B | — |

Retention, not churn: it survives the request, and dropping all 50 clients returned live heap to 3,853 B against a 3,003 B baseline. The 36.6 KB is the 16 KiB input + 16 KiB output buffers ureq allocates per pooled connection. Sharing collapses the fleet onto one pool capped by the agent rather than multiplied by session count.

Who pays, precisely: an idle session pays nothing. The version fetch sends `Connection: close` (#1243), so it pools nothing. The 36.6 KB starts once a session transfers media.

**Does sharing serialize? No.** This was the open question — a prior attempt at a hand-written delegating wrapper reportedly failed to bring a fleet up and blamed a single `ureq::Agent` serializing sessions. It does not reproduce. Against a server that answers no request until N have arrived, one shared `UreqHttpClient` held **64 requests in flight simultaneously**. `ureq::Agent` holds its pool lock across checkout only, never across the request. Each request still occupies one `spawn_blocking` thread — but so does each request through a client-per-bot, so sharing changes nothing there either way.

**What sharing does cost — 1: idle-pool contention.** The default agent retains 3 idle connections and 2 per host. 8 concurrent workers × 20 requests against a fixture with 5 ms latency:

| | TCP connections for 160 requests | reuse | wall clock |
| --- | ---: | ---: | ---: |
| one shared client (default agent) | 31–33 | 79–81% | 108 ms |
| one client per worker | 8 | 95% | 106 ms |
| one shared client, both caps at 8 | 8 | 95% | 107 ms |

Sharing at the default pool size roughly quadruples TCP connects — over TLS, handshakes — without costing wall clock here. The remedy needs no new API: `UreqHttpClient::with_agent` already accepts an agent sized for the intended concurrency. Both caps must be raised, not just the global one: a fleet's media requests converge on one CDN authority, so `max_idle_connections_per_host` binds first.

**What sharing does cost — 2: session isolation at the TLS layer.** Media URLs carry per-request auth and the client sends no cookies, so a shared connection carries nothing between sessions that a shared source IP does not already carry — except TLS session resumption, which lets a server link two sessions across a source-IP change. `agent_docs/observability.md` names this as the reason sharing stays opt-in; the setter's docs now say it too, so nobody opts in on the memory figure alone.

## Checked and not changed

- **"The buffers are per request, not per agent" (would have closed the whole thing).** Refuted by measurement: 36,580 B/client survives the request and is released only on drop. 50 clients retain 1.83 MB.
- **"`with_http_client_arc` is mechanically impossible under the typestate."** It is not. `cast()` re-tags every parameter without touching a field, so the `_arc` setter is the same one-line transition `with_backend_arc` already uses.
- **A pooled `HttpClient` (new impl, or reworking `UreqHttpClient`).** Not needed, on two counts. `UreqHttpClient` already has a pool — `max_idle_connections(3)`, `max_idle_connections_per_host(2)` — so "an agent without a pool" is a wrong premise. And the one thing a fleet wants changed, larger caps, is reachable today via `with_agent`: 79% → 95% reuse, identical to a client per worker. A larger hard-coded default would be wrong anyway, since it trades idle memory for reuse and only the integrator knows their concurrency.
- **"No consumer in this repo builds more than one bot."** Not the case: `examples/multi_session_metrics.rs` (2) and `tests/e2e/tests/process_footprint.rs` (16).

## Validation

- `cargo fmt --all`
- `cargo clippy -p whatsapp-rust -p whatsapp-rust-ureq-http-client --all-targets -- -D warnings` — clean
- `cargo test --workspace --lib` — 4,421 passed, 0 failed
- `cargo test --doc` on both touched crates — 17 passed
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features` on both touched crates — clean
- New tests: `bot::tests::shared_http_client_is_one_client_across_bots` (two bots, one `Arc`, `Arc::ptr_eq` against each other and against the caller's handle), `bot::tests::by_value_http_clients_stay_independent`, and in the ureq crate `a_shared_client_runs_concurrent_requests_concurrently`, `the_gate_gives_up_instead_of_waiting_forever`, `the_gate_releases_once_its_peers_arrive`

The measurement harnesses were scratch examples in the ureq crate, removed before commit; the concurrency one survives as a test.

`Semver Checks (informational)` is red on this PR and is not from it — it checks `-p wacore -p wacore-binary -p waproto`, none of which this diff touches, and its findings are in `wacore/src/iq/mex_operations.rs`, `wacore/binary/src/error.rs`, and a removed `simd` feature. The job is `continue-on-error: true` by design ("Advisory only … not to block the PR").

Branch note: pushed to `claude/botbuilder-shared-http-client-u5m0c0` rather than `feat/bot-builder-shared-http-client` — this session is pinned to that branch.

### DIFF
--- a/http_clients/ureq-client/src/lib.rs
+++ b/http_clients/ureq-client/src/lib.rs
@@ -355,6 +355,7 @@ mod tests {
     use std::io::{Read, Write};
     use std::net::TcpListener;
     use std::thread;
+    use std::time::Duration;
 
     fn spawn_fixed_size_server(body_size: usize) -> String {
         let listener = TcpListener::bind("127.0.0.1:0").expect("bind ephemeral port");
@@ -527,7 +528,7 @@ mod tests {
         assert_eq!(resp.status_code, 200);
 
         let (headers, body) = rx
-            .recv_timeout(std::time::Duration::from_secs(5))
+            .recv_timeout(Duration::from_secs(5))
             .expect("server should capture the request");
         assert_eq!(
             parsed_content_length(&headers),
@@ -564,7 +565,7 @@ mod tests {
         assert_eq!(resp.status_code, 200);
 
         let (headers, body) = rx
-            .recv_timeout(std::time::Duration::from_secs(10))
+            .recv_timeout(Duration::from_secs(10))
             .expect("server should capture the request");
         assert_eq!(parsed_content_length(&headers), Some(payload.len()));
         assert_eq!(body, payload);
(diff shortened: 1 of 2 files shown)


jlucaso1:
Review round addressed in 710115c. Three of the findings were valid and one of them corrected the PR's own framing, so the description above has been rewritten rather than amended.

**The framing was wrong.** Codex pointed at `agent_docs/observability.md`, which already documents sharing via `UreqHttpClient`'s `Clone` — cloning shares the `ureq::Agent` and therefore the pool, so `with_http_client(shared.clone())` shared before this PR. My original "cannot share without a delegating wrapper" was false for the default client. What genuinely has no route is a non-`Clone` client, or one erased to `Arc<dyn HttpClient>` by a host that picks it at runtime: nothing implements `HttpClient` for `Arc<dyn HttpClient>`, so it cannot reach the by-value setter at all. The measured per-bot cost is unaffected — it is what the *default* does, one `UreqHttpClient::new()` per builder — but the docs now distinguish the new remedy from the existing one.

**TLS linkage now documented.** Sharing costs session isolation: media URLs carry per-request auth and no cookies are sent, but TLS session resumption lets a server link two sessions across a source-IP change. The observability doc names that as the reason sharing is opt-in; the setter's docs say it too now, so nobody opts in on the memory figure alone.

**Per-host cap.** The remedy named only `max_idle_connections`. A fleet's media requests converge on one CDN authority, so `max_idle_connections_per_host` (2 by default) binds first. The 95%-reuse figure was measured with both raised; the doc now says both.

**Barrier hang** (raised independently by Greptile, CodeRabbit and Codex — and worse than stated, since nextest's default *and* ci profiles warn on slow tests without terminating them, so it would have sat on the job with no verdict). Replaced with a deadline-bounded gate. It differs from the suggested diff in two ways, both deliberate: the release is sticky, because a serialized client arrives one request at a time and otherwise each straggler serves its own full deadline; and the failure is an explicit assertion on a starved flag rather than a 503, so the message names the cause instead of surfacing as a status mismatch. Two tests cover the gate's own timeout path, since the concurrency test only ever takes the happy path. `N` also raised 8 → 64 to match the figure the setter's docs quote.

`Semver Checks (informational)` is red and is not from this PR: it checks `-p wacore -p wacore-binary -p waproto`, none of which this diff touches, and its findings sit in `wacore/src/iq/mex_operations.rs`, `wacore/binary/src/error.rs`, and a removed `simd` feature. The job is `continue-on-error: true` by design. Every other check was green on the previous commit.

---

PR #1274, opened 2026-08-10
perf(sqlite): let sibling devices share one connection pool

jlucaso1:
## Summary

Every `SqliteStore` constructor (`new`, `new_for_device`, `with_config`, `with_config_for_device`) builds its own r2d2 pool. A process holding N sessions against the **same** database file calls one of them per session and ends up with N SQLite connections.

A connection costs memory before it reads a row:

- **lookaside** — `SQLITE_DEFAULT_LOOKASIDE` is `1200,40` = **48,000 B** (46.9 KiB). Confirmed for *this* build rather than quoted from upstream: `PRAGMA compile_options` on the bundled SQLite 3.51.3 does **not** list `DEFAULT_LOOKASIDE`, and in the amalgamation the `ctime.c` table (L22738) is compiled before `sqliteLimit.h` defines it (L23636) — so its absence proves no `-D` override, and the header default applies.
- **page cache** — `SqliteStoreConfig::cache_size_kib`, 512 by default. A cap on growth, not a reservation.

`shared.rs` already names this problem for sibling *crates* ("two pools mean two WAL writers fighting over the file lock, two page caches, and two busy queues"). Sibling *devices* in one process had no equivalent. Everything else in the store is already device-scoped — every method is `*_for_device(&self, …, device_id)` — so only that field ever needed to vary.

## Changes

- `SqliteStore::share_for_device(&self, device_id) -> SqliteStore` — clones `pool`, `db_semaphore`, `reads`, `snapshot_safe` and `database_path`, swaps only `device_id`. The returned store owns the pool handles, so it cannot outlive them: dropping the store it came from closes nothing. The doc says why (per-connection cost) and what it does not do (does not create the device row; does not isolate writes, with the measured throughput cost; does not split the resource report).
- `examples/per_connection_memory.rs` — the harness that produced every number below. RSS-based, because SQLite's page cache and lookaside are `sqlite3_malloc` allocations from the bundled C library that a Rust `GlobalAlloc` wrapper never sees.
- Doc-only: `cache_size_kib` now records that it is a cap and not the floor; `SqliteStoreConfig` points at the new method.

No default changed. No new dependency, no `unsafe`, no global state, nothing outside `storages/sqlite-storage/`.

## Cost

**Bytes per connection vs per handle.** Resident-set delta, release build, 50 sessions that have each done two point reads (an idle session's steady state). The **marginal** column is the growth from the first session to the last divided by the sessions in between — it is the number that describes the batch, and unlike a total-over-N average it cannot be flattered by allocator memory the seeding connection left behind for the first session to reuse:

| `cache_size_kib` | pool per session (marginal) | shared pool (marginal) |
| --- | --- | --- |
| 512 (default) | 122.6 KiB | **0.0 KiB** |
| 64 | 122.0 KiB | 0.5 KiB |
| 16 | 104.3 KiB | 0.0 KiB |
| 1 | 90.5 KiB | 0.0 KiB |

Scaling at the default cap — the handle side stays at one connection's worth of RSS in total (163,840 B) no matter how many sessions hang off it:

| sessions | pool per session (marginal) | shared pool (marginal) | shared pool (total RSS) |
| --- | --- | --- | --- |
| 10 | 104.4 KiB | 0.0 KiB | 163,840 B |
| 50 | 122.6 KiB | 0.0 KiB | 163,840 B |
| 200 | 128.7 KiB | 0.1 KiB | 188,416 B |

N=1 does not regress: the base store *is* session 1, so no spare connection is opened ahead of the fleet.

**The pragma result, reported even though the API went ahead.** Lowering `cache_size_kib` was measured first, because a better default would have been a better outcome than new API. It caps growth but not the floor: 512 → 64 changed nothing for an idle session (122.6 → 122.0 KiB marginal), and even 512 → 1 only reached 90.5 KiB. So the whole knob is worth ~25% of the per-connection cost at its most extreme setting, against ~100% for sharing. Under a worst case that *does* fill the cap (every session full-scans a 4 MB table) it matters more — 631.3 → 150.9 KiB/session marginal for 512 → 64 — but that is the cap doing its job and buying hit rate, so the default stays where it is.

**How this squares with #1235.** That audit put the SQLite profile at *"~530 KiB + ~512 KiB"* marginal RSS per session, which is four times the 122.6 KiB here — and both are right, because they measure sessions in different states. #1235's sessions are connected and doing real work, so each one's page cache fills toward the 512 KiB cap; these are idle, two point reads in. The warm mode of this harness closes the gap and corroborates it: make every session scan a 4 MB table and the figure goes to 631.3 KiB/session, just past #1235's. Which cuts the same way — a session that actually uses its cache saves *more* by sharing a connection, not less. #1235's other finding also explains why this harness reads RSS rather than a Rust allocator hook: *"SQLite's page cache allocates through the amalgamation's own `malloc`, not the Rust global allocator. My counting allocator saw exactly 0 bytes."*

**Concurrent write latency, both modes.** Release, file-backed WAL, every session writing 200 records continuously, each arm against its own fresh database:

| sessions | shared pool total | pool per session total | shared spread (fastest→slowest) | per-pool spread |
| --- | --- | --- | --- | --- |
| 8 | 190 ms | 98 ms | 190 → 190 ms | 52 → 97 ms |
| 32 | 799 ms | 323 ms | 795 → 799 ms | 201 → 320 ms |
| 64 | 1.72 s | 681 ms | 1713 → 1718 ms | 367 → 677 ms |

So sharing costs ~2.5x aggregate write throughput (~7.5k vs ~20k writes/s, flat in N). The cause is not lock contention but pipelining: a private connection lets one session's queueing and `spawn_blocking` dispatch overlap another's SQLite work, which is the same effect `read_erased`'s comment already records. In exchange the shared queue is FIFO-fair — every session finishes within 5 ms of every other — where separate connections leave fairness to SQLite's busy handler and its random backoff, measured at ~2x spread.

`read_pool_size` does not change this: 0 vs 4 gave 190/193 ms shared and 98/89 ms separate at 8 sessions. A pure-write burst never touches reader connections, so the read side cannot buy back write serialization.

**That is why this is opt-in and not a default.** It is for mostly-idle fleets, which is the shape sessions actually have; a continuously-writing fleet should keep a store per session.

**Caveat carried over from the external sweep** that motivated this batch (harness not in this repo, not reproduced here): its −31.0% figure came from *idle* sessions and does not exercise concurrent writes at all. That comparison was also not ABBA-paired — two sweeps at different times, with an unmodified client interleaved through both reading 0.334 → 0.324 MiB/session, i.e. −2.9% drift against a −31.0% effect. Enough for an effect that size, not enough for one of a few percent. The numbers in this PR are in-process and independent of it.

## Checked and not changed

- **The pragma alternative** — numbers above. `cache_size_kib` reaches at most ~25% of the per-connection cost, and only by setting a cap so low it defeats the cache. Left at 512.
- **Reducing lookaside per connection** — not possible here. `SQLITE_DBCONFIG_LOOKASIDE` is C-API only (no pragma), and `connection_init` receives a `&mut SqliteConnection` whose `internal_connection: NonNull<ffi::sqlite3>` is `pub(super)` in diesel, so the handle cannot be reached at all — let alone without `unsafe`. `sqlite3_config(SQLITE_CONFIG_LOOKASIDE, …)` is process-global and start-time only, which the constraints rule out twice over. The 46.9 KiB is fixed.
- **`SharedSqlite`** — solves the adjacent problem (sibling *crates* sharing the file) and is the precedent for the reasoning, but it hands out a raw connection runner, not a store: it carries no `device_id` and implements none of the storage traits, so a second session cannot be built on it.
- **whatsmeow's `sqlstore.Container`** — one container, one pool, N devices, which is the shape this converges on. Not copied as API: Go's `database/sql` pools freely because its callers do not need the write serialization this store's single permit provides, so the write-throughput trade above is ours to state, not theirs.
- **`pool_size`** — untouched. It is 1 for the documented reason, and sharing does not change that reason. Note the docs are scoped accordingly: `db_semaphore` carries `pool_size` permits, so siblings serialize at the default of 1 and inherit a wider pool's width otherwise.
- **`resource_report()`** — left as is, and documented. It describes the pool, which siblings share, so each handle reports the same whole-pool estimate; a fleet's reports must be counted once per pool rather than summed. Attributing shared bytes to one arbitrary handle would be a worse answer than saying they belong to the pool.
- **Semantics of every existing `*_for_device` method** — untouched; the new method only produces a store with a different `device_id`.
- **The mutation paths that take a pooled connection without the write permit** (`delete_identity_for_device`, `delete_session_for_device`, `put_sender_key_for_device`, `delete_sender_key_for_device`, `set_app_state_sync_key_for_device`) — real, unchanged by this PR, and already documented where the decision is made. At `pool_size = 1` they serialize on the pool checkout instead of the permit; at `pool_size > 1` they are exactly the hazard the `SqliteStore::reads` comment describes and the reason `read_pool_size` is a second pool rather than a wider first one. Routing them through `with_retry` is what would make `pool_size > 1` safe, and it deserves its own PR rather than riding along on a memory change.
- **`tests/e2e/tests/process_footprint.rs`** (from #1238) — the neighbouring harness, kept and not duplicated. It measures whole-session marginal RSS with an `RssAnon`/`RssFile` split and needs the mock server; this one isolates the per-*connection* cost and runs without it. Different question, no overlap to collapse.

## Validation

- `cargo fmt --all`
- `cargo clippy --workspace --all-targets -- -D warnings` (excluding `whatsapp-rust-voip-cli`, whose `alsa-sys` build script needs ALSA headers absent from this container — unrelated to this change)
- `cargo test --workspace` (excluding the same crate and `e2e-tests`, which needs the mock server): 60 test binaries, 0 failures, including the crate's 78 tests and both doctests.
- New tests in `share_for_device_tests`:
  - `siblings_share_the_database_but_not_each_other_s_rows` — writes through one handle, reads through the other; proves both see one database and that `device_id` isolates their rows in both directions.
  - `a_fleet_of_handles_opens_one_connection` — counts connections via `pool.state().connections`: 8 handles → 1, 8 separate stores → 8. Also asserts the shared write permit via `Arc::ptr_eq`, so the trade-off is asserted, not assumed.
  - `concurrent_writes_serialize_across_siblings_at_the_default_pool_size` — runs the burst both ways, printing timings, and asserts the two properties that must hold on any machine: no write fails, and no sibling starves on the shared queue. `pool_size` is set explicitly because the serialization claim holds for that value. Wall-clock thresholds are deliberately not asserted; the release numbers above come from the example instead.

**Known-failing check:** `Semver Checks (informational)` is red, and was already red before this branch. It runs against `-p wacore -p wacore-binary -p waproto` only — none of which this PR touches (the diff is two files under `storages/sqlite-storage/`). Its findings are pre-existing drift versus the published 0.7.0: regenerated `wacore::iq::mex_operations` items, a removed `simd` feature, and `BinaryError::UnexpectedFormatByte`. The job is informational by design and its own summary says so.

### DIFF
--- a/storages/sqlite-storage/examples/per_connection_memory.rs
+++ b/storages/sqlite-storage/examples/per_connection_memory.rs
@@ -201,14 +201,21 @@ async fn main() {
     let cache_kib: u32 = args.get(2).and_then(|a| a.parse().ok()).unwrap_or(512);
     let warm = args.iter().any(|a| a == "warm");
 
-    let mut path = std::env::temp_dir();
-    path.push(format!("wa_percon_{}.db", std::process::id()));
-    for suffix in ["", "-wal", "-shm"] {
-        let mut p = path.clone().into_os_string();
-        p.push(suffix);
-        let _ = std::fs::remove_file(p);
+    // A directory of our own, created exclusively: `create_dir` fails outright
+    // if anything already sits at the path (a symlink included), so nobody who
+    // can write to the shared temp directory can pre-place one and redirect the
+    // database, WAL and shm files this then writes.
+    let dir = std::env::temp_dir().join(format!("wa_percon_{}", std::process::id()));
+    std::fs::create_dir(&dir).expect("exclusive scratch directory");
+    // A guard, not a tail cleanup: two of the modes below return early.
+    struct ScratchDir(std::path::PathBuf);
+    impl Drop for ScratchDir {
+        fn drop(&mut self) {
+            let _ = std::fs::remove_dir_all(&self.0);
+        }
     }
-    let url = path.to_string_lossy().into_owned();
+    let scratch = ScratchDir(dir.clone());
+    let url = dir.join("bench.db").to_string_lossy().into_owned();
 
     if mode == "compile-options" {
         compile_options(&url).await;
@@ -271,9 +278,5 @@ async fn main() {
     );
 
     drop(stores);
-    for suffix in ["", "-wal", "-shm"] {
-        let mut p = path.clone().into_os_string();
-        p.push(suffix);
-        let _ = std::fs::remove_file(p);
-    }
+    drop(scratch);
 }
--- a/storages/sqlite-storage/examples/per_connection_memory.rs
+++ b/storages/sqlite-storage/examples/per_connection_memory.rs
@@ -1,0 +1,279 @@
+//! What one extra SQLite connection costs, measured rather than estimated.
+//!
+//! A process that holds N WhatsApp sessions against the *same* database file
+//! opens N stores, and every constructor builds its own r2d2 pool — so N
+//! connections. This harness prices that: it seeds one database, then opens N
+//! sessions two ways and reports the resident-set delta per session.
+//!
+//! ```text
+//! cargo run -p whatsapp-rust-sqlite-storage --release \
+//!     --example per_connection_memory -- <pools|handles> <sessions> <cache_kib> [warm]
+//! cargo run -p whatsapp-rust-sqlite-storage --release \
+//!     --example per_connection_memory -- writes <sessions> <read_pool_size>
+//! cargo run -p whatsapp-rust-sqlite-storage --release \
+//!     --example per_connection_memory -- compile-options
+//! ```
+//!
+//! * `pools` — one `SqliteStore::new_for_device` per session (today's shape).
+//! * `handles` — one store, then `share_for_device` per session (one pool).
+//! * `warm` — every session scans the seeded table first, filling its page
+//!   cache to the `cache_kib` cap. Without it each session only does the small
+//!   reads an idle session does, which is the realistic steady state.
+//! * `writes` — the other side of the trade: every session writes at once,
+//!   both ways, reporting total time and the spread between the fastest and
+//!   slowest session (i.e. whether anyone starves).
+//!
+//! Resident set, not a Rust allocator counter: SQLite's page cache and
+//! lookaside are `sqlite3_malloc` allocations from the bundled C library, which
+//! a `GlobalAlloc` wrapper never sees. RSS is coarse (page-granular, and
+//! includes whatever the allocator declines to return), so it is read after a
+//! settle and divided across enough sessions for the per-session figure to
+//! outweigh the noise. Linux only.
+
+// The numbers *are* this binary's output; there is no logger to route them
+// through, and a measurement harness whose result lands in a log filter would
+// be worse than useless.
+#![allow(clippy::print_stdout)]
+
+use std::time::Duration;
+
+use diesel::prelude::*;
+use wacore::store::traits::SignalStore as _;
+use whatsapp_rust_sqlite_storage::{SqliteStore, SqliteStoreConfig};
+
+/// Rows of ~1 KiB each: enough database that a full scan can fill a 512 KiB
+/// page cache several times over, so the cap is what bounds a warm connection.
+const SEED_ROWS: usize = 4_000;
+const ROW_BYTES: usize = 1_024;
+
+fn rss_bytes() -> u64 {
+    // Field 2 of /proc/self/statm is resident pages. 4 KiB is the page size on
+    // every platform this harness is meant to run on; a wrong constant would
+    // scale every number here equally, so comparisons stay valid regardless.
+    let statm = std::fs::read_to_string("/proc/self/statm").expect("/proc/self/statm");
+    let pages: u64 = statm
+        .split_whitespace()
+        .nth(1)
+        .and_then(|f| f.parse().ok())
+        .expect("resident pages");
+    pages * 4096
+}
+
+fn db_err(e: diesel::result::Error) -> wacore::store::error::StoreError {
+    wacore::store::error::StoreError::Database(Box::new(e))
+}
+
+/// Seed a database large enough that page caches have something to hold.
+async fn seed(url: &str) {
+    let store = SqliteStore::new_for_device(url, 1).await.expect("open");
+    let record = vec![0x5au8; ROW_BYTES];
+    for chunk in (0..SEED_ROWS).collect::<Vec<_>>().chunks(200) {
+        let batch: Vec<_> = chunk
+            .iter()
+            .map(|i| {
+                (
+                    format!("seed.{i}:0").into(),
+                    bytes::Bytes::from(record.clone()),
+                )
+            })
+            .collect();
+        store.put_sessions_batch(&batch).await.expect("seed write");
+    }
+}
+
+/// The reads an idle session actually does on connect: a couple of point
+/// lookups. Also forces r2d2 to open the connection, which is the allocation
+/// this harness is pricing.
+async fn touch(store: &SqliteStore) {
+    store.get_session("seed.0:0").await.expect("point read");
+    store.get_session("seed.1:0").await.expect("point read");
+}
+
+/// A full table scan, which pulls pages in until the cache cap stops it.
+async fn scan(store: &SqliteStore) {
+    #[derive(QueryableByName)]
+    struct Count {
+        #[diesel(sql_type = diesel::sql_types::BigInt)]
+        n: i64,
+    }
+    store
+        .shared()
+        .read(|conn| {
+            diesel::sql_query("SELECT count(record) AS n FROM sessions")
+                .get_result::<Count>(conn)
+                .map(|c| c.n)
+                .map_err(db_err)
+        })
+        .await
+        .expect("scan");
+}
+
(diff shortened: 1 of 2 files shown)


jlucaso1:
Review round: six of the findings were valid and are fixed in b2fa5ea; two are declined with reasons. The PR body now carries the corrected numbers.

**Fixed**

- **Both write arms shared one database** (CodeRabbit, Codex). Real: `put_session` is an upsert, so the second arm measured conflict updates over the first's rows and warmed pages. Each arm now gets its own database file.
- **Seeding through the allocator being measured** (Codex). The sharpest one. The harness now also reports the **marginal** cost — RSS growth from the first session to the last, over the sessions in between — so whatever the seeding connection left behind for session 1 to reuse is subtracted out rather than spread across the average. It moves the conclusion not at all and sharpens it: **122.6 KiB per additional session** with a pool each, **0.0 KiB** with shared handles.
- **Hard-coded 4 KiB page size** (CodeRabbit, Codex). Now reads `Rss:` from `/proc/self/smaps_rollup`, which is already in kB, so nothing depends on the kernel's page size — and it needs no libc.
- **Write serialization claimed unconditionally** (CodeRabbit). Correct: `db_semaphore` carries `pool_size` permits, so siblings serialize at the default of 1 and inherit a wider pool's width otherwise. Doc scoped, and the test now sets `pool_size` explicitly and says so in its name.
- **Zero sessions** (CodeRabbit) — rejected up front.
- **Predictable scratch path / umask** (Greptile, both rounds). The harness now creates its own directory with `DirBuilder::mode(0o700)`, which is both an exclusive create (it fails if anything, symlink included, already occupies the path) and owner-only regardless of umask.

**Declined**

- **`spawn_blocking` for the harness's filesystem I/O** (CodeRabbit). Skipping. This is a measurement harness on a current-thread runtime with nothing else scheduled at the points it reads `/proc`; routing those reads through a thread hop would add allocation and scheduling noise to the very RSS number being taken. The repo's blocking-work rule is about not stalling the client's read loop, which this binary does not have.
- **`resource_report()` counting a shared pool once per sibling** (Codex). Real, and now documented on `share_for_device` rather than papered over. The report describes the *pool*, and siblings share one, so each handle returns the same whole-pool estimate; a fleet's reports must be counted once per pool instead of summed. I did not change the behaviour because the alternatives are worse: attributing shared bytes to one arbitrary handle invents an owner, and returning `None` for siblings discards a real measurement. The bytes belong to the pool, and the point of this PR is that there is now only one pool left to count.

`Semver Checks (informational)` is red and was red before this branch — it runs only against `wacore`, `wacore-binary` and `waproto`, none of which this PR touches (the diff is two files under `storages/sqlite-storage/`). Its findings are pre-existing drift versus the published 0.7.0.

---

jlucaso1:
On the unguarded mutation paths (`delete_identity_for_device`, `delete_session_for_device`, `put_sender_key_for_device`, `delete_sender_key_for_device`, `set_app_state_sync_key_for_device`): the observation is factually correct — they check a connection out of the pool directly, without taking `db_semaphore`. Declining anyway, for three reasons.

**It is unchanged by this PR.** `git show origin/main` has the same bodies; `share_for_device` clones the pool and semaphore and touches neither. Two separately-constructed stores at `pool_size > 1` have exactly the same exposure today. The task this branch came from also rules the change out explicitly: don't alter the semantics of existing `*_for_device` methods.

**At the default it is not a bug.** With `pool_size = 1` the pool hands out one connection at a time, so those writes serialize on the checkout rather than on the permit. Same outcome, different mechanism.

**At `pool_size > 1` it is the documented reason not to raise `pool_size`,** and the repo already says so at the point the decision is made — the `SqliteStore::reads` field comment:

> Deliberately a second pool rather than extra connections in the main one: several write paths check a connection out directly, without the semaphore, and are serialized today only because the pool hands out one connection at a time. Growing that pool would let two of them run at once and deadlock on the write-lock upgrade — the exact failure this change exists to avoid.

That is why `read_pool_size` exists as a separate pool, and why `pool_size` is 1 with a doc telling callers to leave it there. Repeating that rationale on `share_for_device` is the duplication `AGENTS.md` warns against ("Comments carry the *why* of a decision, at the single point where it is made. Repeating a rationale at call sites is how it goes stale"), so the `share_for_device` doc states only what it changes: siblings share whatever write width `pool_size` configured, which is 1 by default.

Adding the suggested `pool_size: 2` racing test would encode a configuration the crate documents as unsafe. If routing every mutation through `with_retry` is wanted, it is a worthwhile change on its own — it is what would make `pool_size > 1` safe — but it belongs in its own PR, against `main`, where it can be reviewed as the concurrency change it is rather than as a rider on a memory one.

Happy to open that as a follow-up issue if you'd like it tracked.

---

PR #1306, opened 2026-08-14
fix(appstate): box traced resync future

nizarfadlan:
## Summary

- keep the deep traced `resync_app_state` future behind a `BoxFuture` boundary
- preserve the existing public `async fn` API and the `wa.appstate.resync` tracing span
- add a downstream-style compile regression test for consumers with the `tracing` feature

## Problem

A downstream crate that enabled `whatsapp-rust/tracing` and awaited `Client::resync_app_state` could hit Rust's recursion depth limit while computing the concrete async future layout. The library crate's `#![recursion_limit = "512"]` does not apply to downstream crates, so adding the attribute only to the doctest would not address the public API boundary.

## Solution

The public method now normalizes the iterator and awaits a private boxed helper. The helper returns `wacore::runtime::BoxFuture` and keeps the instrumented async implementation in a private graph function. This type-erases the deep future before it crosses into a consumer crate, while retaining the existing tracing instrumentation and public API.

## Verification

- `cargo test -p whatsapp-rust --test app_state_resync_future_compat --features tracing`
- `cargo test -p whatsapp-rust --doc --features tracing`
- `cargo nextest run -p whatsapp-rust --lib --features tracing`
- `cargo check -p whatsapp-rust --all-features`
- `cargo clippy -p whatsapp-rust --all-targets --features tracing -- -D warnings`

All checks pass.

### DIFF
--- a/src/features/app_state_resync.rs
+++ b/src/features/app_state_resync.rs
@@ -205,13 +205,32 @@ impl Client {
     /// # Ok(())
     /// # }
     /// ```
-    #[cfg_attr(feature = "tracing", tracing::instrument(name = "wa.appstate.resync", level = "debug", skip_all, fields(mode = ?mode), err(Debug)))]
     pub async fn resync_app_state(
         &self,
         collections: impl IntoIterator<Item = WAPatchName>,
         mode: AppStateResyncMode,
     ) -> Result<AppStateResyncReport, AppStateError> {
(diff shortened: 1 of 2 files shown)


jlucaso1:
Thanks <3

nizarfadlan:
> Thanks <3

You're welcome, brother

PR #1307, opened 2026-08-14
fix(receipt): address a read our own account authored

### LINKED ISSUE #1305: receipt: an own-device `read` fan-out is never classified as a self-receipt, so a primary-device read never clears companion unread
## What happens

With read receipts **enabled**, a chat read on the primary device never clears the companion's unread badge.
Reproduced live at `ed7723a9` (also present on `main`):

1. A DM arrives at the companion from a LID-addressed peer (`265051105669299@lid`, `sender_alt` carrying the peer's PN). Message ids `AC7B00A0DDAC597FAE20C2A5DB4EDAC1` (07:01:55Z) and `AC240BE154E1B438920A9718643457AE` (07:15:07Z). Chat-store badges them correctly.
2. The primary reads the chat. The peer's device shows blue ticks, so the primary emitted its `read` receipt and the server fanned it.
3. The companion receives, ~30-60 s after each message (human latency, ids matching exactly):

```json
{"message_ids":["AC7B00A0DDAC597FAE20C2A5DB4EDAC1"],"offline":false,
 "source":{"chat":"150376166744228@lid","is_from_me":false,"is_group":false,"sender":"150376166744228@lid"},
 "timestamp":"2026-08-14T07:02:18Z","type":"Read"}
```

`150376166744228@lid` is the **receiving account's own LID** (the connect span logs `lid=150376166744228:2@lid` for this device). The receipt is the server's self-fanout of the primary's read - but it is dispatched as a plain peer `Read` whose chat is the account's own LID. The chat-store's Read arm then advances outgoing-row statuses in a chat keyed by the own LID, finds nothing, and the badge never moves.

## Why the parse is wrong

`src/receipt.rs`, inbound path (both the simple and the aggregated `<participants>` shapes):

```rust
.source(crate::types::message::MessageSource {
    chat: from,
    sender: default_sender,
    sender_alt: participant_pn,
    ..Default::default()
})
```

- `is_from_me` is never computed: nothing compares `from` against the device's own PN or LID user.
- The `recipient` attribute - which carries the actual peer chat on a self-fanned receipt - is written on the outbound self-fanout (`src/receipt.rs:169-170`) but never read inbound.
- `ReceiptType::parse` maps only the literal `read-self` (`wacore/src/types/presence.rs:68`). `read-self` is the flavor the sender uses when read receipts are **disabled** (the outbound side documents this split at `src/receipt.rs:80-83`). With receipts enabled - the default configuration - the fan-out arrives as plain `read` from an own device, and nothing recognizes it.

The store side is already correct and unreachable: `apply_receipt`'s `ReadSelf | PlayedSelf` arm (`storages/chat-store/src/store.rs:1678`) routes the LID onto the real chat key, advances the monotonic read state, and recomputes the badge - but only fires on `ReceiptType::ReadSelf`, which this wire shape never produces.

WA Web's `WAWebHandleMsgReceiptParser` classifies a receipt authored by the own account (`isMeAccount`) and re-addresses the chat from `recipient`; whatsmeow's `parseReceipt` does the same (`from.User == own user` → `IsFromMe`, `Chat = recipient` when the attribute is present).

## Why receipts are the only carrier

Primaries do not write a `MarkChatAsRead` app-state mutation for an ordinary chat open: a **snapshot** resync of `regular_low` performed immediately after the read applied cleanly and carried no mark for the chat (while an archive performed on the primary moments later synced and applied within seconds, so the app-state pipeline itself is healthy). Dropping the receipt fan-out therefore breaks multi-device read sync entirely for accounts with read receipts enabled.

## Fix sketch

In the inbound receipt handler, after `attrs.jid("from")` (same treatment in the aggregated path):

- Compare `from`'s user against the device's **own PN and own LID** users (the device store's identities, not the LID cache).
- When own: set `source.is_from_me`, re-address `source.chat` from the `recipient` attribute when present (falling back to `from`), and treat `Read`/`Played` as `ReadSelf`/`PlayedSelf` - either by promoting the type or by branching on `is_from_me` in the chat-store arm.

The existing `ReadSelf` arm then does the rest unchanged.

nizarfadlan:
## Summary

Classifies a receipt our own account authored, so a chat read on the primary device reaches this one as a read of the peer's thread instead of a read of a thread with ourselves.

Fixes #1305.

## Motivation

The primary reads a DM. The server fans that `read` back to every companion, addressed `from` our own account with the peer named in `recipient`. Nothing on the inbound path looked at either: the event went out with `chat` set to our own JID and `is_from_me` never computed, so the chat store advanced read state on a thread that does not exist and the unread badge stayed put.

This is the whole carrier for the feature. A primary does not write a `MarkChatAsRead` app-state mutation when a chat is simply opened, so dropping the receipt drops multi-device read sync entirely — and with read receipts enabled, which is the default, the fan-out arrives as a plain `read` that nothing recognised as ours.

## Design notes

**The author is `from` on a DM and `participant` on a group.** WA Web computes `h = !isSender && isMeAccount(author)` in both `handleChatSimpleReceipt` and `handleGroupSimpleReceipt`, and the group one keys it off the participant. Since `default_sender` already resolves to exactly that JID, one `is_own_jid` call covers both. Keying off `from` alone would have left every group read unclassified.

**A DM moves to its peer thread; a group does not.** WA Web re-addresses `chat` to `recipient` on a DM, but in a group the chat is already the group and `recipient` names the author of the message that was read. Extending the DM rule to groups would file group reads under a participant. The attribute is therefore not carried as a `recipient` on the group source either — it is not one.

**Only a read or a play is acted on.** WA Web gates the self branch on `isReadOrPlayedReceipt` (`read`, `read-self`, `played`, `played-self`) and drops the rest, and two of the ones it drops matter:

- A **delivery** we authored says one of our own devices received the message, not that the peer did. Re-addressing it would put a delivered tick on the peer's thread that no peer earned.
- A **retry** is the one receipt whose target another pipeline re-derives for itself. `resolve_retry_chat_info` reads `source.chat` as the wire `from` — that is how it tells a retry from one of our own devices apart from a peer's — and then re-addresses to `recipient` on its own. Handing it a chat already re-addressed here makes the peer-device branch stop matching, and the requester resolves to the peer's bare JID instead of the device that asked, so the re-encryption goes to the wrong session.

**`read-self` needs the same move even though it has nothing to promote.** It is the flavour a primary sends when read receipts are turned off. Its type was already right, but its chat was still our own account, so that half of the bug was invisible behind the other one.

**Promoting the type rather than making consumers branch on `is_from_me`.** WA Web maps `read` and `read-self` onto one ack (`RECEIPT_TYPES_TO_ACK`): the wire type does not carry self-ness, the author does. Folding it in puts that fact in the type consumers already branch on, so a read synced from another device reaches the same handler whether or not the account has read receipts enabled. The chat store's `ReadSelf | PlayedSelf` arm — which routes the LID onto the real chat key and recomputes the badge — was already correct and simply unreachable; it needed no change.

**Aggregated receipts are left alone.** `handleAggregateReceipt` rejects anything that is not a group or broadcast, the parser never reads `recipient` in that shape, and `deaggregateGroupedByTypeReceipt` explicitly sets it to null. A self fan-out cannot arrive aggregated, and each `<user>` names a different peer that one stanza-level attribute could not address anyway.

## Behaviour

- A self receipt that names no peer thread is kept but left exactly as it arrived. WA Web throws on this shape; we log it and decline to classify, because a `ReadSelf` still addressed to our own account would advance read state on a phantom thread. `is_from_me` stays unset rather than half-classifying.
- Same for a `recipient` that names something other than a DM. It cannot happen on any shape WA Web produces — it dispatches on `from` and parses the attribute as a user JID — but a chat we would not route there is not the peer thread this is for.
- `chat` drops a device from the recipient and `recipient` keeps it, the same split `parse_message_info` makes for a message we sent.
- Everything a peer authored is byte-for-byte what it was before.

## Compatibility

`Receipt` gains populated `source.is_from_me` and `source.recipient` on the DM self shape, and its `type` reads `ReadSelf`/`PlayedSelf` where it used to read `Read`/`Played`. A consumer matching on `Read` to detect "the peer read this" is unaffected — that arm only ever saw peer receipts before, and still does. A consumer that was matching `Read` and reading `chat` as our own JID to detect the fan-out by hand would need to move to `ReadSelf`, but there was nothing usable to do with it.

## Validation

- [x] `cargo fmt --all`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features`
- [x] `cargo nextest run -p whatsapp-rust --lib` — 1698 passed
- [x] `cargo nextest run --workspace --exclude e2e-tests` — 4909 passed

Fifteen tests: ten on the addressing itself, five driving `handle_receipt` end to end. Each decision was checked against its own mutation, and each mutation fails the test that claims it:

| Mutation | Test that catches it |
| --- | --- |
| act on every type we authored | `a_delivery_we_authored_is_left_alone`, `a_retry_we_authored_keeps_the_wire_from` |
| gate on `Read`/`Played` only | `a_read_self_we_authored_still_moves_to_the_peer_thread` |
| re-address a group read to its recipient | `a_group_read_we_authored_keeps_the_group_as_its_chat` |
| check the author against `from`, not the participant | `a_group_read_from_our_own_participant_is_a_self_read` |
| classify a receipt that names no peer thread | `a_read_we_authored_without_a_peer_thread_is_left_alone` |
| drop the non-DM recipient guard | `a_read_we_authored_naming_a_non_peer_thread_is_left_alone` |
| drop the promotion | six of the read/played tests |

## Known limitation, checked and left out

A read of a **status** update is classified by WA Web the same way (`handleStatusReceipt` also runs `isMeAccount` on the participant), and is not covered here. Our receipt path resolves `participant` into `sender` for groups only, so `from` stays `status@broadcast` and the author is never seen. Fixing it means changing what `sender` holds on every status receipt, which is a separate change to a separate pipeline, and status unread is not the state #1305 is about.

### DIFF
--- a/src/receipt.rs
+++ b/src/receipt.rs
@@ -16,15 +16,121 @@ use wacore_binary::OwnedNodeRef;
 /// per chunk, so a large catch-up doesn't produce one oversized stanza.
 const MAX_RECEIPT_IDS_PER_STANZA: usize = 256;
 
-fn normalize_self_fanout_type(receipt_type: ReceiptType, is_self_fanout: bool) -> ReceiptType {
-    if !is_self_fanout {
-        return receipt_type;
+/// Whether `chat` names a DM thread — the shape WA Web hands to
+/// `handleChatSimpleReceipt`. Same exclusions as
+/// [`MessageSource::is_self_fanout`](crate::types::message::MessageSource::is_self_fanout),
+/// so the two cannot disagree on what counts as one.
+fn is_peer_thread(chat: &Jid) -> bool {
+    !chat.is_group() && !chat.is_status_broadcast() && !chat.is_newsletter()
+}
+
+/// How a simple `<receipt>` is addressed once the author is known.
+struct ReceiptAddressing {
+    chat: Jid,
+    recipient: Option<Jid>,
+    is_from_me: bool,
+    receipt_type: ReceiptType,
+}
+
+impl ReceiptAddressing {
+    /// The receipt exactly as it arrived on the wire, for every shape this
+    /// classification does not act on. `is_from_me` stays unset there rather
+    /// than half-classifying: a consumer would otherwise find a receipt flagged
+    /// as ours and still addressed to our own account.
+    fn as_received(from: Jid, receipt_type: ReceiptType) -> Self {
+        Self {
+            chat: from,
+            recipient: None,
+            is_from_me: false,
+            receipt_type,
+        }
+    }
+}
+
+/// Classify a simple `<receipt>` the way WA Web's `handleChatSimpleReceipt` and
+/// `handleGroupSimpleReceipt` do — `h = !isSender && isMeAccount(author)`, where
+/// the author is `from` on a DM and `participant` on a group. Extracted from the
+/// handler so the addressing can be asserted without spinning a transport;
+/// `author_is_own_account` is the caller's `is_own_jid` verdict on that author.
+fn address_receipt(
+    from: Jid,
+    recipient: Option<Jid>,
+    receipt_type: ReceiptType,
+    is_group: bool,
+    author_is_own_account: bool,
+) -> ReceiptAddressing {
+    if !author_is_own_account {
+        return ReceiptAddressing::as_received(from, receipt_type);
     }
 
-    match receipt_type {
+    // Only a read or a play is actionable when we authored it: WA Web gates the
+    // self branch on `isReadOrPlayedReceipt` and drops every other flavour. Two
+    // of the ones it drops matter here.
+    //
+    // A delivery we authored says one of our own devices received the message,
+    // not that the peer did, so re-addressing it would put a delivered tick on
+    // the peer's thread on our own device's behalf.
+    //
+    // A retry is the one receipt whose target another pipeline re-derives for
+    // itself: `resolve_retry_chat_info` reads `chat` as the wire `from` to spot a
+    // retry from one of our own devices, so handing it a chat we already
+    // re-addressed makes it aim the re-encryption at the peer thread instead of
+    // the device that asked for it.
+    if !matches!(
+        receipt_type,
+        ReceiptType::Read | ReceiptType::ReadSelf | ReceiptType::Played | ReceiptType::PlayedSelf
+    ) {
+        return ReceiptAddressing::as_received(from, receipt_type);
+    }
+
+    // On a DM the thread is the peer named by `recipient`; `from` is our own
+    // account, so keeping it would file the read under a thread with ourselves.
+    // On a group the thread is already `from`, and the attribute names the author
+    // of the read message rather than a recipient, so it is not carried as one.
+    let (chat, recipient) = if is_group {
+        (from, None)
+    } else {
+        match recipient {
+            // WA Web rejects a self receipt that names no peer thread. We keep
+            // the event rather than drop it, but exactly as it arrived: a
+            // `ReadSelf` still addressed to our own account would advance read
+            // state on a thread that is really us.
+            None => {
+                log::warn!(
+                    "Own-account receipt from {} names no peer thread; leaving it unclassified",
+                    from.observe()
+                );
+                return ReceiptAddressing::as_received(from, receipt_type);
+            }
+            Some(peer) if !is_peer_thread(&peer) => {
+                log::warn!(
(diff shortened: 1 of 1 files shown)


jlucaso1:
Thank you

PR #1313, opened 2026-08-16
feat(lid): act on the server's refresh_lid ack flag

jlucaso1:
## What

An `<ack refresh_lid="true">` is the server telling us the LID-PN mapping we hold for that peer is stale. Nothing read the attribute.

## Why it matters

That flag is the only invalidation this client gets for LID-PN mappings. `src/lid_pn_cache.rs` states outright that entries never expire, and no other path re-queries an existing mapping: `swap_pn_lid_namespace`, `resolve_recipient_to_lid` and the usync helpers are all cache-aside, so a hit is returned and never revalidated.

So a mapping that goes stale stays stale for the lifetime of the process. Every Signal address derived from it keeps resolving to the wrong identity, and the encrypt path has no way to notice.

## What it does

`refresh_lid_peer_from_ack` reads the flag and the ack's sender; `maybe_refresh_lid_from_ack` spawns the refresh. Both ack entry points (`handle_ack_response_arc`, `handle_ack_response_owned`) call it **before** taking the response waiter, because a send ack carries the flag whether or not anything is waiting on it, and both entry points return early when no waiter exists.

The common path costs one failed attribute lookup on the read loop.

## Addressing

The refresh is addressed by **phone number**, not by the LID:

- Asking about a LID we already suspect is wrong would look the answer up under the very key in doubt.
- WA Web maps the peer through `toPn` before its query, and drops the refresh when no mapping exists to map through. A peer we have never resolved has nothing stale to correct, so it gets no query.
- A peer that arrives as a phone number is already the side the server resolves from, so it is queried as-is.

Target resolution keys on the **bare user** and accepts both server families (`@lid`/`@hosted.lid`, `@s.whatsapp.net`/`@hosted`), because that is how the cache and the usync query both address a number, and because the message learn path (`src/message/receive.rs`) records hosted pairs into this same cache — a hosted peer derives its Signal address from it and can go stale identically.

## The query

`reconcile_lid_mappings_via_usync`, which is already what an observational conflict (`RecordOutcome::NeedsUsync`) uses for an authoritative live re-resolve. `LidQuerySpec` requests `vec[UsyncProtocol::Lid]` and nothing else, and persists through `add_lid_pn_mapping(.., LearningSource::Usync)`.

This matches what WA Web pins on the refresh:

```js
var r = toPn(peer);
r && workerSafeFireAndForget("syncContactListJob",
  { contactIds: [r], shouldSyncDevice: false, mode: "query" });
```

`mode: "query"` is `UsyncMode::Query` and `shouldSyncDevice: false` means no `<devices>` — both are what `LidQuerySpec` carries.

The contact query (`is_on_whatsapp`) would be the wrong carrier: it also requests `<contact>` and `<business>`, and `parse_response` opens with `reject_result_errors(&response, LEGACY_RESULT_PROTOCOLS)?`, which rejects the whole response on a result-level error in *any* requested subprotocol. An unrelated verified-name failure would discard a correction the server did return.

A failed query leaves the mapping exactly as stale as it was: nothing to roll back, and nothing to retry more cheaply than the next ack carrying the flag again.

## Coalescing

The flag rides on the send-message ack (WA Web's `sendMsgAckSyncParser`), so a burst of sends to one stale peer is acked one message at a time and every ack repeats it. Without a guard the second onward would duplicate a query already in flight.

Guarded per bare number with the same `Arc<Mutex<HashSet<String>>>` + `scopeguard` shape `handle_retry_receipt` uses for `pending_retries`, and cleared on teardown so a socket that dies mid-query does not leave a key suppressing the next refresh. Reported from `memory_report()`, which `tests/report_coverage.rs` requires for any growable field on `Client`.

## Tests

Nine, split so each stage is asserted without the spawn in between.

Gate (`src/client/node_io.rs`):
- an ack with no flag, and one with `refresh_lid="false"`, request nothing (WA Web's parser is a strict `=== "true"`)
- a flagged ack names its sender as the peer
- a flagged ack with no sender names nobody

Addressing (`src/client/lid_pn.rs`):
- a flagged LID peer maps to its phone number, device suffix dropped (the mapping is per user)
- a LID peer with no mapping resolves to nothing
- a phone-number peer is queried directly
- `@hosted.lid` and `@hosted` resolve like their strict counterparts
- groups, newsletters and status broadcast resolve to nothing
- a number already being refreshed does not start a second query, and the owning call clears its key even when the query fails

Fictitious JIDs throughout.

## Verification

- `cargo fmt --all`
- `cargo test -p whatsapp-rust --lib` — 1713 passed, 0 failed
- `cargo test -p whatsapp-rust --test report_coverage` — 5 passed
- `cargo clippy -p whatsapp-rust --lib --tests -- -D warnings` — clean

Semver Checks is advisory (`continue-on-error`) and fails on `main` too, on six pre-existing `waproto` findings; it exits before reaching `wacore`.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1311,11 +1311,14 @@ pub struct Client {
 
     pub(crate) pending_retries: Arc<std::sync::Mutex<HashSet<String>>>,
 
-    /// Phone numbers with a `refresh_lid` re-resolve in flight, keyed by the
-    /// bare user. A burst of sends to one stale peer is acked one message at a
-    /// time, and every one of those acks carries the flag, so without this the
-    /// same query would go out once per ack while the first is still pending.
-    pub(crate) pending_lid_refreshes: Arc<std::sync::Mutex<HashSet<String>>>,
+    /// Identities with a `refresh_lid` re-resolve in flight, keyed by
+    /// `(connection_generation, PN-side JID)`. A burst of sends to one stale
+    /// peer is acked one message at a time, and every one of those acks carries
+    /// the flag, so without this the same query would go out once per ack while
+    /// the first is still pending. The generation scopes a reservation to the
+    /// connection that took it, so a refresh left parked on a dead socket
+    /// cannot suppress the next connection's.
+    pub(crate) pending_lid_refreshes: Arc<std::sync::Mutex<HashSet<(u64, String)>>>,
 
     /// Track retry attempts per message to prevent infinite retry loops.
     /// Key: "{chat}:{msg_id}:{sender}", Value: retry count plus the most
--- a/src/client.rs
+++ b/src/client.rs
@@ -497,6 +497,10 @@ pub struct MemoryReport {
     /// [`Self::node_waiters`]. Each retains a filter and a oneshot sender.
     pub sent_node_waiters: usize,
     pub pending_retries: usize,
+    /// Numbers with a `refresh_lid` re-resolve in flight. Bounded by the
+    /// number of distinct peers acked at once; a value that stays high
+    /// means refreshes are not completing, not that many were requested.
+    pub pending_lid_refreshes: usize,
     pub presence_subscriptions: usize,
     pub app_state_key_requests: usize,
     /// Expanded app-state keys the processor holds in memory. No capacity cap
@@ -646,6 +650,11 @@ impl std::fmt::Display for MemoryReport {
         writeln!(f, "  node_waiters:           {}", self.node_waiters)?;
         writeln!(f, "  sent_node_waiters:      {}", self.sent_node_waiters)?;
         writeln!(f, "  pending_retries:        {}", self.pending_retries)?;
+        writeln!(
+            f,
+            "  pending_lid_refreshes:  {}",
+            self.pending_lid_refreshes
+        )?;
         writeln!(
             f,
             "  presence_subscriptions: {}",
@@ -1301,6 +1310,12 @@ pub struct Client {
     pub(crate) pending_device_sync: crate::pending_device_sync::PendingDeviceSync,
 
     pub(crate) pending_retries: Arc<std::sync::Mutex<HashSet<String>>>,
+
+    /// Phone numbers with a `refresh_lid` re-resolve in flight, keyed by the
+    /// bare user. A burst of sends to one stale peer is acked one message at a
+    /// time, and every one of those acks carries the flag, so without this the
+    /// same query would go out once per ack while the first is still pending.
+    pub(crate) pending_lid_refreshes: Arc<std::sync::Mutex<HashSet<String>>>,
 
     /// Track retry attempts per message to prevent infinite retry loops.
     /// Key: "{chat}:{msg_id}:{sender}", Value: retry count plus the most
--- a/src/client/accessors.rs
+++ b/src/client/accessors.rs
@@ -312,6 +312,11 @@ impl Client {
             .lock()
             .unwrap_or_else(|p| p.into_inner())
             .len();
+        let pending_lid_refreshes_count = self
+            .pending_lid_refreshes
+            .lock()
+            .unwrap_or_else(|p| p.into_inner())
+            .len();
 
         // `get()`, not `get_group_cache()`: a report must not be what builds the
         // cache, so an un-warmed client still reports zero entries.
@@ -455,6 +460,7 @@ impl Client {
             node_waiters: self.node_waiter_count.load(Ordering::Relaxed),
             sent_node_waiters: self.sent_node_waiter_count.load(Ordering::Relaxed),
             pending_retries: pending_retries_count,
+            pending_lid_refreshes: pending_lid_refreshes_count,
             presence_subscriptions,
             app_state_key_requests,
             app_state_key_cache,
--- a/src/client/lid_pn.rs
+++ b/src/client/lid_pn.rs
@@ -10,6 +10,7 @@
 //! - Bidirectional lookup (LID to PN and PN to LID)
 
 use std::sync::Arc;
+use std::sync::atomic::Ordering;
 
 use anyhow::Result;
 use log::debug;
@@ -360,13 +361,15 @@ impl Client {
         let runtime = client.runtime.clone();
         runtime
             .spawn(Box::pin(async move {
-                client.reconcile_lid_mappings_via_usync(phones).await;
+                let jids = phones.iter().map(|p| Jid::pn(p.as_str())).collect();
+                client.reconcile_lid_mappings_via_usync(jids).await;
             }))
             .detach();
     }
 
-    async fn reconcile_lid_mappings_via_usync(&self, phones: Vec<String>) {
-        let jids: Vec<Jid> = phones.iter().map(|p| Jid::pn(p.as_str())).collect();
+    /// Takes JIDs rather than bare numbers because a `refresh_lid` peer can be
+    /// hosted, and usync validates `Hosted` as a server distinct from `Pn`.
+    async fn reconcile_lid_mappings_via_usync(&self, jids: Vec<Jid>) {
         let sid = self.generate_request_id();
         match self.execute(LidQuerySpec::new(jids, sid)).await {
             Ok(resp) => {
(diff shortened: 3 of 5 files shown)


jlucaso1:
Pushed `d66b7dc` addressing the review. Summary of what I took and what I did not.

## Taken: mapping-only query (Codex P2)

Correct, and it is worse than described. `IsOnWhatsAppSpec` requests `<contact>`, `<lid>` and `<business>`, and `parse_response` opens with:

```rust
reject_result_errors(&response, LEGACY_RESULT_PROTOCOLS)?;
```

`LEGACY_RESULT_PROTOCOLS` covers Contact, Lid, Business, Status, Picture and Devices, so a **result-level** error on any requested subprotocol rejects the entire response before `persist_lid_mappings` runs. An unrelated verified-name failure would have discarded a correction the server did return, leaving the stale mapping installed.

Now routed through `reconcile_lid_mappings_via_usync`, which is already what `RecordOutcome::NeedsUsync` uses for an authoritative re-resolve. `LidQuerySpec` asks for `vec[UsyncProtocol::Lid]` and nothing else, and persists via the same `add_lid_pn_mapping(.., LearningSource::Usync)`.

It also turns out to be closer to WA Web, not further. The consumer of the flag is:

```js
function f(e, t) {
  var n = t.refreshLid;
  if (n) {
    var r = o("WAWebLidMigrationUtils").toPn(e);
    r && workerSafeFireAndForget("syncContactListJob",
      { contactIds: [r], shouldSyncDevice: false, mode: "query" });
  }
}
```

WA Web pins exactly two knobs — `mode: "query"` and `shouldSyncDevice: false` — and `LidQuerySpec` is the spec that carries both (`UsyncMode::Query`, no `<devices>`). The PR body's claim that `is_on_whatsapp` was "the narrower query" was wrong; `LidQuerySpec` is narrower still.

## Taken: coalescing (Codex P2)

Correct. The flag rides on the **send-message ack** — the parser confirms it:

```js
new WADeprecatedWapParser("sendMsgAckSyncParser", function (e) {
  e.assertTag("ack");
  return { ..., refreshLid: e.hasAttr("refresh_lid") ? e.attrString("refresh_lid") === "true" : false, ... }
});
```

So a burst of sends to one stale peer is acked one message at a time and every ack repeats the flag. Guarded on the bare number with the same `Arc<Mutex<HashSet<String>>>` + `scopeguard` shape `handle_retry_receipt` uses for `pending_retries`, and cleared on teardown so a socket that dies mid-query does not leave a key suppressing the next refresh. Reported from `memory_report()` (`tests/report_coverage.rs` enforces that).

Incidentally the parser also confirms the strict `=== "true"` check, which is what the gate test pins.

## Taken: hosted namespaces (Codex P2)

Correct. `is_lid()` is `server == Server::Lid` and `is_pn()` is `server == Server::Pn`, both exact, so a `@hosted.lid` or `@hosted` ack was dropped — while `src/message/receive.rs:2011` uses `is_lid_family()`/`is_pn_family()` and records those pairs into this same cache. A hosted peer could therefore hold a mapping that no ack could ever refresh.

Target resolution now keys on the **bare user** rather than swapping namespaces, which is how the cache and the usync query both address a number anyway, so the family check is the whole fix rather than a half-extension.

## Not taken: stale reverse mapping (Greptile P1)

This is WA Web's behaviour, deliberately, and deleting the entry would be a regression.

`WAWebLidPnCache`:

```js
n.add = function (lid, entry) {
  if (!lid.isLid()) throw err("WAWebLidPnCache - Invalid add call");
  this.$1.set(lid.toString(), entry);            // LID map: always overwrite
  var pn = entry.phoneNumber;
  pn != null && this.$3(pn, entry);              // PN map: _updatePnMap
};
n.$3 = function (pn, entry) {                    // _updatePnMap
  var k = pn.toString(), existing = this.$2.get(k);
  existing != null
    ? existing.phoneNumberCreatedAt <= entry.phoneNumberCreatedAt && this.$2.set(k, entry)
    : this.$2.set(k, entry);
};
```

`add` never deletes the superseded LID key, and our `add_guarded` matches it including the `<=` timestamp guard. There is a `remove(lid)`, but it deletes the PN entry too and is not on the add path.

The asymmetry is the design, not an oversight. The LID map is historical — every LID ever seen resolves to its number — while the PN map is *current*, which is what the `created_at` guard enforces. That is what `lid_pn_cache.rs:11` documents. Deleting `LID_old → PN` would make an incoming stanza addressed from the old LID unresolvable, which is strictly worse than keeping it: the callers named in the finding all take a LID that **arrived on the wire** and map it to a number, and for an old LID that answer is still right.

The direction the server is disputing is the one used to *address outgoing* traffic, PN → LID, and that one does update: `lid_pn_write_policy(Usync, lid_unseen: true, exact: false)` returns `(write: true, _)`, `Usync` is not a stale source so `created_at = now_secs()`, and `add_guarded`'s `existing.created_at <= entry.created_at` therefore passes.

Separately, none of this is introduced here — it is the write semantics every usync path in the crate has always had.

## Semver Checks

Pre-existing and advisory (`continue-on-error`); it fails on `main` too, on six `waproto` findings, and exits before reaching `wacore`. This PR adds no public surface beyond the `MemoryReport` field.

## Verification

`cargo fmt --all`; `cargo test -p whatsapp-rust --lib` → 1713 passed, 0 failed; `cargo test -p whatsapp-rust --test report_coverage` → 5 passed; `cargo clippy -p whatsapp-rust --lib --tests -- -D warnings` → clean.

---

jlucaso1:
Fixed in `ebc1027`. CodeRabbit and Codex found the same race independently, and both are right — it was in the coalescing I added in the previous commit.

The cause was the teardown sweep, which I should not have written. Its keys are released by a `scopeguard`, and `scopeguard`'s default strategy runs the closure on **drop** as well as on completion. So a refresh whose query dies with the socket frees its own key whether the task finishes or is dropped at shutdown — there was never anything stale for the sweep to collect. All it could do was release a key a live task still owned, which is exactly the reconnect race: old refresh spans the disconnect, new connection takes the same number, old guard fires and frees the new reservation, next ack duplicates the query.

Dropping the sweep is Codex's first suggested option ("let the owning guard clear the old entry") and makes the reservation owned end to end by the task that took it, so there is no generation or token to keep in sync. `pending_retries` keeps its sweep — its comment says stale keys from detached cleanup *do* linger there — but that is a different mechanism and not one this set shares.

Regression test added: `a_disconnect_does_not_release_a_live_reservation` inserts a reservation, calls `disconnect().await`, and asserts the key survives. That fails if anyone reintroduces a bulk clear.

`cargo test -p whatsapp-rust --lib` → 1714 passed, 0 failed. Clippy clean.

Semver Checks remains the pre-existing advisory failure (`continue-on-error`, red on `main` too, six `waproto` findings, exits before `wacore`).

---

jlucaso1:
Fixed the test comment in `76ef5aa`. It restated why teardown does not sweep the set, which is a decision made in `cleanup_connection_state_inner` — `AGENTS.md` asks for the *why* at the single point where the decision is made, precisely so one copy cannot go stale against the other. The test doc now states the expected behaviour and points at that function.

The other flagged range (`refresh_lid_mapping_for`, ~1199-1272) I left as is. Its comment explains why the coalescing exists at all — a burst of sends is acked one message at a time and every ack repeats the flag — and that decision is made there, not in lifecycle. The lifecycle comment answers a different question: why the set is not swept on teardown. Two decisions, two points, no duplicated rationale between them.

`cargo test -p whatsapp-rust --lib client::lid_pn::tests` → 53 passed. Clippy clean.

---

jlucaso1:
Both new Codex findings were right, and both were regressions in code I added earlier in this branch. Fixed in `ce7d540`. I called the PR settled one comment too early — that was wrong of me.

## Hosted namespace was discarded at the query

`refresh_lid_target` accepted the hosted families but returned a bare user, and `reconcile_lid_mappings_via_usync` rebuilt it with `Jid::pn`. A peer that acked as `user@hosted` was queried as `user@s.whatsapp.net`.

What decided it was `wacore/src/iq/usync/query.rs`:

```rust
fn validate_user_jid(jid: &Jid, index: usize) -> Result<(), UsyncValidationError> {
    let supported = !jid.user.is_empty()
        && matches!(jid.server, Server::Pn | Server::Lid | Server::Hosted | Server::HostedLid | ...);
```

```rust
if !matches!(pn_jid.server, Server::Pn | Server::Hosted) {
    return Err(UsyncValidationError::InvalidPnJid { index });
}
```

usync models `Hosted` and `Pn` as distinct servers, and `pn_jid` accepts exactly those two — so the namespace is part of the identity on the wire, not decoration. The *cache* genuinely does not care (`resolve_encryption_jid` looks a `Server::Hosted` target up under the same bare-user key, using the namespace only to pick `HostedLid` for the result), which is what made the bare-user shortcut look fine. It was fine for the lookup and wrong for the query.

`refresh_lid_target` now returns the PN-side `Jid` with `Lid → Pn` and `HostedLid → Hosted`, and `reconcile_lid_mappings_via_usync` takes `Vec<Jid>` instead of `Vec<String>` (its existing caller builds `Jid::pn` at its own call site, unchanged).

## Reservation could suppress the next connection's refresh

The sharper half of the finding: removing the teardown sweep fixed the ownership race but opened a suppression window. Teardown does not await the detached tasks, so a refresh parked on a dead socket's IQ still holds its key while the next connection is already acking — that new ack hits the dedup gate and returns, then the old query fails and releases the key, and *neither* query refreshed the mapping.

Keyed by `(connection_generation, JID)` instead. A new connection's refresh is a different reservation, so it proceeds; a task can only ever release its own. Within a generation duplicates still coalesce. `connection_generation` is incremented in `cleanup_connection_state_inner`, i.e. exactly where the sweep used to be, so it is already the right boundary — and no sweep can come back, because a leaked key from generation N cannot block N+1.

## Tests

Two added, three updated (12 total on the feature):
- `refresh_lid_target_keeps_the_hosted_namespace` — `@hosted.lid` and `@hosted` both resolve to `pn@hosted`, not `pn@s.whatsapp.net`
- `a_stale_reservation_does_not_suppress_the_next_connection` — a reservation from the previous generation survives untouched *and* the new refresh runs and releases its own

## Side effect

Binary size dropped from **+21.12 KiB to +3.84 KiB** vs `main` (`.text whatsapp_rust` +17.78 KiB → +7.79 KiB) once the contact query stopped being reachable from this path.

## Verification

`cargo test -p whatsapp-rust --lib` → 1715 passed, 0 failed; `--test report_coverage` → 5 passed; `cargo clippy -p whatsapp-rust --lib --tests -- -D warnings` → clean.

---

PR #1314, opened 2026-08-16
feat(ib): act on the server's client_expiration deadline

jlucaso1:
## What

`<ib><client_expiration t=…>` is the server naming the date it expects to stop accepting the running client build. Nothing read it — it fell through to `warn!("Unhandled ib child: <client_expiration>")` and was dropped.

## How the gap was found

whatspec's `srvreq` document lists exactly one `<ib>` parser, and it is this one:

```json
{ "tag": "ib", "module": "WASmaxInClientExpirationClientExpirationRequest",
  "shape": { "parserName": "parseClientExpirationRequest",
    "fields": [ { "name": "from", "type": "jid" },
                { "name": "clientExpirationT", "wireName": "t", "type": "integer",
                  "sourcePath": ["client_expiration"], "parserRequired": false } ] } }
```

Cross-checked against WA Web's `<ib>` dispatcher, whose `INFO_TYPE` union is `dirty`, `edge_routing`, `offline`, `offline_preview`, `tos`, `thread_metadata`, `client_expiration`, `priority_offline_complete`, `recovery_nonce`. We handle the first four and `thread_metadata`; `client_expiration` is the one with a real consequence.

## What it does

The deadline is recorded on the device, stamped with the build it was issued against, and announced as `Event::ClientExpirationChanged`.

It is **notice, not an instruction**. The client keeps connecting until the server actually refuses it — the stanza is about the *build*, not this connection — so whether to ship a newer version or page an operator is the consumer's call, not something this client should decide by disconnecting.

## The decision rule

`ServerClientExpiration::decide` mirrors WA Web's `handleServerClientExpiration`:

```js
function e(newTime) {
  if (newTime == null) { clearServerClientExpirationOverride(); return; }
  const n = getServerClientExpirationOverride()?.timestamp;
  if (n != null && newTime >= n) return;
  const r = futureUnixTime(3 * DAY_SECONDS);
  setServerClientExpirationOverride("" + Math.max(r, newTime), VERSION_BASE);
}
```

Two rules, both load-bearing:

- **A deadline only moves closer.** A later answer — a stale retransmit, or a host that hasn't caught up — must not hand the build an extension the server never granted.
- **At least three days of notice.** A server that says "now" still has to leave a window in which the build can be replaced.

They interact in a way worth stating, because I got it wrong first and a test caught it: the comparison is against the *stored* value, which the floor may have already moved later than the `t` that produced it. So a repeated abrupt deadline is still sooner than the stored floor and gets re-floored against the new now — a server that keeps signalling expiry holds a rolling minimum notice rather than pinning one date. A dated deadline, by contrast, settles: once stored it is not sooner than itself. Both are pinned by tests.

The record carries the build version because a deadline issued against one build says nothing about the next; `applies_to` is how a consumer checks.

## Also

`priority_offline_complete` is now named as a recognized no-op. WA Web parses it into its info-bulletin union and then has **no `case` for it** in the dispatch switch, so it is a marker the client is expected to do nothing with — naming it stops it reading as a gap in the warn log.

Not included, deliberately: `tos` (notice ids) and `recovery_nonce` (CTWA business access-token nonce, `use_case` 547) are UI-surface concerns this library does not own, and `edge_routing`'s `dns_domain`, which would change handshake routing and deserves its own change.

## Storage

One nullable `TEXT` column holding the record as JSON, with a migration. A column per field buys nothing here — nothing queries or orders by a version triple. An undecodable value reads as "no deadline" rather than failing the whole device load; the next `<ib>` restates it.

## Tests

Twelve. Eight pure ones on the decision rule (first deadline, abrupt deadline floored, later deadline rejected, equal deadline rejected, sooner deadline accepted, abrupt repeat rolls forward, dated repeat settles, version scoping) and four on the stanza path (persisted and announced, unchanged deadline stays silent so a consumer alerting on this event doesn't fire every reconnect, missing `t` withdraws, withdrawing something never held is silent).

## Verification

- `cargo fmt --all`
- `cargo test -p whatsapp-rust --lib` — 1719 passed, 0 failed
- `cargo test -p wacore --lib` — 1463 passed, 0 failed
- `cargo test -p whatsapp-rust-sqlite-storage --lib` — 78 passed, 0 failed
- `cargo clippy -p wacore -p whatsapp-rust -p whatsapp-rust-sqlite-storage --all-targets -- -D warnings` — clean

`EventKind::ClientExpirationChanged` is appended, and the build-time tripwire now points at it as the last variant.

---
_Generated by [Claude Code](https://claude.ai/code/session_012wUgndtwXJDG9X8Maacc2J)_

### DIFF
--- a/src/handlers/ib.rs
+++ b/src/handlers/ib.rs
@@ -1,6 +1,8 @@
 use super::traits::StanzaHandler;
 use crate::client::Client;
-use crate::types::events::{DirtyState, Event, EventKind, OfflineSyncPreview};
+use crate::types::events::{
+    ClientExpirationChanged, DirtyState, Event, EventKind, OfflineSyncPreview,
+};
 use async_trait::async_trait;
 use futures::FutureExt;
 use log::{debug, info, warn};
@@ -8,6 +10,8 @@ use std::sync::Arc;
 use wacore::appstate::patch_decode::WAPatchName;
 use wacore::iq::dirty::{DirtyBit, DirtyType};
 use wacore::stanza::wire_tags::StanzaTag;
+use wacore::store::commands::DeviceCommand;
+use wacore::store::device::{ClientExpirationUpdate, ServerClientExpiration};
 
 /// Handler for `<ib>` (information broadcast) stanzas.
 ///
@@ -275,6 +279,14 @@ async fn handle_ib_impl(client: Arc<Client>, node: &wacore_binary::NodeRef<'_>) 
                     }))
                     .detach();
             }
+            "client_expiration" => {
+                handle_client_expiration(&client, child).await;
+            }
+            // WA Web parses this into its info-bulletin union and then has no
+            // `case` for it in the dispatch switch, so it is a marker the
+            // server sends and the client is expected to do nothing with.
+            // Named here so it does not read as a gap.
+            "priority_offline_complete" => {}
             "thread_metadata" => {
                 // Present in some sessions; safe to ignore for now until feature implemented.
                 debug!("Received thread metadata, ignoring for now.");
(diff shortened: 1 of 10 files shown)


jlucaso1:
All three findings were valid. Fixed in `2ad3ee6`.

## Cross-build suppression (Greptile P1, CodeRabbit 🟠, Codex P2)

Correct, and it is the sharper kind of mistake: I added `applies_to` in this same PR *precisely* for this, then didn't use it at the point that matters. An upgrade left the previous build's record in place, and its nearer date read as "sooner than the new one", so the only notice that applied to the running build was rejected as `Unchanged`.

Scoped at a single point rather than inline in the comparison:

```rust
pub fn held_for(current: Option<&Self>, version: (u32, u32, u32)) -> Option<&Self> {
    current.filter(|held| held.applies_to(version))
}
```

`decide` consults that first, so the forward-only rule and the floor both operate on a deadline that actually describes this build. The withdrawal path scopes the same way, with one distinction worth stating: a leftover record from another build is *dropped* (it describes a build no longer running) but **silently**, because this build never had a deadline to withdraw and announcing one would report a change that did not happen.

Codex's alternative — clear the record atomically inside `SetAppVersion` — also works, but it puts expiration knowledge into the version command and still leaves the window between a version change and the next write. Scoping at the point of use has no window.

On WA Web parity: WA Web *does* compare version-blind here, because its stored record is read back by a consumer that checks `appVersion` itself. Here the record is what the comparison consults, so the scoping has to move to the point of use. Noted in the code so the divergence doesn't read as an oversight.

## An unusable `t` was treated as a withdrawal (Codex P2, CodeRabbit 🟠)

Codex's framing is the more complete one. `optional_u64` returns `None` for an absent attribute *and* for one it cannot parse, and this path read `None` as a withdrawal — so a malformed `t` would clear a deadline the server never retracted. CodeRabbit's overflow case is the same bug at its worst: `9223372036854775808 as i64` wraps to `i64::MIN`, which the minimum-notice floor then turns into a plausible-looking deadline.

Presence is now read first, and the value parsed straight to `i64`, which rejects a non-numeric value and one past `i64::MAX` in one step. The sign check is WA Web's `min = 0` from `attrIntRange(node, "t", 0, undefined)`.

```rust
let raw = child.attrs().optional_string("t");
let t = match raw.as_deref() {
    None => None,                       // genuine withdrawal
    Some(raw) => match raw.parse::<i64>() {
        Ok(t) if t >= 0 => Some(t),
        _ => { warn!(...); return; }    // unusable: touch nothing
    },
};
```

## String-literal child dispatch (CodeRabbit 🟠)

Fair, and it cites the repo's own rule. `InfoBulletinType` now binds WA Web's frozen `INFO_TYPE` list and the handler dispatches on `try_from(child.tag.as_ref())`, so a renamed tag is a one-attribute change instead of an arm that still compiles and never matches. It converts the pre-existing arms too, so the file no longer mixes both styles.

Hand-written rather than generated, and the module says why: whatspec's `notif` document describes the tags a *stanza* arrives under, not the children inside one, and `srvreq` carries a parser for `client_expiration` alone. The rest exists only in WA Web's dispatcher, so there is nothing to bind against — the `schemas_unlisted.rs` situation, not a gap in the emitter. Open enum with a `Unknown` default, so an unrecognized bulletin is still logged and skipped.

## Tests

Four added (16 total):
- `an_old_builds_deadline_does_not_suppress_the_running_ones` — asserts the new deadline is the *later* one, so the test fails without the scoping
- `scoping_does_not_weaken_the_forward_only_rule`
- `an_unusable_t_leaves_the_deadline_alone` — `9223372036854775808`, `-1`, `soon`, `""`; each must leave the stored deadline untouched and announce nothing

## Verification

`cargo test -p whatsapp-rust --lib` → 1720 passed; `-p wacore --lib` → 1465 passed; `-p whatsapp-rust-sqlite-storage --lib` → 78 passed; `cargo clippy -p wacore -p whatsapp-rust -p whatsapp-rust-sqlite-storage --all-targets -- -D warnings` → clean.

Semver Checks is the pre-existing advisory failure (`continue-on-error`, red on `main`, `waproto`-only findings).

---

PR #1320, opened 2026-08-18
perf(voip): make the MLow encoder attributable per stage, then cut what that showed

jlucaso1:
## Summary

`mlow_encode` is one bench row, so a change inside the codec moves it without saying which stage moved. This adds per-stage rows first, then makes only the changes those rows can measure.

**The `golden.rs` checksum is the axis.** It pins the validated codec output byte-exactly, and the encoder front-end is additionally pinned against the `smpl` C reference through fixtures (`fe_dump.json`, `lsf_quant_io.json`, `pitchio_ground_truth.json`) that are **not reproducible in this repository** — `testdata/PROVENANCE.md` documents each as the output of a C harness linked against the proprietary reference, and `inbound_capture_frames.json` explicitly as "NOT Rust-reproducible". So the decision recorded up front: **this branch does not change floating-point arithmetic**, and every change is accepted only if `GOLDEN_CHECKSUM` and `GOLDEN_FRAMES_CHECKSUM` both hold. They do.

While wiring the rows into CI, a second finding: **the VoIP benchmark was never running in CodSpeed at all.** Fixed here too, and verified — see *Validation*.

Instruction counts below are callgrind `Ir` and are deterministic; wall times are the **minimum** per row and are not reliable on this machine below ~10%. Where the two disagree, trust the instruction count.

## Changes

**Bench: nine stage rows (`codec_stages`), driving production functions through a new `stage_bench` harness** behind a `bench-internals` feature. The harness builds every table, twiddle set and pooled buffer in `Stages::new` and primes the cross-frame state with real frames, so a timed body is per-frame work only. Two setup-vs-body traps surfaced while building it:

- `Stages::new` runs real encodes. Built per sample, divan's calibration run timed it — putting a **3.6 ms** outlier on a **2.3 µs** row. The harness is now built once, before `divan::main`.
- The analysis path never calls the entropy coder, so the range-coder CC tables were cold at the first timed body: one sample paid **3.5 ms and ~1000 allocations** for a table build a live stream pays once.

The `voip_benchmark` target still requires only `voip-mlow`; `codec_stages` is `#[cfg]`-gated *inside* the bench. Listing `bench-internals` in `required-features` would make `cargo bench --features voip-mlow --bench voip_benchmark` silently run nothing — the same cargo behaviour this PR fixes for CodSpeed below.

**Every row varies its inputs the way production does**, which is what makes the ×3 below a frame rather than three copies of one path. Each stage that runs once per internal frame carries a captured input set per frame — its own `lpcbuf` window and the `a`/NLSF derived from it, its own `f2`, its own post-`build_ltp_buf` `ltp_buf`, its own `predcoefs`/residual/lags/perceptual weights — and the committed envelope is threaded forward as production threads it (frame *f*'s `brec` becomes frame *f+1*'s `prev_lsfq` **and** `prev_nlsf`, feeding the quantizer, the envelope reconstruction and the CELP interpolation alike). The LPC root search, pitch survivor blocks, LSF refinement early-exit and CELP fixed-codebook search are all data-dependent, so a frozen input would pin them to one path.

**Known limitation, measured rather than assumed.** The stateful rows advance the same cross-frame state production advances while replaying a fixed three-frame cycle. `pitch_search` and `perc_model_frame` are **exactly 3-periodic from the first iteration** — pinned by `stage_rows_are_three_periodic`. `celp_subframes_frame` is not: its pulse counts wander (`28,19,31,…` early against `26,16,33,…` after 20 iterations) because the excitation history evolves against a repeating residual. Its timing spread stays inside 1%, so the fixed-size beam search dominates, but that row is a steady-state estimate rather than a reproduction of one production frame. Restoring state per iteration would put the restore inside the timed body — the exact failure this harness exists to avoid — so it is documented instead. `analyze_frame` carries no such caveat, which is why stage rows are reported as shares *of it*.

**CI: the `core` CodSpeed shard now passes `--features voip-mlow,bench-internals`.** It previously passed none, and both `voip_benchmark` and `sframe_varint_benchmark` carry `required-features` — which cargo resolves by **silently skipping** the target, not by erroring. Of `main`'s 288 CodSpeed benchmarks, **not one came from either file**. The whole VoIP media plane had never been regression-gated.

**Four production changes, none touching a floating-point operation:**

- **FFT recursion plan.** `n` is fixed per stream (512 for LPC analysis, 576 for the perceptual model), so the whole `(n, p, m)` chain is known at init. The recursion was re-deriving `p` with `smallest_factor` and re-finding its twiddle table by **linear scan** at every node; both are now a depth index into a precomputed level plan. (The rewrite initially dropped the `n == 1` arm — a length-1 plan builds no base table, so the fallback indexed an empty one. Caught in review, guard restored, pinned by `fft_of_length_one_is_the_identity`.)
- **`CelpEncoder::encode_subframe`** allocated a `vec[0.0f32; 160]` scratch per rate *inside* its rate loop. Now pooled and hoisted out. (Eliminating the twin entirely is bit-identical and removes a copy — but measured **12.6% slower** at 0.36% fewer instructions, so the twin stays.)
- **`smpl_nlsf2a`** allocated four `Vec`s sized by the fixed LPC order of 16; three become stack arrays.
- **`smpl_lsf_quant::get_maxi_k`** allocated its `used` mask per call; `n ≤ 17`, so it goes on the stack — the same fix `smpl_celp::smpl_get_maxi_k` already carries.

## Cost

**Where the encoder's work goes — by instruction count**, deterministic and what CodSpeed gates:

| stage | `Ir` per call | ×/frame | per frame | share of `analyze_frame` |
| --- | ---: | ---: | ---: | ---: |
| **FFT total** | | | **6,154,596** | **43.5%** |
| — `fft576_roundtrip` | 825,768 | 6 | 4,954,608 | 35.0% |
| — `fft512_forward` | 399,996 | 3 | 1,199,988 | 8.5% |
| `celp_subframes_frame` | 1,504,085 | 3 | 4,512,255 | 31.9% |
| `pitch_search` | 478,718 | 3 | 1,436,154 | 10.2% |
| **`analyze_frame`** | **14,140,131** | 1 | 14,140,131 | 100% |

The FFT rows are subsets of `lpc_front_end` and `perc_model_frame`, not additional to them.

Wall time, latest run, minimum per row against `analyze_frame` = 1.786 ms: perc 194.0 µs ×3 = 32.6%, CELP 176.9 µs ×3 = 29.7%, pitch 55.7 µs ×3 = 9.4%, LPC 55.0 µs ×3 = 9.2%, LSF 7.22 µs ×3 = 1.2%, entropy 2.30 µs = 0.1% — 82.2% accounted; FFT totals 33.5%. **Across runs the FFT's wall-time share ranged 31.7–35.0%.** An earlier revision of this description claimed it "never moved", which was wrong: the FFT rows and `analyze_frame` respond differently to machine state. The instruction share is the figure to use; the gap (43.5% vs ~33%) is coherent, since the FFT is dense float at high IPC and retires more instructions per unit time than the rest of the encoder.

**What the changes bought** (`Ir` per stage call, differenced across two iteration counts so startup and setup cancel):

| row | base | after | Δ |
| --- | ---: | ---: | ---: |
| `mlow_encode` | 14,133,063 | 14,027,712 | **−0.75%** |
| `lsf_quantize` | 81,726 | 77,482 | **−5.19%** |
| `fft512_forward` | 403,317 | 395,643 | **−1.90%** |
| `lpc_front_end` | 449,855 | 442,174 | −1.71% |
| `fft576_roundtrip` | 820,112 | 817,317 | −0.34% |
| `perc_model_frame` | 1,675,003 | 1,669,418 | −0.33% |
| `celp_subframes_frame` | 1,518,418 | 1,514,996 | −0.23% |
| `pitch_search` *(control — untouched then)* | 530,851 | 530,856 | **+0.00%** |

Allocations (identical `min`/`median`/`max` — deterministic):

| row | allocs before | allocs after | bytes before | bytes after |
| --- | ---: | ---: | ---: | ---: |
| `mlow_encode` | 633 | **534** (−15.6%) | 519.8 KB | **499.2 KB** (−4.0%) |

**Wall time: no claim.** Interleaved A/B moved the byte-identical `pitch_search` control by **+2.6% to +8.3%**, and `lsf_quantize` — code unchanged between two builds — swung **−4.6% to +11.0%** on binary layout alone. This optimizes instruction and allocation count. The one wall-time signal that cleared the noise floor was a regression (the CELP twin-buffer removal, **+12.6%** at fewer instructions, from a loop-carried memory dependency), fixed by pooling the twin instead.

## Checked and not changed

- **Real-input FFT and specialized radix kernels — characterized, then refused.** `rfft_forward_ordered_sc` fills `cin[i].im = 0.0` and runs a full complex transform over real input; a real-input FFT via a complex N/2 plus a twiddle pass is roughly half that work, and specialized radix-2/3/4 kernels would remove the recursion and dispatch on top. Against the measured **43.5% of encoder instructions** that is worth on the order of **17–22% of the encoder** — by far the largest item found. Both change floating-point operation order by construction, which flips both golden checksums, invalidates `twiddle_table_is_bit_exact`, and puts `front_end_a_matches_c`'s `|dA| < 5e-3` tolerance — calibrated against *this* FFT's error profile — back in question. Revalidating means re-running C harnesses linked against the proprietary `smpl` reference, which this repository does not contain and cannot reproduce.
- **The FFT's per-node overhead was worth ~2%, and that is itself the finding.** Removing 5361 trial divisions and 5361 linear table scans per frame bought only −1.90% on `fft512_forward`. The rest is butterfly arithmetic — **128,256 complex multiplies per frame** — which is what IPC 4.83 predicts, and what makes *doing less arithmetic* the only remaining FFT lever.
- **`fft_rec`'s callers and call counts, previously unknown**, are measured: **5361 recursion nodes per encoded frame** across **15 top-level transforms** — 3 forward at N=512 and 6 forward + 6 inverse at N=576. The 512 chain is pure radix-2 (511 nodes); the 576 chain is `2^6 · 3^2` (319 nodes) and is the only one reaching the radix-3 levels and the O(n²) prime base case.
- **The decoder runs zero FFTs.** `fft_rec` is called 0 times per decode. The external profile's `fft_rec` share mixed both call directions and both codec ends; here all of it is encode-side. `mlow_decode` is **47.9 µs** against `mlow_encode`'s ~**1.8 ms** — the decode side is not worth optimizing.
- **The external profile's "~450 allocations and ~250 KB per codec invocation" is confirmed in order of magnitude and corrected in attribution**: 633 allocs / 519.8 KB for encode, 51 / 25.9 KB for decode. The external divisor mixed both ends and included warmup.
- **No data-layout or prefetch change is proposed.** At IPC 4.83 there is no stall to recover. CodSpeed's valgrind breakdown corroborates independently: on `mlow_encode`, cache-miss time is **1.3%** of instruction-execution time.
- **`param_decode_match.rs`, the twiddle tables, and the wire format are untouched**, as is `[patch.crates-io]` for `curve25519-dalek`.
- **No `unsafe`, no global state, no unbounded cache.** The FFT level plan is per-`FftScratch`, built at construction and sized by the fixed `n`.

## Validation

- `cargo fmt --all`; `cargo clippy -p wacore --features voip-mlow,bench-internals --all-targets -- -D warnings` clean; `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features` clean.
- `cargo test -p wacore --features voip-mlow,bench-internals --release --lib voip::mlow` — **84 passed, 0 failed**, including `golden_roundtrip_no_regression` (decode-output *and* encoder-bitstream checksums), `twiddle_table_is_bit_exact`, `fft_of_length_one_is_the_identity`, `stage_rows_are_three_periodic`, `front_end_a_matches_c`, `lsf_quant_matches_c`, `pitch_estimator_matches_c_ground_truth`, `nrgres_fcbg_match_c_reference`.
- The checksums holding across all four production changes is the proof that none altered arithmetic. Its limit was demonstrated in review: it says nothing about paths the corpus never exercises, which is how the length-1 FFT regression survived until a reviewer read the code.
- **Three build modes verified, by listing the rows rather than asserting.** The library under plain `voip-mlow` compiles no `stage_bench`. `cargo bench --features voip-mlow --bench voip_benchmark` builds and lists all the pre-existing rows (MLow, SRTP, SFrame, engine, framing) with no `codec_stages`. Adding `bench-internals` adds the nine stage rows.
- **The CodSpeed fix is verified, not assumed.** The branch run returns real Simulation results with flamegraphs for `wacore/benches/voip_benchmark.rs::mlow_encode` and `…::codec_stages::perc_model_frame`. Both were absent from every previous run.
- **CodSpeed baseline:** every VoIP row it uploads is new, so the first run establishes a baseline rather than comparing against one. A green "no regressions" here is not a passed comparison.
- **Semver Checks (informational) is red and is not from this PR.** Every failure is `waproto` generated-protobuf churn (`AIRichResponseContentItem` renames, `EncryptMessageOutput.message_key`, new variants on exhaustive enums). No `wacore` item appears; the job is informational by design.

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -43,8 +43,15 @@ jobs:
       # (codspeed.io/docs -> sharded-benchmarks).
       matrix:
         shard:
+          # `features` is required here, not optional: `voip_benchmark` and
+          # `sframe_varint_benchmark` carry `required-features`, and a bench
+          # target whose required features are off is SILENTLY SKIPPED by cargo
+          # rather than erroring. Without this the entire VoIP media plane --
+          # MLow encode/decode, E2E-SRTP, SFrame, the codec stage rows -- built
+          # nowhere and uploaded nothing, so none of it was regression-gated.
           - name: core
             packages: -p wacore -p wacore-noise
+            features: --features voip-mlow,bench-internals
           - name: proto-signal
             packages: -p wacore-binary -p wacore-libsignal -p wacore-appstate
           # The client crate's own benchmarks. Its fixture needs crate-private
--- a/wacore/Cargo.toml
+++ b/wacore/Cargo.toml
@@ -153,7 +153,12 @@ harness = false
 [[bench]]
 name = "voip_benchmark"
 harness = false
-required-features = ["voip-mlow", "bench-internals"]
+# Only `voip-mlow` here on purpose: cargo SILENTLY SKIPS a bench target whose required features are
+# off, so listing `bench-internals` too would make the previously-working
+# `cargo bench --features voip-mlow --bench voip_benchmark` quietly run nothing -- dropping the MLow,
+# SRTP, SFrame and framing rows along with the stage rows. The `codec_stages` module is `#[cfg]`-gated
+# inside the bench instead, so it is the only part that needs the extra feature.
+required-features = ["voip-mlow"]
 
 [[example]]
 name = "voip_profile"
--- a/wacore/Cargo.toml
+++ b/wacore/Cargo.toml
@@ -43,6 +43,10 @@ voip-mlow = ["voip"]
 # Heap profiling of the codec hot paths via the `voip_profile` example (dhat as global allocator).
 # Dev/profiling only; off by default and not pulled into normal builds.
 dhat-heap = ["dep:dhat"]
+# Per-stage MLow codec bench surface (`voip::mlow::stage_bench`), so `voip_benchmark` can attribute
+# encoder CPU to a stage. Pure bench scaffolding: `#[doc(hidden)]` would hide it from rustdoc but
+# still compile it into every consumer, so it is a feature. Dev/bench only; off by default.
+bench-internals = []
 
 [dependencies]
 aes = { workspace = true }
@@ -149,7 +153,7 @@ harness = false
 [[bench]]
 name = "voip_benchmark"
 harness = false
-required-features = ["voip-mlow"]
+required-features = ["voip-mlow", "bench-internals"]
 
 [[example]]
 name = "voip_profile"
--- a/wacore/benches/voip_benchmark.rs
+++ b/wacore/benches/voip_benchmark.rs
@@ -244,7 +244,9 @@ mod codec_stages {
     }
 
     /// Whole analysis half of an encode (1x/frame): VAD, high-pass, and the per-internal-frame LPC /
-    /// perc / pitch / LSF / CELP chain. With `entropy_encode` this accounts for all of `mlow_encode`.
+    /// perc / pitch / LSF / CELP chain. With `entropy_encode` this is the codec work -- `mlow_encode`
+    /// additionally sanitizes its 960 input samples and allocates a fresh output buffer, which the
+    /// `mlow_encode` vs `mlow_encode_reused_output` pair isolates.
     #[divan::bench]
     fn analyze_frame(bencher: Bencher) {
         run(bencher, |s| s.analyze_frame() as f64);
--- a/wacore/benches/voip_benchmark.rs
+++ b/wacore/benches/voip_benchmark.rs
@@ -28,6 +28,7 @@ fn main() {
     // bench function does before its closure, and `Stages::new` runs three real encodes: left inside,
     // it sized `entropy_encode`'s sample loop to a single iteration and put one 3.6 ms /
     // 2000-allocation outlier on a row whose real cost is 1.9 us and zero allocations.
+    #[cfg(feature = "bench-internals")]
     codec_stages::warm();
     divan::main();
 }
@@ -218,6 +219,7 @@ fn mlow_decode(bencher: Bencher) {
 // Each row's doc gives its per-frame multiplicity, which is what makes the rows sum: reading them
 // against `mlow_encode` needs `stage_time * calls_per_frame`, not `stage_time` alone. Decode has no
 // row here on purpose -- it runs none of these stages (notably, zero FFTs).
+#[cfg(feature = "bench-internals")]
 mod codec_stages {
     use super::*;
     use std::sync::{Mutex, OnceLock};
--- a/wacore/benches/voip_benchmark.rs
+++ b/wacore/benches/voip_benchmark.rs
@@ -24,6 +24,11 @@ use wacore::voip::sframe::SframeSession;
 static ALLOC: divan::AllocProfiler = divan::AllocProfiler::system();
 
 fn main() {
+    // Build the codec-stage harness BEFORE divan starts. Divan's calibration run times whatever the
+    // bench function does before its closure, and `Stages::new` runs three real encodes: left inside,
+    // it sized `entropy_encode`'s sample loop to a single iteration and put one 3.6 ms /
+    // 2000-allocation outlier on a row whose real cost is 1.9 us and zero allocations.
+    codec_stages::warm();
     divan::main();
 }
 
@@ -200,6 +205,107 @@ fn mlow_decode(bencher: Bencher) {
             *i += 1;
             black_box(dec.decode(black_box(p.as_slice())))
         });
+}
+
+// --- Codec stages: which part of the encoder a change actually moved ---
+//
+// `mlow_encode` is one row, so a codec-internal change moves it without saying WHICH stage moved.
+// These rows call the same production functions the analyzer calls, one stage per row, via the
+// `stage_bench` harness (`wacore::voip::mlow::stage_bench`). The harness builds its tables, twiddles
+// and pooled buffers in `Stages::new` and primes the cross-frame state with real frames, so a timed
+// body is per-frame work only -- what a call pays every 60 ms, never what the stream resolves once.
+//
+// Each row's doc gives its per-frame multiplicity, which is what makes the rows sum: reading them
+// against `mlow_encode` needs `stage_time * calls_per_frame`, not `stage_time` alone. Decode has no
+// row here on purpose -- it runs none of these stages (notably, zero FFTs).
+mod codec_stages {
+    use super::*;
+    use std::sync::{Mutex, OnceLock};
+    use wacore::voip::mlow::stage_bench::Stages;
+
+    /// One harness for every row, built once by [`warm`] before divan starts. Divan runs rows
+    /// sequentially, so the lock is uncontended; its ~20 ns is inside every row's timed body, which
+    /// matters only for `entropy_encode` (~1% of its 1.9 us) and is why the harness is shared rather
+    /// than rebuilt.
+    ///
+    /// Sharing is also the faithful shape: the stateful stages (perc history, CELP ACB/ZIR, pitch
+    /// predictor) advance exactly as they do in a live call, and each row runs enough iterations to
+    /// settle into its own steady state, so no row depends on which ran before it.
+    static STAGES: OnceLock<Mutex<Stages>> = OnceLock::new();
+
+    /// Construct the harness outside any timed region. Called from `main` before `divan::main`.
+    pub fn warm() {
+        STAGES.get_or_init(|| Mutex::new(Stages::new()));
+    }
+
+    fn run(bencher: Bencher, mut stage: impl FnMut(&mut Stages) -> f64) {
+        let cell = STAGES.get().expect("warm() ran in main");
+        bencher.bench_local(move || black_box(stage(&mut cell.lock().expect("bench harness"))));
+    }
+
+    /// Whole analysis half of an encode (1x/frame): VAD, high-pass, and the per-internal-frame LPC /
+    /// perc / pitch / LSF / CELP chain. With `entropy_encode` this accounts for all of `mlow_encode`.
+    #[divan::bench]
+    fn analyze_frame(bencher: Bencher) {
+        run(bencher, |s| s.analyze_frame() as f64);
+    }
+
+    /// Range coder writing the analyzed parameters to the wire (1x/frame). The other half of
+    /// `mlow_encode`, and the half that does NOT dominate.
+    #[divan::bench]
+    fn entropy_encode(bencher: Bencher) {
+        run(bencher, |s| s.entropy_encode() as f64);
+    }
+
+    /// LPC front-end for one internal frame (3x/frame): window, forward FFT, power spectrum, DCT
+    /// autocorrelation, Levinson, bandwidth expansion, A->NLSF.
+    #[divan::bench]
+    fn lpc_front_end(bencher: Bencher) {
+        run(bencher, |s| s.lpc_front_end() as f64);
+    }
+
+    /// The bare forward real FFT at the LPC size (N=512, pure radix-2), 3x/frame. Isolated from
+    /// `lpc_front_end` so an FFT-local change is attributable without the DCT/Levinson around it.
+    #[divan::bench]
+    fn fft512_forward(bencher: Bencher) {
+        run(bencher, |s| s.fft512_forward() as f64);
+    }
+
+    /// Forward + inverse real FFT at the perceptual-model size (N=576 = 2^6 * 3^2, mixed radix 2/3),
+    /// 6x/frame -- 12 of the frame's 15 FFTs, and the only path that reaches the radix-3 levels and
+    /// the O(n^2) prime base case.
+    #[divan::bench]
+    fn fft576_roundtrip(bencher: Bencher) {
+        run(bencher, |s| s.fft576_roundtrip() as f64);
+    }
+
+    /// Perceptual model for one internal frame (3x/frame): two `smpl_perc_model` calls, i.e. two
+    /// `fft576_roundtrip`s plus the windowing and the masking smooth around them.
+    #[divan::bench]
+    fn perc_model_frame(bencher: Bencher) {
+        run(bencher, |s| s.perc_corrs_frame() as f64);
+    }
+
+    /// Multi-stage pitch estimator for one internal frame (3x/frame).
+    #[divan::bench]
+    fn pitch_search(bencher: Bencher) {
+        run(bencher, |s| s.pitch_search() as f64);
+    }
+
+    /// LSF vector quantizer plus the envelope reconstruction and NLSF->A after it (3x/frame). The
+    /// row that covers `get_maxi_k` and `smpl_nlsf2a`.
+    #[divan::bench]
+    fn lsf_quantize(bencher: Bencher) {
+        run(bencher, |s| s.lsf_quantize() as f64);
+    }
+
+    /// CELP excitation encoder for one internal frame (3x/frame): perceptual weighting plus four
+    /// `encode_subframe` calls, so `encode_subframe` itself runs 12x/frame. The fixed-codebook pulse
+    /// search inside it is the single largest stage of the encoder.
+    #[divan::bench]
+    fn celp_subframes_frame(bencher: Bencher) {
+        run(bencher, |s| s.celp_subframes_frame() as f64);
+    }
 }
 
 // --- Crypto + framing: the seam where avoidable per-packet allocations live ---
--- a/wacore/src/voip/mlow/analysis.rs
+++ b/wacore/src/voip/mlow/analysis.rs
@@ -1351,7 +1351,11 @@ pub mod stage_bench {
         hp: Vec<f32>,
         hp_pitch_hist: Vec<f32>,
         /// LPC front-end inputs, captured for internal frame 0.
-        lpcbuf: [f32; SMPL_LPC_BUF_LEN],
+        /// The `lpcbuf` window each internal frame analyzes. Production advances the start by
+        /// `f * SMPL_INTF_LEN`, and the coefficients that fall out drive the data-dependent root
+        /// search in `smpl_a2nlsf_16`, so cycling only the long/short window flag would still be
+        /// three passes over one signal.
+        lpcbufs: Vec<[f32; SMPL_LPC_BUF_LEN]>,
         /// Zero-padded windowed buffer + output, for the bare FFT row.
         fft_in: Vec<f32>,
         fft_out: Vec<f32>,
@@ -1453,7 +1457,7 @@ pub mod stage_bench {
                 pcm_at: 0,
                 hp,
                 hp_pitch_hist,
-                lpcbuf,
+                lpcbufs: Vec::new(),
                 fft_in,
                 fft_out,
                 fft_scratch,
@@ -1487,7 +1491,9 @@ pub mod stage_bench {
                 let start = SMPL_LPC_HIST_LEN - SMPL_LPC_PRE + f * SMPL_INTF_LEN;
                 lpcbuf_f.copy_from_slice(&s.es.hp_full[start..start + SMPL_LPC_BUF_LEN]);
                 let windowed_f = smpl_window_lpc20(&lpcbuf_f, f < 2);
-                let (_, f2_f) = smpl_lpc_analyze_with_f2(&windowed_f, &mut s.fft_scratch);
+                let (a_f, f2_f) = smpl_lpc_analyze_with_f2(&windowed_f, &mut s.fft_scratch);
+                let nlsf_f = smpl_a2nlsf_16(&a_f);
+                s.lpcbufs.push(lpcbuf_f);
 
                 let perc_corrs: Vec<Vec<f32>> = s.with_ctx(f, [[0.0; 2]; SMPL_SUBFR_COUNT], |cs| {
                     compute_perc_corrs(cs).into()
@@ -1512,8 +1518,8 @@ pub mod stage_bench {
                     block_lags[sf] = [pr.lags[2 * sf], pr.lags[2 * sf + 1]];
                 }
                 let fe = FrontEndLsf {
-                    a: s.a,
-                    nlsf: s.nlsf,
+                    a: a_f,
+                    nlsf: nlsf_f,
                     prev_lsfq: &s.prev_lsfq,
                     prev_voiced: s.es.prev_voiced,
                     intf: f,
@@ -1622,7 +1628,7 @@ pub mod stage_bench {
         /// one, and the two differ in window-generation work.
         pub fn lpc_front_end(&mut self) -> f32 {
             let intf = self.next_intf(Self::ROW_LPC);
-            let windowed = smpl_window_lpc20(&self.lpcbuf, intf < 2);
+            let windowed = smpl_window_lpc20(&self.lpcbufs[intf], intf < 2);
             let (a, _f2) = smpl_lpc_analyze_with_f2(&windowed, &mut self.fft_scratch);
             let nlsf = smpl_a2nlsf_16(&a);
             nlsf[0]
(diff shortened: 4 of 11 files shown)


jlucaso1:
All four review findings were valid and are fixed in c0bbcda. Two of them changed the reported numbers, so the *Cost* table in the description has been rewritten — flagging that here because the headline attribution moved.

**@coderabbitai** — both correct.
- `stage_bench` is now behind a `bench-internals` feature, gated at the module and the re-export, with `required-features` on the `voip_benchmark` target. You were right that `#[doc(hidden)]` was the wrong tool: it hides from rustdoc but still compiles into every consumer. I had modelled it on the `e2e_srtp`/`sframe` precedent, but those are production modules that merely stay undocumented, whereas this one is pure scaffolding — different thing, different mechanism. Both build modes are verified.
- The `WARMUP_FRAMES` doc/const contradiction was a stale comment from when the constant was 2. Fixed.

**@chatgpt-codex-connector** — all four correct, and the `intf` ones were the substantive catch. Every stage production runs once per internal frame behaves differently at `intf` 0, 1 and 2 — LPC window length (`f < 2`), the perc model's `is_last` trailing window, the conditional LSF quantizer's `intf > 0` requirement, and the pitch predictor's `reset_cond` after frame 2. Each row pinned itself to one index while the attribution multiplied it by three, so the table was reporting three copies of one path instead of a frame. Every such row now cycles 0, 1, 2 (and `pitch_search` resets the predictor at each packet boundary), and `analyze_frame` cycles eight phase-shifted frames like `mlow_encode` does.

That was not cosmetic:

| row | before | after | |
| --- | ---: | ---: | --- |
| `pitch_search` | 51.0 µs | **73.2 µs** | understated by 30% — every call was taking the conditional path |
| `perc_model_frame` | 194.6 µs | **208.2 µs** | understated by 7% — never reached the short trailing window |
| `lsf_quantize` | 7.86 µs | 7.13 µs | now includes frame 0's non-conditional quantizer |

Net effect on the conclusion: the accounted share of `analyze_frame` went from 82% to 84.4%, and pitch search moved from 8.9% to 11.9% of the encode. The FFT total is unchanged at ~35%, so the main finding holds — but it holds on better numbers.

One thing I did *not* take, for the record: the earlier `celp_subframes_frame` change that eliminated the rate-loop scratch buffer entirely. It is bit-identical and removes both an allocation and a copy, but it measured **+12.6% wall** at 0.36% *fewer* instructions — putting the fixed-codebook synthesis on the same memory the rest of the iteration reads and rewrites creates a loop-carried dependency. Pooling the twin instead keeps the codegen and still removes the allocation. Details in the description under *Cost*.

---

jlucaso1:
Round two. Two of the three findings were valid and are fixed in 71a120e; one is incorrect and I'm not applying it.

**@coderabbitai — the `--features` finding is wrong, and the merge-risk flag on it should come off.**

The claim is that `--features voip-mlow,bench-internals` alongside `-p wacore -p wacore-noise` "makes Cargo fail before CodSpeed uploads benchmarks." That is not cargo's behavior. Unqualified features are applied to whichever *selected* package declares them; cargo errors only when **none** of the selected packages has the feature. Your script established that `wacore-noise` lacks them, which is not the failure condition.

Verified twice, both directly:

```
$ cargo check -p wacore -p wacore-noise --features voip-mlow,bench-internals --benches
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 57.60s
EXIT=0
```

and against the real job — the CodSpeed run for `d923796` (commit `c809125`) reached `status: processing`, i.e. it built the benchmarks and ran them. Had the cargo invocation failed there would be no run at all.

The namespaced form is a defensible style preference, but I'd rather not churn CI config on a premise that doesn't hold — and the unqualified form matches how the sibling `client` shard is written.

**@chatgpt-codex-connector — both findings valid, both fixed.**

- *CELP inputs.* Correct, and it was the one that mattered. `celp_subframes_frame` cycled `intf` but handed `run_celp_subframes` the same captured predcoefs, residual, lags and perceptual weights every iteration. Since the fixed-codebook search is data-dependent — which this bench file already documents as the reason `mlow_encode` streams eight frames — the encoder's largest stage could settle onto a path the varied workload never takes. The harness now captures one input set per internal frame, each from the live state at that frame's own offset, so cycling the index cycles the data. Effect was small (167.2 µs vs 173.8 µs) but it was measuring the wrong thing for the right-looking number.
- *Wrapper coverage.* Also correct — I overclaimed. `analyze_frame` + `entropy_encode` is the *codec* work; `MlowEncoder::encode_into` additionally sanitizes its 960 input samples and `mlow_encode` allocates a fresh output `Vec` where the harness reuses one. Docs corrected, and they now point at the existing `mlow_encode` vs `mlow_encode_reused_output` pair, which already isolates the allocation half of that gap.

**Re-measured attribution** (5 runs, minimum per row, spreads shown — this replaces the table in the description):

| stage | min | ×/frame | per frame | share | spread |
| --- | ---: | ---: | ---: | ---: | --- |
| `perc_model_frame` | 196.9 µs | 3 | 591 µs | 33.3% | 196.9–197.4 |
| `celp_subframes_frame` | 167.2 µs | 3 | 502 µs | 28.2% | 167.2–170.0 |
| `pitch_search` | 74.4 µs | 3 | 223 µs | 12.6% | 74.4–76.0 |
| `lpc_front_end` | 54.9 µs | 3 | 165 µs | 9.3% | 54.9–55.1 |
| `lsf_quantize` | 7.03 µs | 3 | 21 µs | 1.2% | 7.0–7.1 |
| `entropy_encode` | 2.23 µs | 1 | 2 µs | 0.1% | 2.2–2.5 |
| accounted | | | 1504 µs | 84.7% | |
| **`analyze_frame`** | **1.776 ms** | 1 | 1776 µs | 100% | 1776–1815 |

| FFT subset | min | ×/frame | per frame |
| --- | ---: | ---: | ---: |
| `fft512_forward` | 40.3 µs | 3 | 121 µs |
| `fft576_roundtrip` | 80.9 µs | 6 | 485 µs |
| **FFT total** | | | **606 µs — 34.1%** |

The headline is unchanged across every revision of the harness: the FFT is about a third of the encoder, and it is the arithmetic, not the dispatch, that costs.

---

jlucaso1:
Round three, in 0912138. One of these is a bug I introduced, and one forces me to withdraw a number I posted earlier.

**@greptile-apps — the length-1 FFT finding is correct and was my regression.** Verified: `FftTwiddles::new(1)` splits nothing, so `levels` is empty, `base_n` stays 1 and `base` stays empty; the depth-indexed fallback then ran `for k in 0..1 { … bt[0] … }` against that empty table. A length-1 transform panicked where the pre-plan recursion returned `out[0] = x[0]`. Unreachable at the production sizes (512 and 576) and by every existing test, which is exactly why it slipped through. Guard restored and pinned with `fft_of_length_one_is_the_identity`, so the plan rewrite can't silently drop it again. Good catch — this is the kind of thing the checksum tests structurally cannot see.

**@chatgpt-codex-connector — the pitch-input finding is correct, and it invalidates a number I posted.** You were right that resetting the predictor fixed the *conditional mode* but not the *workload*: the row replayed one frozen `ltp_buf` and one frame's `f2`, while production runs `build_ltp_buf` before every estimator call and recomputes `f2` per internal frame. The harness now captures a post-`build_ltp_buf` snapshot plus that frame's own `f2` per internal frame, and the row replays the matching pair.

The consequence matters more than the fix: with faithful inputs **`pitch_search` reads 55.1 µs, not 73.2 µs**. So my earlier comment claiming this row had been "understating its stage by 30% (51.0 → 73.2 µs)" was wrong — the 73.2 µs came from the intermediate configuration that reset the predictor but froze the data. **I withdraw that claim.** The real correction from the original 51.0 µs is small; what the review actually fixed was representativeness, not magnitude.

**Re-measured** (5 runs, minimum per row, spreads shown):

| stage | min | ×/frame | per frame | share | spread |
| --- | ---: | ---: | ---: | ---: | --- |
| `perc_model_frame` | 193.8 µs | 3 | 581 µs | 33.2% | 193.8–194.2 |
| `celp_subframes_frame` | 167.9 µs | 3 | 504 µs | 28.8% | 167.9–169.3 |
| `pitch_search` | 55.1 µs | 3 | 165 µs | 9.4% | 55.1–56.1 |
| `lpc_front_end` | 54.9 µs | 3 | 165 µs | 9.4% | 54.9–55.0 |
| `lsf_quantize` | 7.26 µs | 3 | 22 µs | 1.2% | 7.3–7.3 |
| `entropy_encode` | 2.30 µs | 1 | 2 µs | 0.1% | 2.3–2.3 |
| accounted | | | 1439 µs | 82.2% | |
| **`analyze_frame`** | **1.750 ms** | 1 | 1750 µs | 100% | 1750–1791 |

| FFT subset | min | ×/frame | per frame |
| --- | ---: | ---: | ---: |
| `fft512_forward` | 40.3 µs | 3 | 121 µs |
| `fft576_roundtrip` | 79.6 µs | 6 | 478 µs |
| **FFT total** | | | **598 µs — 34.2%** |

Worth noting what has and hasn't moved across five revisions of this harness. Individual stage rows have shifted by up to 26% as each input-fidelity bug was fixed — which is a fair signal that stage-level numbers deserve the scrutiny they've been getting. **The FFT share has read 34.1%, 34.2%, 34.7% and 35.0% throughout.** That is the number the PR's conclusion rests on, and it has been stable against every change to how the rows are driven.

83 tests pass, both golden checksums hold, clippy and rustdoc clean.

---

jlucaso1:
Round four, in 5c0fa44. All three findings valid, all fixed.

**@coderabbitai — the `predcoefs` finding is the sharpest one on this PR so far.** You're right, and the way it's wrong is worse than it looks: the capture loop *computed* each frame's `a` via `smpl_lpc_analyze_with_f2`, then bound only `f2_f` and threw `a_f` away, so `FrontEndLsf` got frame 0's `a`/`nlsf` and `brec`/`predcoefs` came out identical for f = 0, 1, 2 — which is exactly the failure mode the field doc I'd just written claimed the per-frame capture existed to prevent. Fixed by binding `a_f` and deriving `nlsf_f`.

It measurably mattered: **`celp_subframes_frame` is 183.4 µs with per-frame coefficients, against 167.9 µs with frame 0's.** The row was ~9% cheap because it wasn't doing the frame's real work.

Test placement also fixed — the new test had been inserted between `twiddle_table_is_bit_exact` and the comment describing it.

**@chatgpt-codex-connector — the LPC buffer finding is correct too**, and same class: the row cycled the long/short window flag but windowed the single frame-0 `lpcbuf`, so it was three passes over one signal wearing two window shapes, while production advances the start by `f * SMPL_INTF_LEN` and the resulting coefficients feed the data-dependent root search in `smpl_a2nlsf_16`. As you noted, the setup loop was already constructing the per-frame buffers — they're now stored and replayed.

**Re-measured** (5 runs, minimum per row):

| stage | min | ×/frame | per frame | share | spread |
| --- | ---: | ---: | ---: | ---: | --- |
| `perc_model_frame` | 189.6 µs | 3 | 569 µs | 31.3% | 189.6–202.7 |
| `celp_subframes_frame` | 183.4 µs | 3 | 550 µs | 30.2% | 183.4–185.2 |
| `pitch_search` | 57.0 µs | 3 | 171 µs | 9.4% | 57.0–58.4 |
| `lpc_front_end` | 55.7 µs | 3 | 167 µs | 9.2% | 55.7–55.9 |
| `lsf_quantize` | 7.09 µs | 3 | 21 µs | 1.2% | 7.1–7.1 |
| `entropy_encode` | 2.24 µs | 1 | 2 µs | 0.1% | 2.2–2.3 |
| accounted | | | 1481 µs | 81.4% | |
| **`analyze_frame`** | **1.820 ms** | 1 | 1820 µs | 100% | 1820–1851 |

| FFT subset | min | ×/frame | per frame |
| --- | ---: | ---: | ---: |
| `fft512_forward` | 40.9 µs | 3 | 123 µs |
| `fft576_roundtrip` | 83.1 µs | 6 | 499 µs |
| **FFT total** | | | **621 µs — 34.1%** |

Worth stating plainly, since this is now four rounds of the same class of bug: **every one of these was a benchmark-fidelity defect, and none touched production code.** The stage shares have moved around as each was fixed — CELP 28.2 → 28.8 → 30.2%, pitch 12.6 → 9.4% — which is the honest cost of building attribution from captured inputs rather than a live encode, and a fair argument that per-stage numbers deserve exactly this scrutiny.

**The FFT share has read 34.1%, 34.1%, 34.2%, 34.7%, 35.0% across all five revisions.** It is the one quantity that has never moved, and it is the only one the PR's conclusion depends on — the decision to refuse the arithmetic-changing FFT work rests on the size of that share, not on any individual stage row.

83 tests pass, both golden checksums hold, clippy and rustdoc clean.

---

jlucaso1:
Round five, in 3160f85 — and a correction to a claim I made twice in this thread.

**I overstated the FFT share's stability.** I said it "never moved" across harness revisions, citing 34.1–35.0%. The latest run reads **31.7%**, because the machine got faster and the FFT rows dropped more than `analyze_frame` did (`fft512_forward` −10.6%, `analyze_frame` −5.8%). That is the same wall-time unreliability I documented for everything else in this PR, and I should not have exempted this number from it. The honest wall-time range is **31.7–35.0%**.

So I measured it with the deterministic instrument instead, on current code:

| | per frame | of `analyze_frame` |
| --- | ---: | ---: |
| `fft512_forward` | 399,996 `Ir` ×3 = 1,199,988 | |
| `fft576_roundtrip` | 825,768 `Ir` ×6 = 4,954,608 | |
| **FFT total** | **6,154,596 `Ir`** | **43.5%** |
| `celp_subframes_frame` | 1,504,085 ×3 = 4,512,255 | 31.9% |
| `pitch_search` | 478,718 ×3 = 1,436,154 | 10.2% |
| `analyze_frame` | 14,140,131 `Ir` | 100% |

**By instructions the FFT is 43.5% of the encoder, not ~33%.** Both instruments are correct and the gap is coherent: the FFT is dense float at high IPC, so it retires more instructions per unit of time than the rest of the encoder. It also means the refused real-FFT work is worth *more* than I estimated — roughly **17–22% of encoder instructions**, not the ~15% I quoted from the wall-time share. The refusal stands regardless; the prize was simply bigger than stated.

**@chatgpt-codex-connector — the LSF finding is valid and fixed.** The row cycled `intf` while replaying one `a`/NLSF and one previous-envelope snapshot, and `lsf_quant_core`'s `abs_qerr < 0.25` early exit made that fix the refinement-iteration count. The capture loop was already deriving each frame's `a`; it now keeps them and threads the committed envelope forward as production does (frame *f*'s `brec` becomes frame *f+1*'s `prev_nlsf`).

**@greptile-apps — the state-divergence finding is half right, and I measured which half.** Instrumented the three stateful rows over 30 iterations:

| row | first 9 | last 9 | verdict |
| --- | --- | --- | --- |
| `pitch_search` | `541275, 945711, 995533` ×3 | **identical** | exact 3-cycle |
| `perc_model_frame` | `16054104, 16511156, 16562593` ×3 | **identical** | exact 3-cycle |
| `celp_subframes_frame` | `28,19,31,29,23,29,27,18,30` | `26,16,33,30,19,27,24,12,31` | **wanders** |

For pitch and perc the concern doesn't arise — they're exactly periodic from the first iteration, and `stage_rows_are_three_periodic` now pins that so it can't silently regress. You're right about CELP: its excitation history keeps evolving against a repeating residual, so it never settles. Its timing spread stays inside 1% (175.3–177.3 µs), so the fixed-size beam search dominates the exact pulse count — but the row is a steady-state estimate, not a reproduction of one production frame, and that is now recorded as a known limitation on the module rather than papered over.

I'm not restoring state per iteration to close it: that would put the restore cost inside the timed body, which is the exact failure this harness was built to avoid. `analyze_frame` is the row without the caveat — it drives the real encoder over eight cycling frames — which is why the stage rows are reported as shares *of it* rather than trusted standalone.

84 tests pass, both golden checksums hold, clippy and rustdoc clean.

---

jlucaso1:
Round six, in eb457b6. **@chatgpt-codex-connector** — both valid, both fixed, and the second one is the most embarrassing finding on this PR.

**The `required-features` finding is exactly the bug this PR exists to fix, reintroduced by me.** Putting `bench-internals` in the bench target's `required-features` meant `cargo bench -p wacore --features voip-mlow --bench voip_benchmark` — which worked before this branch — **silently ran nothing**. Cargo skips a bench target whose required features are off rather than erroring, which is precisely the mechanism I diagnosed for the CodSpeed shard three commits earlier and wrote a paragraph about. I then walked into it, and it would have taken out the MLow, SRTP, SFrame, engine and framing rows too, not just the new ones.

Fixed by requiring only `voip-mlow` on the target and `#[cfg]`-gating `codec_stages` inside the bench. Verified both ways rather than asserted:

```
$ cargo bench --features voip-mlow --bench voip_benchmark -- --list
├─ call_second_two_peers   ├─ mlow_decode              ╰─ framing
├─ crypt_payload_one       ├─ mlow_encode                 ├─ encode_rtp
├─ e2e_srtp_protect        ├─ mlow_encode_reused_output   ├─ parse_rtcp
├─ e2e_srtp_unprotect      ├─ sframe_decrypt              ├─ parse_rtp
├─ engine_inbound_packet                                  ╰─ parse_stun
├─ engine_outbound_frame                              (no codec_stages — correct)

$ cargo bench --features voip-mlow,bench-internals ...   → codec_stages: 9 rows
```

**The `prev_lsfq` threading finding is also correct.** I threaded the committed envelope into `fe.quantize`'s `prev_nlsf` argument but left `FrontEndLsf::prev_lsfq` and `smpl_lpc_interpol` pointed at the packet-start snapshot — so the rows were mixing one frame's envelope with another's. Production sets *both* `prev_lsfq` and `prev_nlsf` to the frame's committed NLSF after every internal frame (`analysis.rs`: `prev_nlsf = nlsf_out.clone(); prev_lsfq = nlsf_out;`), and all three consumers read it. One threaded value now feeds the quantizer, the envelope reconstruction and the interpolation that builds the CELP residual.

**Re-measured** (5 runs, minimum per row, against `analyze_frame` = 1.786 ms):

| stage | min | ×/frame | per frame | share |
| --- | ---: | ---: | ---: | ---: |
| `perc_model_frame` | 194.0 µs | 3 | 582 µs | 32.6% |
| `celp_subframes_frame` | 176.9 µs | 3 | 531 µs | 29.7% |
| `pitch_search` | 55.7 µs | 3 | 167 µs | 9.4% |
| `lpc_front_end` | 55.0 µs | 3 | 165 µs | 9.2% |
| `lsf_quantize` | 7.22 µs | 3 | 22 µs | 1.2% |
| `entropy_encode` | 2.30 µs | 1 | 2 µs | 0.1% |
| accounted | | | 1469 µs | 82.2% |

FFT total 598 µs — 33.5% by wall time; **43.5% by instruction count**, which is the figure the PR's conclusion uses.

That is nine benchmark-fidelity defects found by review across six rounds, plus one production regression, and **still none of the production optimizations has been challenged** — the four arithmetic-neutral changes and both golden checksums have held throughout. The measurement scaffolding turned out to be far harder to get right than the thing it measures, which is worth saying plainly given the PR's whole argument is "don't optimize what you can't attribute."

84 tests pass, both checksums hold, clippy and rustdoc clean.

---

PR #1347, opened 2026-08-24
Update every dependency to its current release

jlucaso1:
`cargo outdated -w` reports the workspace as fully current after this. Most of it is a lockfile move; three needed a manifest change and are worth explaining.

## opus 0.3 -> 0.4

The release swapped its native backend from `audiopus_sys` to `opusic-sys`. The Rust API `src/voip/audio.rs` uses is unchanged, so no code moved.

The reason to call it out is the build script. `audiopus_sys` compiles the vendored Opus through a `CMakeLists.txt` that declares a minimum below 3.5, and CMake 4 refuses that outright:

```
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.
```

Any consumer building this crate where pkg-config cannot find a system Opus hits that before compiling a line of Rust, which is every GitHub-hosted macOS and Windows runner. Downstream can work around it with `CMAKE_POLICY_VERSION_MINIMUM=3.5`, but the fix belongs here.

## rtc-* beta.1 -> beta.2

Still exact-pinned, for the reason already in the manifest: 0.21 is a pre-1.0 beta where a minor bump may break a call at runtime rather than at compile time, so the pin moves deliberately or not at all. This moves it deliberately.

## libsqlite3-sys 0.37 -> 0.38

Safe because diesel 2.3 accepts `>=0.17.2, <0.39.0`, so diesel and this crate resolve to a single copy. Two would mean two bundled SQLite builds exporting the same symbols.

## Verification

`cargo check --workspace --all-targets` and `cargo fmt --check` are clean locally. The heavy suites are left to CI, which is what actually exercises the codec round-trips and the VoIP stack these bumps touch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -61,23 +61,23 @@ dependencies = [
 
 [[package]]
 name = "aes-gcm"
-version = "0.11.0"
+version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fdf011db2e21ce0d575593d749db5554b47fed37aff429e4dc50bc91ac93a028"
+checksum = "7f2b8006a0c83f52b62ba44a97b58bf76fe2f70a329e588f67f89691d93d498f"
 dependencies = [
  "aead 0.6.1",
  "aes 0.9.2",
  "cipher 0.5.2",
  "ctr 0.10.1",
+ "ctutils",
  "ghash",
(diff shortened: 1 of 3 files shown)


jlucaso1:
@codspeedbot explain why this is faster

PR #1397, opened 2026-09-03
feat(presence): let the host own the automatic available announcement

bryantebeek:
## Summary

The client announces `<presence type="available">` on its own at three points, matching WhatsApp Web: on connect when the push name is already known (`Client::handle_success`, `src/client/node_io.rs`, reconnection path), when the push name arrives through the initial app-state sync (`src/client/app_state.rs`, the `setting_pushName` mutation), and when the server updates it (`Client::update_push_name_and_notify`, `src/client/accessors.rs`). That is the right default for an interactive client. It is the wrong one for a companion that stays connected around the clock: every automatic reconnect shows the account online, and an available device is where the server directs reads, so the phone stops being notified of new messages while the companion is up.

This adds `PresencePolicy` with `Automatic` (default, unchanged behaviour) and `Manual`, which suppresses exactly those three announcements. Explicit `Presence::set` / `set_available` / `set_unavailable` calls are untouched, so a manual host still decides when the account is available and remains responsible for saying so.

## WA Web

Nothing about the wire changes. The three automatic sends already carry the comments that tie them to WhatsApp Web (`WAWebPushNameSync`: presence deferred until a push name exists, then sent immediately), and `Automatic` keeps them. `Manual` is the host-side opt-out for a device class WhatsApp Web itself never is: one that is connected but not being looked at.

Second opinion on the effect: Baileys documents the same thing from the other direction. Its `markOnlineOnConnect` option marks the client available on connect and its docs note that "if `markOnlineOnConnect` is true, your phone won't receive notifications while Baileys is connected" ([Configuration](https://baileys.wiki/docs/socket/configuration/), [Presence and Receipts](https://baileys.wiki/docs/socket/presence-receipts/)).

## Design

- **One decision point.** `Client::send_automatic_available` (`src/features/presence.rs`) checks the policy and forwards to `Presence::set_available`. The three call sites switch from calling `set_available` to calling it, and log the two outcomes apart (`Ok(true)` sent, `Ok(false)` suppressed, `Err` failed). The policy check precedes the push-name check on purpose: a manual host must not be warned that a stanza it never asked for could not be sent.
- **Storage.** One `AtomicBool` on `Client` (`automatic_presence`), initialised to `true`, following `skip_history_sync`. `Client::set_presence_policy` / `Client::presence_policy` expose it, so the policy can be changed live like the other runtime toggles; changing it sends or retracts nothing by itself.
- **Surface.** `BotBuilder::with_presence_policy` and `ClientBuilder::with_presence_policy`, both defaulting to `Automatic`; `PresencePolicy` is `#[non_exhaustive]` and re-exported from `features` and the crate root like `PresenceStatus`. No feature gate: the type is one enum and the field one bool, and the policy is read on the core connection path.
- **Not done.** No automatic `unavailable` on disconnect and no "available while a window is focused" helper; those are host policy, and `Presence::set` already covers them.

## Testing

- `manual_policy_suppresses_only_the_automatic_announcement`: with `Manual` and an empty push name, `send_automatic_available` answers `Ok(false)` without reaching the push-name check, while an explicit `set_available` still fails with `PushNameEmpty`; switching back to `Automatic` at runtime makes the automatic path validate again.
- `presence_policy_defaults_to_automatic`, and `low_level_builder_installs_options_and_owned_services` now also covers the `ClientBuilder` setter.
- Locally: `cargo fmt --all -- --check`, `cargo clippy -p whatsapp-rust --all-targets -- -D warnings`, `cargo test -p whatsapp-rust --lib` (1868 passed), `cargo check -p whatsapp-rust --no-default-features`, `RUSTDOCFLAGS="-D warnings" cargo doc -p whatsapp-rust --no-deps`. The workspace-wide clippy and `--all-features` doc runs need `libopus`/`libasound2-dev`, which this machine does not have; CI covers them.

Motivation: [omarchy-whatsapp](https://github.com/bryantebeek/omarchy-whatsapp), an always-connected Linux desktop companion that advertises `available` only while its panel is focused. It has been running on a fork carrying this change; this PR is the same patch rebased onto `main`.


jlucaso1 (approved):
Thank you

PR #1507, opened 2026-09-16
Add chat lock verbs, LockChatUpdate event, and collection resync from version 0

FelipeMayerDev:
## What

Surfaces WhatsApp's chat lock (the hidden "locked chats" folder) the same way mute and pin are surfaced, plus one new sync primitive:

- **`ChatActions::lock_chat` / `unlock_chat`** — send the `lock` syncd mutation (`WAWebLockChatSync`, regular_low, index `["lock", jid]`, `LockChatAction.locked`). The schema (`schemas::LOCK_CHAT`) already existed; only the verbs were missing.
- **`Event::LockChatUpdate`** — appended (frozen `Event` API: index-based serialization forbids inserting), dispatched from `dispatch_chat_mutation` for the `"lock"` kind, with the same `jid`/`timestamp`/`action`/`from_full_sync` shape as `MuteUpdate`. `EventKind::LockChatUpdate` appended with the capacity guard updated.
- **`Client::resync_app_state_collection`** — resets the stored version of a collection and reruns the full sync, replaying every mutation the server still holds. `full_sync: true` alone pages from the stored version, so a consumer that missed past mutations (it had no handler for them at the time) has no other way to ask for them. Whatsmeow's `removeAppState` resync is the same maneuver. Consumers must be idempotent/order-guarded — the replay arrives with `from_full_sync: true` and original timestamps.

## Verification

- `cargo test -p whatsapp-rust --lib chat_actions` — includes new round-trip tests: lock verb produces `["lock", jid]` on regular_low and dispatches `LockChatUpdate`; both Set directions dispatch; a syncd Remove on `lock` is not claimed.
- `cargo test -p wacore --lib` (1541 tests), `cargo clippy -p whatsapp-rust -p wacore --all-targets`, `cargo fmt --all --check` clean.

Consumer: ZapFast uses this for a locked-chats folder (hide from list/search/unread badge, generic notifications, secret-code reveal in search); the version-0 replay recovers locks recorded on the phone before the client could handle them.


jlucaso1:
Thanks for that, can you solve coderabbit reviews things and also solve the conflicts?

Salientekill:
Interesting

src/client/app_state.rs
```diff
@@ -949,6 +947,31 @@ impl Client {
...
+    /// (keeping sync keys and mutation MACs in the store) and re-runs the full
+    /// sync, so every mutation the server still holds is decoded and dispatched
+    /// again. The consumer's handlers see the replay with `from_full_sync: true`
+    /// and must be idempotent/order-guarded (ZapFast's mute/pin/lock handlers
+    /// are). Whatsmeow's "remove app state" resync is the same maneuver.
+    pub async fn resync_app_state_collection(self: &Arc<Self>, name: WAPatchName) {
+        let backend = self.persistence_manager.backend();
+        if let Err(error) = backend
+            .set_version(name.as_str(), wacore::appstate::hash::HashState::default())
+            .await
```
jlucaso1: I think the resync itself is useful, but I don't think we should reset the collection state directly here before acquiring the collection reservation.

The existing app-state synchronization code is quite deliberate about serializing changes to the persisted version / LTHash / mutation MAC state. `process_sync_task(AppStateSync)` acquires the collection reservation, but at this point `set_version(...)` has already modified the store outside that reservation. A concurrent sync or patch send could therefore observe or write app-state while this collection is being reset.

We already seem to have the machinery for this exact operation in the rebuild path: `BatchedSyncRequest { rebuild: true, wait_for_holder: true }` together with `stand_collection_down()`. That path acquires the collection reservation first, checks the sync scope, deletes the version, clears the mutation MACs in the intended order, and then requests the snapshot.

Could we implement `resync_app_state_collection()` on top of that existing rebuild path instead of manually writing `HashState::default()` here?

Something conceptually like:

```rust
self.sync_collections_batched_with(
    vec![name],
    self.sync_scope(None),
    BatchedSyncRequest {
        rebuild: true,
        wait_for_holder: true,
    },
)
.await
```

(or a small internal helper wrapping the same machinery, if we don't want the public API coupled to the batched implementation).

This would also keep the resync behavior consistent with the concurrency/reconnect invariants already enforced by the app-state subsystem.

jlucaso1 (changes requested):

src/client/app_state.rs
FelipeMayerDev: Resolved in 58ccb3a. `resync_app_state_collection` now delegates to `resync_app_state([name], AppStateResyncMode::Snapshot)`, which uses `sync_collections_batched_with` with `rebuild: true` and `wait_for_holder: true`. That reuses the reservation, scope fencing, version/MAC reset ordering, pagination and outcome reporting instead of writing `HashState` directly.

jlucaso1 (approved):

jlucaso1:
Thanks <3

